PR #1, opened 2023-02-19
optional URI NUT

gandlafbtc:
According our discussion on telegram.


callebtc:
This is now addressed in #2

PR #3, opened 2023-03-09
NUT07: spendable check

callebtc:


callebtc:
calle, [9. Mar 2023 at 17:34:47 (9. Mar 2023 at 17:35:18)]:
An Improvement of the /check spendable operation could be using Bloom filters in a future upgrade. A mint could provide a bloom filter (the mint would have to recompute the bloom filter after each spend) to a wallet and the wallet could check whether a proof definitely not in the set of spent secrets.

It remains to be seen how feasible it is to update the bloom filter for a mint. It could be too computationally intense or the the data requirements could be too large to transfer the filter to a wallet after each update of the spent secrets table.

PR #5, opened 2023-03-20
NUT-08: Dynamic amount of tokens for change

Semisol:
Use `ceil(log2(fee reserve))` for the amount of blank tokens to send.

The reason this works:
- You only need `n` bits to represent a binary number that is smaller or equal to an `n` bit binary number.
- Any amount of tokens can be represented as a binary number with bits `ceil(log2(amount))`, and any binary number can be converted into a set of tokens.
- The fee reserve change is smaller or equal to the fee reserve.


BilligsterUser (approved):

08.md
```diff
@@ -11,11 +11,11 @@ The problem is also described in [this gist](https://gist.github.com/callebtc/a6
 
 ## Description
 
-Before requesting a Lightning payment as described in [NUT-05][05], `Alice` produces a predefined number of `BlindedMessage` which are similar to ordinary blinded messages but their value is yet to be determined by the mint `Bob` and are thus called *blank outputs*. The predefined number is chosen to be `n_blank_outputs := 4` by convention which means that we can return at minimum 93.75% (but up to 100%) of the overpaid fees back to the user (because each blank output can account for at least 50% of the remaining overpaid fees).
+Before requesting a Lightning payment as described in [NUT-05][05], `Alice` produces a number of `BlindedMessage` which are similar to ordinary blinded messages but their value is yet to be determined by the mint `Bob` and are thus called *blank outputs*. The amount of blinded outputs is `ceil(log2(fee_reserve))`, which ensures that any amount <= the fee reserve can be returned back by the mint
```
callebtc, line 14:
```suggestion
Before requesting a Lightning payment as described in [NUT-05][05], `Alice` produces a number of `BlindedMessage` which are similar to ordinary blinded messages but their value is yet to be determined by the mint `Bob` and are thus called *blank outputs*. The number of necessary blank outputs is `ceil(log2(fee_reserve))` which ensures that any overpaid fees can be represented and be returned by the mint to the wallet.
```

08.md
```diff
@@ -11,11 +11,11 @@ The problem is also described in [this gist](https://gist.github.com/callebtc/a6
...
+The wallet wants to pay an invoice with `amount := 100 000 sats` and determines by asking the mint that `fee_reseve` is `1000 sats`. The wallet then provides `101 000 sats` worth of proofs and 10 blank `outputs` to make the payment. The mint pays the invoice and determines that the actual fee was `100 sat`, i.e, the overpaid fee to return is `fee_return = 900 sats`. The mint splits the amount `900` into summands of `2^n` which is `4, 128, 256, 512`. The mint inserts these amounts into the blank `outputs` it received form the wallet and generates 4 new promises. The mint then returns these `BlindedSignature`'s to the wallet together with the successful payment status.
```
callebtc, line 18:
```suggestion
The wallet wants to pay an invoice with `amount := 100 000 sats` and determines by asking the mint that `fee_reseve` is `1000 sats`. The wallet then provides `101 000 sats` worth of proofs and 10 blank `outputs` to make the payment (since log2(1000)=0.97...). The mint pays the invoice and determines that the actual fee was `100 sat`, i.e, the overpaid fee to return is `fee_return = 900 sats`. The mint splits the amount `900` into summands of `2^n` which is `4, 128, 256, 512`. The mint inserts these amounts into the blank `outputs` it received form the wallet and generates 4 new promises. The mint then returns these `BlindedSignature`'s to the wallet together with the successful payment status.
```

callebtc (approved):

08.md
```diff
@@ -24,7 +24,7 @@ The wallet asks the mint for the `fee_reserve` for paying a specific bolt11 invo
 
 Here we describe how the mint generates `BlindedSignature`'s for the overpaid fees. The mint `Bob` returns in `PostMeltResponse` the field `change` **ONLY IF** `Alice` has previously provided `outputs` for the change **AND** if the Lightning `actual_fees` were smaller than the `fee_reserve`.
 
-If the `overpaid_fees = fee_reserve - actual_fees` is positive, `Bob` decomposes it to values of `2^n` (as in [NUT-00][00]). `Bob` then takes at most `n_blank_outputs` **largest** of these amounts and imprints them into the `ouptputs` provided by `Alice`. 
+If the `overpaid_fees = fee_reserve - actual_fees` is positive, `Bob` decomposes it to values of `2^n` (as in [NUT-00][00]) and then imprints them into the `blank_outputs` provided by `Alice`. 
 
 `Bob` then signs these blank outputs (now with the imprinted amounts) and thus generates `BlindedSignature`'s. `Bob` then returns a payment status to the wallet, and, in addition, the up to `n_blank_outputs` blinded signatures it generated for the overpaid fees.
```
callebtc, line 29:
```suggestion
`Bob` then signs these blank outputs (now with the imprinted amounts) and thus generates `BlindedSignature`'s. `Bob` then returns a payment status to the wallet, and, in addition, all blind signatures it generated for the overpaid fees.

Importantly, while `Bob` does not necessarily return the same number of blind signatures as it received blank outputs from `Alice` (since some of them may be of value 0), `Bob` **must** return the all blank signatures with a value greater than 0 in the same order as the blank outputs were received and should omit all blind signatures with value 0. For example, if `Bob` receives 10 blank outputs but the overpaid fees only occupy 4 blind signatures, `Bob` will only return these 4 blind signatures with the appropriate imprinted amounts and omit the remaining 6 blind signatures with value 0. Due to the well-defined order of the returned blind signatures, `Alice` has can map the the blind signatures returned from `Bob` to the blank outputs it provided so that she can further apply the correct unblinding operations on them.
```

callebtc:
I added some further clarifications on the order of the returned blind signatures because they need to be mappable back to the blank outputs a wallet provides. 

I think the entire document deserves another round of reading and maybe restructuring, so that 1) Explanation of the concept 2) Example 3) And clear rules of the protocol are bit better separated and the reader can skim through the document faster.

callebtc:
### Tracking progress (please report):

- [ ] Nutshell
- [ ] LNbits
- [ ] Feni
- [ ] Cashu.me
- [ ] [Nutstash](https://github.com/gandlafbtc/nutstash-wallet)
- [ ] [Peanut](https://github.com/cashubtc/peanut)
- [ ] [cashu-ts](https://github.com/cashubtc/cashu-ts)
- [ ] [ngutech21/cashu-rs](https://github.com/ngutech21/cashu-rs)
- [ ] [clarkmoody/cashu-rs](https://github.com/clarkmoody/cashu-rs)

BilligsterUser:
[cashu-ts](https://github.com/cashubtc/cashu-ts) => [pr #59](https://github.com/cashubtc/cashu-ts/pull/59)

PR #18, opened 2023-05-14
Add a few test vectors for NUT-00

thunderbiscuit:
This is just an attempt to show what adding vectors to the NUTs could look like. Happy to hear if there might be a better format or approach. If this looks good, I can take some time and add a bunch more, pulling from the Python, TypeScript, and Go codebases.

I tried to take inspiration from the [BIP-32](https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki) test vectors, but with codeblocks so as to make it easier to copy in some cases.

[Trezor uses a different approach](https://github.com/trezor/python-mnemonic/blob/master/vectors.json) with JSON objects containing the tests, but I find this approach not as universal because in many cases it helps to add failing tests and explain why they should fail, and JSON doesn't support comments.

### DIFF
--- a/test-vectors/00-tests.md
+++ b/test-vectors/00-tests.md
@@ -1,4 +1,23 @@
 # NUT-00 Test Vectors
+## Blind Diffie-Hellman key exchange section
+The following are test vectors related to the [Blind Diffie-Hellman key exchange section](https://github.com/cashubtc/nuts/blob/main/00.md#blind-diffie-hellmann-key-exchange-bdhke).
+
+### Hash-to-curve function
+The hash to curve function takes a message of any length and outputs a valid point on the secp256k1 curve.
+```shell
+# Test 1 (hex encoded)
+Message: 0000000000000000000000000000000000000000000000000000000000000000
+Point:   0266687aadf862bd776c8fc18b8e9f8e20089714856ee233b3902a591d0d5f2925
+
+# Test 2 (hex encoded)
+Message: 0000000000000000000000000000000000000000000000000000000000000001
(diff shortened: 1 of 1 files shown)


thunderbiscuit:
Also I realize it's my first PR on the Cashu GitHub org. I should say hi first.

I've been following from a distance the Cashu protocol for a few months but didn't take the time to dig too hard into it. I finally did a few weeks ago. Super interesting stuff. I'm still working through the mental models and how/where these ideas can be best leveraged, and there is nothing like reading a spec and just hacking away at implementing it. I have started a small personal Kotlin library focused on the client-side of the protocol to get my hands dirty and understand better the pieces of that puzzle. So far I'm pulling tests from the Python and TS libraries.

I'm an Android developer and a maintainer of the [language bindings for the Bitcoin Development Kit](https://github.com/bitcoindevkit/bdk-ffi).

callebtc:
Thank you very much, this is great! Can I convince you to adopt the vectors tested [here](https://github.com/cashubtc/cashu/blob/main/tests/test_mint.py)? 

Afaik, we use the same tests in Nutstash (link above) and in [Feni](https://github.com/cashubtc/cashu-feni/blob/master/mint/mint_test.go).

thunderbiscuit:
Sounds good. I'll take a look and start adding.

Also I think I got one test incorrect; a `cashuB` prefix should not be wrong under the V3 token, because of course the token allows for incrementing this version flag (me: 🤦‍♂️). Will fix.

thunderbiscuit:
Oh I should have opened it as a draft. I still have to add the unblinding and verification steps to complete NUT-00. Will add them in my next PR!

PR #21, opened 2023-05-15
NUT-07: add new field `pending`

callebtc:
This is necessary in order for a wallet to determine whether a token it used for making a Lightning payment is still in-flight or not.

In short, LN payments can have 3 states but our tokens only 2 (because there's only `spendable` or not). We need another boolean state (`pending`).

Using this, wallets can determine the state of an outgoing Lightning payment as follows:

- if a token is `not spendable` and `not pending`, the payment succeeded
- if a token is `spendable` and `pending`, the payment is in-flight
- if a token is `spendable` and `not pending`, the payment failed

### DIFF
--- a/07.md
+++ b/07.md
@@ -1,7 +1,7 @@
 NUT-07: Token state check
 ==========================
 
-`optional` `author: calle`
+`optional` `author: calle` `author: tb`
 
 ---
 
--- a/07.md
+++ b/07.md
@@ -21,9 +21,11 @@ A wallet can internally represent the combination of these booleans the proof's 
 
 ## Use cases
 
-**Example 1:** When `Alice` prepares a token to be sent to `Carol`, she can mark these tokens in her database as *db:pending*. She can then, periodically or upon user input, check with the mint if the proof is still spendable (mint returns `True` for `spendable`) or if it has been redeemed by `Carol` already (mint retrns `False` for `spendable`). If the proof is not spendable anymore (and, thus, has been redeemed by `Carol`), she can safely delete the proof from her database.
+#### Example 1: Ecash transaction 
+When `Alice` prepares a token to be sent to `Carol`, she can mark these tokens in her database as *db:pending*. She can then, periodically or upon user input, check with the mint if the proof is still spendable (mint returns `True` for `spendable`) or if it has been redeemed by `Carol` already (mint retrns `False` for `spendable`). If the proof is not spendable anymore (and, thus, has been redeemed by `Carol`), she can safely delete the proof from her database.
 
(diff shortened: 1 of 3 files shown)


gandlafbtc:
LGTM

BilligsterUser (approved):
👍

07.md
```diff
@@ -1,11 +1,22 @@
...
+With the spendable check, wallets can ask the mint whether a specific token is already spent (`spendable`) or pending in a transaction (`pending`). 
+
+- A token is `not spendable` if it has been redeemed and its secret is in the list of spent secrets of the mint.
+- A token is `pending` if it is being processed in a transaction (either being redeemed for another token or associated with an ongoing payment). A `pending` token cannot be used in another transaction until it is `not pending`.
+
+**Important:** Before deleting spent tokens from their database, wallets MUST check if the token is still `spendable` to make sure that they don't delete an unspent token by accident.
```
thunderbiscuit, line 13: Other uses of the MUST word in the spec write it in bold (`**MUST**`).

07.md
```diff
@@ -1,11 +1,22 @@
...
+
+## Use cases
+
+**Example 1:** When `Alice` prepares a token to be sent to `Carol`, she can mark these tokens in her database as *db:pending*. She can then, periodically or upon user input, check with the mint, if the token is still spendable (mint returns `True` for `spendable`) or if it has been redeemed by `Carol` already (mint retrns `False` for `spendable`). If the token is not spendable anymore (and, thus, has been redeemed by `Carol`), she can safely delete the token from her database.
```
thunderbiscuit, line 17: I think this comma between _mint_ and _if_ might be a typo. Reads better as "_check with the mint if the token is still spendable_".

07.md
```diff
@@ -1,11 +1,22 @@
...
+
+**Example 2:** If Alice pays a very slow Lightning invoice (for example a HODL invoice) and closes her wallet in the mean time, the next time she comes online, she can check all tokens marked as *db:pending* in her database to determine, whether the payment is still in flight (mint returns `True` for `pending` and `spendable`), it has succeeded (mint returns `False` for `pending` and `spendable`), or it has failed (mint returns `False` for `pending` and `True` for `spendable`).
```
thunderbiscuit, line 19: _Meantime_ is written as one word instead of two.

thunderbiscuit (commented):
ACK 16e0a39830b681dbebd9c97b389a68b351c2b0c6.

I pointed a few nits if you're interested but nothing worth blocking over.

A general thought on the wording of those token states (which might belong more in an issue than on this PR so feel free to ignore). It makes sense to have booleans in this json object for communication between the mint and client, but it might be nice to outline better wording for those states than the negation of `spendable` and `pending` as `not spendable` and `not pending`. In my mind, those two booleans immediately combine become an enum, and I'm likely to call the enum variants something like `spendable/burned/inflight` or `live/burned/pending`.

This is more of an implementation detail (the json fields should stay spendable/pending) but as [outlined in this awesome and very liked PR](https://github.com/Xekyo/bips/pull/1) from Murch, providing good wording is always a big win.

07.md
callebtc: I appreciate this a lot, thank you!

PR #22, opened 2023-05-15
Add clarification around public key format mints should use

thunderbiscuit:
This PR is a small addition to the definition of the keyset type, stating that the public keys must be serialized using the compressed public key format used everywhere in bitcoin when building keysets.

This is not unique to bitcoin so I couldn't find a BIP to link to for further information on the format, but information about it is basically found everywhere, including many questions and answers on Stack Overflow and a section of Mastering Bitcoin by Antonopoulos. I chose an article from learnmeabitcoin.com for the link. 

Note that this is also the approach used in one of the [BIPs defining SegWit](https://github.com/bitcoin/bips/blob/master/bip-0143.mediawiki#restrictions-on-public-key-type), where only compressed keys are accepted, standardizing the length required to share them as well as making them shorter in general.

### DIFF
--- a/02.md
+++ b/02.md
@@ -5,7 +5,7 @@ NUT-02: Keysets and keyset ID
 
 ---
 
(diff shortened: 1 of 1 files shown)


callebtc:
Thank you for clarifying this and sorry for the delay!

PR #27, opened 2023-06-06
Add test vectors

thunderbiscuit:
Just slowly making my way through.

- [x] NUT-01
- [x] NUT-02
- [ ] NUT-03
- [ ] NUT-04
- [ ] NUT-05
- [ ] NUT-06
- [ ] NUT-07
- [ ] NUT-08
- [ ] NUT-09

Feel free to comment or request additional tests.


AngusP:
@thunderbiscuit thanks for taking the time to add tests 🎉  -- for NUT-3, this issue might be useful for some vectors: https://github.com/cashubtc/cashu/issues/228

Effectively implementations **should** support any amount (including over 21,000,000 * 100,000,000 sats, which is impossible), and let their upstream (lightning or whatever) fail rather than the mint request itself breaking, but it's probably Ok for impls ("**must**") to be able to support up to 51 bit uints (ceil log2 of 2.1 quadrillion sats), so probably 8-byte integers (uint64)

Also notable here the issue was with the backing database, rather than the mint itself because Python handles very large integers pretty well

thunderbiscuit:
@AngusP thanks for the idea for the extra tests! Related to this is #28.

This PR is getting a bit old and I don't want people to wait for me to finish it all before merging, just because I'm not sure how much time I'm going to have to dedicate to this over the fall.

@callebtc if you like the tests here, feel free to comment/request changes/merge and not wait for me to add tests for all the NUTs.

test-vectors/02-tests.md
```diff
@@ -0,0 +1,16 @@
+# NUT-02 Test Vectors
```
thunderbiscuit, line 1: The following keysets require a cross-check from at least one different implementation (mine is [cashu-client](https://github.com/thunderbiscuit/cashu-client), particularly the short keyset.

I looked at the test suite from the Python and Go implementations but could not find explicit keysets in string format I could easily copy/paste; they're defined somewhere else and I wasn't sure how to pull them out quickly. The long keyset here comes from the [cashu-rs test suite](https://github.com/clarkmoody/cashu-rs/blob/8560f87862ecece791cb5f7b71b80839bcbe639e/src/keyset.rs#L290-L358) so I am more confident in, but the short one is a custom one I built (if I remember correctly) by cutting out the tail end of a keyset given to me by the default mint on the Python implementation cli wallet. These should not be accepted as "official test vectors" unless they correspond to the keyset IDs other implementations get.

thesimplekid: The short keyset and id looks good to me.
https://github.com/thesimplekid/cashu-crab/commit/59250bc79ea61aeb2eafbe98a0b691ae47ab47fd

thunderbiscuit: Awesome! Thanks for the quick test. It also gives me a new place to look into for tests by other implementations. Cheers!

callebtc (approved):
beautifully diligent work 🙏

PR #31, opened 2023-06-16
NUT-XX: Animated QR codes

gandlafbtc:
Ok this is a draft for the animated QR codes. 

It is certainly not the most (byte wise) efficient solution, but it is easy to implement and that was the basis for the design choices. If efficiency is the more important, we could lean more towards what is done [here](https://github.com/gre/qrloop) . In my opinion, keeping it simple and easy to include into all kind of libs/apps outweigh the efficiency gains we would get from that though. 

Let me know what you guys think.


thunderbiscuit:
Love the animated QRs.

I've been looking into them for the bitcoin development kit mobile bindings for a little while and figured I'd share some thoughts.

The most prominent and well-developed "maybe-standard?" for them is developed by [Blockchain Commons](https://github.com/BlockchainCommons). They have a standard called [UR (unified resource)][https://github.com/BlockchainCommons/crypto-commons) used by [more and more wallets](https://github.com/BlockchainCommons/Gordian-Developer-Community/blob/master/README.md#urs) which is basically using CBOR to encode wallet-related data (PSBTs, keys, others, etc.). Building on this, they have defined a standard for [animated QR codes](https://www.blockchaincommons.com/devs/animated-qrs.html) which has implementations in [some popular languages](https://github.com/BlockchainCommons/Research/blob/master/papers/bcr-2020-005-ur.md#implementations).

Their docs are very thorough but all over the place; you have to dig in to find related pages.

Some extra links:
Implementations: [page 1](https://github.com/BlockchainCommons/crypto-commons#bc-ur) [page 2](https://github.com/BlockchainCommons/Research/blob/master/papers/bcr-2020-005-ur.md#implementations)
Spec: https://github.com/BlockchainCommons/Research/blob/master/papers/bcr-2020-005-ur.md
UR types (tokens would have to use `bytes` but could request a new type be added to the spec): https://github.com/BlockchainCommons/Research/blob/master/papers/bcr-2020-006-urtypes.md

The pros of going with an approach like this is that you'll likely find libraries that implement that encoding and decoding (good for interop) and not require all Cashu wallets implement a new standard. The con is that this particular standard is fairly complex and will require time/effort to bring everyone up to speed. It is a very good sign that other bitcoin wallets all over the place are using it however, and probably a strong foundation to build on.

Hopefully this can add to the discussion and shed light on tradeoffs.

thunderbiscuit:
Blockchain Commons has released an implementation guide for the fountain code standard, [and it's here](https://github.com/BlockchainCommons/Research/blob/master/papers/bcr-2024-001-multipart-ur.md). I'm about halfway done building a library [here](https://github.com/kotlin-bitcoin-tools/ur), but you can find working implementations in TS/JS/Java/Swift [here](https://github.com/BlockchainCommons/Research/blob/master/papers/bcr-2020-005-ur.md#implementations).

gandlafbtc:
i tweaked the typescript implementation of bc-ur here https://github.com/gandlafbtc/bc-ur to make it work in the browser. Obviously blockchain commons UR is superior in every way. I changed The nut-number, and added some basic text for the NUT

AngusP:
I agree using UR for animated makes sense, afaik it has the most existing implementations among Bitcoin wallets so is the de-facto standard at the moment, and probably be most familiar to devs.

Ofc there's [BBQR](https://bbqr.org/) or qrloop and probably other options, but the same spirit of "JSON will work for now" should push towards UR vs the other 'technically better' options?

gandlafbtc:
Not even sure if other options are "technically better". I think what BC UR does fits cashus usecase pretty well.

callebtc:
I've recently implemented UR in cashu.me using gandlaf's awesome library `@gandlaf21/bc-ur` here: https://github.com/gandlafbtc/bc-ur

It was *very* easy to add and felt a bit like magic because it just works.

callebtc:
We should pick up this PR again, I think the document can be extremely short since there is a canonical way of how to produce bc-ur animated QR codes and we already have a cross-compatible implementation in cashu.me and nutshell.

@gandlafbtc, what do you think?

PR #33, opened 2023-06-26
NUT-08: handle log(0) and log(1) possibilities

callebtc:
For `fee_reserve == 0` (if the payment is internal for example), and for `fee_reserve == 1` the calculation breaks.

### DIFF
--- a/08.md
+++ b/08.md
@@ -11,7 +11,17 @@ The problem is also described in [this gist](https://gist.github.com/callebtc/a6
 
 ## Description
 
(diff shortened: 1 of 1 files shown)


gandlafbtc (approved):
lgtm

moonsettler (changes requested):
> which ensures that there is at least one output if there is any fee. 

`min(ceil(log2(fee_reserve)), 1)`

min ensures there is at least 1 output, and maximum `ceil(log2(fee_reserve))` outputs.

moonsettler (approved):
my mistake, never mind!

PR #34, opened 2023-06-29
NUT-06: Deprecate `amount` in `POST /split`

### LINKED ISSUE #32: NUT-06: Deprecate amount field on PostSplitRequest
```json
{
  "proofs": Proofs,
  "outputs": BlindedMessages,
  "amount": int //deprecate
}
```

the amount field is not necessary, since it introduces unwanted constraints. 
The client has to manage split amounts (send, return) , but the mint does not need to. The mint only needs to perform split according to the clients request.

callebtc:
The `amount` field in `POST /split` was an unnecessary contraint on the split operation. Now, wallets can choose the amount distribitions of the inputs (proofs) and outputs (and thus returned promises) as they wish, as long as they sum up to the same value. 

Closes #32 

### Wallets:
Wallets simply don't send the `amount` field with the `POST /split` request anymore. They receive a new object with the field `promises` from which they then need to separate the promises to keep (first elements) and send (last elements). 

Previously, they took these two sets of elements from the fields `fst` and `snd` from the response. This is no longer the case! Note that wallets already should know how many elements of the `promises` response they need to slice to get the same as the `fst` and `snd` responses because they have previously generated an array of `outputs` that has the same distribution.

### Mints:
Mints now do no receive an `amount` field anymore. They can simply proceed as usual but remove any security checks that might have previously asserted that the distribution of amounts of `outputs` must adhere to a certain pattern (for example perfectly decompose in powers of two). This is not necessary anymore. They MUST still check whether the sum of `proofs` is equal to the sum of `outputs` and they SHOULD also check whether it equals the sum of generated `promises` (as a sanity check). 

`PostSplitRequest`

```json
{
  "proofs": Proofs,
  "outputs": BlindedMessages,
  "amount": int //deprecate
}
```

`PostSplitResponse`
```json
{
  "promises": BlindSignatures
}
```

### Tracking progress of NUT-06 update (please report):

- [x] Nutshell (Status: 🛠️ [PR](https://github.com/cashubtc/cashu/pull/263))
- [ ] [LNbits](https://github.com/lnbits/cashu)
- [ ] [Feni](https://github.com/cashubtc/cashu-feni)
- [ ] [Cashu.me](https://github.com/cashubtc/cashu.me)
- [ ] [Nutstash](https://github.com/gandlafbtc/nutstash-wallet)
- [ ] [eNuts](https://github.com/cashubtc/peanut)
- [ ] [cashu-ts](https://github.com/cashubtc/cashu-ts) 
- [ ] [ngutech21/cashu-rs](https://github.com/ngutech21/cashu-rs) 
- [ ] [clarkmoody/cashu-rs](https://github.com/clarkmoody/cashu-rs) 
- [x] [cashu-crab](https://github.com/thesimplekid/cashu-crab) (Status: ✅ [commit](https://github.com/thesimplekid/cashu-crab/commit/e8303fcc8c545111ddfbeb72739d5bd757123157))
- [x] [cashu-kotlin](...)

Ping: @BilligsterUser @gandlafbtc @clarkmoody @ngutech21 @gohumble @thesimplekid @KKA11010 @thunderbiscuit


thesimplekid:
Cashu-crab updated in https://github.com/thesimplekid/cashu-crab/commit/e8303fcc8c545111ddfbeb72739d5bd757123157, mint handles both with and without amount and responds accordingly, wallet chooses which to follow. Will deprecate amount fully after transition.

gandlafbtc:
If the amount is not sent anymore, mints that haven't upgraded yet will error. Shouldn't we keep the amount in the request, at least till most mints have upgraded?

callebtc:
> If the amount is not sent anymore, mints that haven't upgraded yet will error, right? Shouldn't we keep the amount in the request, at least till most mints have upgraded?

Same as @thesimplekid, I've upgraded Nutshell to be backwards compatible (has two different responses depending on whether amount is sent or not). I will add the same to LNbits and we should be relatively good. Only issue will be clients that have upgraded but want to talk to old mints but there will always be pain!

ngutech21:
Moksha mint and wallet upgraded https://github.com/ngutech21/moksha/commit/d0d39f5ffc4379ad7f1e2d49f570216cb5d48301

All changes are backwards compatible. I will remove the deprecated code, after all the other implementations have upgraded as well.

Issue #36, opened 2023-07-06
Proposal: sunset cashu_wallet_spec file


thunderbiscuit:
After some preliminary discussion on the Telegram channel, I wanted to open the discussion here and formally propose the `cashu_wallet_spec` file be removed from the repository. I see a few reasons for this:
1. Its name suggest a specification (but for wallets only?) but in fact, given all the other formally numbered NUTs, it ends up being more of a high-level overview of the workflow a wallet might go through. It's named a spec but doesn't have the same structure/name and the other NUT. Feels confusing at first.
2. The workflow it describes will tend to be more incomplete _or_ more complex as additional optional NUTs are added and decisions need to be made as to whether those should be added to the file or not. It's like the little brother you need to decide if you invite to the party or not.
3. It provides a second location where things might be defined and create competing/inconsistent wording. An example of that exists between line 35 of cashu_wallet_spec and line 78 of NUT-00:
> 35: `Alice` generates `N` random secret strings `x` by generating `128` random bits and encoding them `base64` for each requested token.
> 78: `secret` is the secret message (no encoding standard)

<br>

### Other considerations
I _think_ that there isn't anything in that file that is not already in the NUTs, and it can therefore be safely removed, but careful review should be done before removing it in case it adds something that current implementations abide by but that is not defined in any given NUT. 

There might also still be interesting/meaningful clarifications in that file that the NUTs could use. It's probably a good idea to make sure that the information gets integrated in the NUTs if need be rather than simply removed from the repo.

PR #47, opened 2023-09-13
fix double spend comment in wallet spec

elnosh:
should this say that `Carol` needs to swap the received tokens to prevent `Alice` from double spending them?


thunderbiscuit (approved):
ACK e28fc340f05af93313b6c0b2b53001b33452dc1b. Yep that should have said Alice instead.

Just an FYI, take a look at issue #36, which was also discussed over the dev call a few weeks back. If I understand correctly, the file will be removed (soonish?) and ported over to cashu.space, and so if you are actively developing software against the Cashu specification, you should not rely on it as the "final word" on the spec but rather a overview/discussion of how the protocol works.

elnosh:
ah, got it. Thanks for the clarification!

PR #48, opened 2023-09-17
Fix split to send example

sj-fisher:
The "first two tokens" are 8 and 8, which don't add to 40.


thunderbiscuit (approved):
ACK ab676e3. 

This mistake was mine; after writing the note on ordering the blinded messages to preserve privacy, I went back to change the order of the outputs but didn't change the associated sentence. Good catch!

callebtc:
Thank you, must have overlooked that!

Issue #54, opened 2023-09-30
Uniform secret structure would prevent wallet fingerprinting


thunderbiscuit:
The exact structure of secrets is currently left to the wallets to implement and there is no default value recommended in any NUT. Because not all client libraries build their secrets the same way, a mint can, in theory, eventually learn to fingerprint given libraries that produce certain tokens, lowering the anonymity set for those users.

It might make sense to either:
a. recommend  in one of the NUTs that client libraries use a common secret structure
b. enforce a secret size in the protocol for "simple/raw" secrets (not the secrets defined in NUT 10)

Opening this issue for discussion on these ideas.

callebtc:
I propose we all use 32 byte hex strings as secrets!

PR #55, opened 2023-10-03
Cashu `v1` – The Great Cleanup

callebtc:
# The Great Cleanup

X

This is a PR that encapsulates a collection of changes to the protocol that will make our lives easier in the future. It cleans up early mistakes that were made as the protocol grew organically and removes implicit assumptions about the funding sources (e.g. Lightning) and currency units (e.g. "sats") used for Cashu mints that are inherent in the current protocol.
# New `v1` API 
We are introducing the `v1` API where we clean up many input and output models. 

<details>
  <summary>Click to show new endpoints</summary>

### New endpoints
- New: `GET /v1/keys` and `GET /v1/keys/{keyset_id}` (new `Keyset` model)
- New: `GET /v1/keysets` (new `Keysets` model)
- New: `POST /v1/swap` (new `BlindedMessage` model)
- New: `POST /v1/melt/quote/bolt11` (replaces `GET /checkfees`)
- New: `POST /v1/melt/bolt11`
- New: `POST /v1/mint/quote/bolt11` (replaces `GET /melt`)
- New: `POST /v1/mint/bolt11`
- New: `POST /v1/check` (no changes)
- New: `GET /v1/info` (no changes)
- New: `POST /v1/restore` (new `GetInfoResponse`)

</details>

<details>
  <summary>Click to show error response</summary>

### Error responses
Error responses of the mint are now defined as a HTTP 400 response with a JSON body
```json
{
  "detail": "oops",
  "code": 1337
}
```
</details>

<details>
  <summary>Click to show deprecated endpoints</summary>

### Deprecated endpoints
- `GET /keys`, `GET /keys/{keyset_id}`
- `GET /keysets`
- `POST /split`
- `GET /mint`
- `POST /mint`
- `GET /checkfees`
- `POST /melt`
- `POST /check`
- `GET /info`
- `POST /restore`

</details>

# NUTs
- The `GET /checkfees` endpoint (formerly NUT-03) is now retired
- The `POST /v1/swap` endpoint is now in NUT-03 (formerly `POST /split` NUT-06). 
- NUT-06 is therefore free (suggestion: let's use it for deterministic secrets)

# New request and response models
<details>
  <summary>Click to show all changed models</summary>

## NUT-01
`GetKeysResponse`
```json
{
  "keysets": [
    {
      "id": <keyset_id_hex_str>,
      "unit": <currency_unit_str>,
      "keys": {
        <amount_int>: <public_key_str>,
        ...
      }
    }
  ]
}
```
## NUT-02
`GetKeysetsResponse`
```json
{
  "keysets": [
    {
      "id": <keyset_id_hex_str>,
      "unit": <currency_unit_str>,
      "active": <bool>
    },
    ...
  ]
}
```
### `PostSwapRequest`
```json
{
  "inputs": <Proofs>, <-- RENAMED
  "outputs": <BlindedMessages>,
}
```
### `PostSwapResponse`
```json
{
  "signatures": <BlindedSignatures> <-- RENAMED
}
```
## NUT-04
`PostMintQuoteBolt11Request`
```json
{
  "amount": <int>,
  "unit": <str_enum["sat"]>
}
```
`PostMintQuoteBolt11Response`
```json
{
  "quote": <str>,
  "request": <str>,
  "paid": <bool>,
  "expiry": <int>
}
```
`PostMintBolt11Request`
```json
{
  "quote": <str>,
  "outputs": <Array[BlindedMessage]>
}
```
PostMintBolt11Response
```json
{
  "signatures": <Array[BlindSignature]> <-- RENAMED
}
```
## NUT-05
`PostMeltQuoteBolt11Request`
```json
{
  "request": <str>,
  "unit": <str_enum["sat"]>
}
```
`PostMeltQuoteBolt11Response`
```json
{
  "quote": <str>,
  "amount": <int>,
  "fee_reserve": <int>,
  "paid": <bool>,
  "expiry": <int>
}
```
`PostMeltBolt11Request`
```json
{
  "quote": <str>,
  "inputs": <Array[Proof]>
}
```
`PostMeltBolt11Response`
```json
{
  "paid": <bool>,
  "payment_preimage": <str>
}
```
</details>

# Quotes
We are also introducing `quotes`, a general way to register a mint and melt transaction with the mint that will work across different payment methods (bolt11, bolt12, on-chain, ...) and different currency units (sat, msat, usd, ...). Quotes add the ability for Lightning backends to decide the amount and currency of the ecash they need in order to receive or pay a Lightning payment. The flow remains very similar to before and will be illustrated with two examples in the following. 

## Mint Quotes 
To mint ecash (output) via Lightning (input), the wallet first requests a `MintQuote` for a given amount of sats the wallet wants to mint. The `MintQuote` has an `quote` id and includes a Lightning invoice. The user pays the Lightning invoice and then calls `/v1/mint` referencing the previous `quote` it corresponds to.

## Melt Quotes
To melt ecash (input) and make a Lightning payment (output), the wallet requests a `MeltQuote` for a given Lightning invoice it likes to pay. In the `MeltQuote`, the mint tells the wallet how many sats it needs to supply and what the fee reserve is in order for the mint to fulfill this request.

## Diagram
image

# Hexadecimal keyset IDs
Our keyset IDs are ugly and need special treatment for HTTP (base64 urlsafe). We switch to **hexadecimal keyset IDs** that are generated much like the previous ones. We also add a **version byte as a prefix.** 

```
1 - sort public keys by their amount in ascending order
2 - concatenate all public keys to one string
3 - HASH_SHA256 the concatenated public keys
4 - take the first 16 characters of the hex-encoded hash
5 - prefix it with a keyset ID version byte
```

An example implementation in Python:

```python
def derive_keyset_id(keys: Dict[int, PublicKey]) -> str:
	"""Deterministic derivation keyset_id from set of public keys."""
	pubkeys_concat = "".join([p.serialize().hex() for _, p in sorted_keys.items()])
	return "00" + hashlib.sha256(pubkeys_concat.encode("utf-8")).hexdigest()[:14]
```

# `BlindedMessage` now has keyset `id` field
Outputs (`BlindedMessages`), now also have a keyset `id` field, like `Proofs` (inputs) and `BlindSignatures` do. With the `id`, the wallet tells the mint which keyset the client is expecting a signature from during a `/v1/swap` (created outputs), `/v1/melt` (change outputs), or `/v1/mint` (minted outputs). The requested `id` *MUST* be from an `active` keyset (part of the `/v1/keys` response and the `/v1/keysets` response). If the wallet uses an `id` that is not existent or not active (rotated-out of), the mint *MUST* refuse the transaction. 

The `BlindedMessage` becomes
```json
{
  "amount": int,
  "id": str # <-- NEW!
  "B_": hex_str
}
```
<details>
  <summary>Click to show additional remarks</summary>

### Additional remarks
Adding an `id` fixes a race condition we previously created workarounds for (see https://github.com/cashubtc/cashu-ts/pull/64 for example). Essentially, the mint's keys could have rotated between the wallet sending the outputs to sign to the mint, and the mint responding with a signature. We can now get rid of this code by making it part of the protocol.

Another critical issue that results from the same race condition is deterministic secret derivation: If a wallet deterministically derives secrets for keyset A, sends `BlindedMessages` to the mint and the mint rotated keys in the mean time, it would respond with `BlindedSignatures` keyset B. That means the wallet has incremented its deterministic secret derivation counter on the wrong keyset ID.

</details>

### Standardized secrets
Wallets should use standardized secrets (32 bytes of randomness) in lowercase hex. Closes https://github.com/cashubtc/nuts/issues/54


01.md
```diff
@@ -35,11 +35,33 @@ Response of `Bob`:
...
 {
-  "1": "03a40f20667ed53513075dc51e715ff2046cad64eb68960632269ba7f0210e38bc",
-  "2": "03fd4ce5a16b65576145949e6f99f445f8249fee17c606b688b504a849cdc452de",
-  "4": "02648eccfa4c026960966276fa5a4cae46ce0fd432211a4f449bf84f13aa5f8303",
-  "8": "02fdfd6796bfeac490cbee12f778f867f0a2c68f6508d17c649759ea0dc3547528",
-  ...
+  "keysets": [
+    {
+      "id": <keyset_id_str>,
+      "symbol": <currency_symbol_str>,
```
dipunm, line 41: 👀

02.md
```diff
@@ -56,7 +64,7 @@ Here, `MAX_ORDER` refers to the order of the maximum token value that the mint s
 
 ## Keyset ID
 
-A keyset ID is an identifier for a specific keyset. It can be derived by anyone who knows the set of public keys of a mint. The keyset ID **CAN** be stored in a Cashu [token][TokenV3] such that the token can be used to identify which mint or keyset it was generated from. 
+A keyset ID is an identifier for a specific keyset. It can be derived by anyone who knows the set of public keys of a mint. The keyset ID is stored in each `Proof` and therefore in a [token][TokenV3] such that the token can be used to identify which mint or keyset it was generated from.
```
dipunm, line 67: This would imply that a single token can have proofs from multiple keysets.

Is there a use case for this? We could reduce the payload if we don't repeat the keyset id in a response.

While we're refactoring you know..... 😅

00.md
```diff
@@ -86,85 +86,11 @@ An array (list) of `Proof`s. In general, this will be used for most operations i
...
-
-#### 0.2.3 - V3 tokens
+#### 0.2.1 - V3 tokens
 
 `mandatory`
 
-This new token format is easier to parse and to implement for wallet developers. It introduces a new token prefix, a versioning flag, and a URI prefix for making tokens clickable on the web.
-
-**Note:** This format deprecates the V2 token since it achieves the same thing as a V2 token but is less complex to implement.
+This new token format introduces a new token prefix, a versioning flag, and a URI prefix for making tokens clickable on the web.
```
thesimplekid, line 93: Since the old tokens are removed from the nut I think saying new is confusing.

03.md
```diff
@@ -7,7 +7,7 @@ NUT-03: Request mint
 
 Minting tokens is a two-step process: requesting a mint and minting the tokens. Here, we describe the first step. A wallet requests the minting of tokens in exchange for paying a bolt11 Lightning invoice (typically generated by the mint to add funds to its reserves, and typically paid with another Lightning wallet).
 
-To request the minting of tokens, a wallet `Alice` sends a `GET /mint?amount=<amount_sat>` request with the requested amount `<amount_sat>` in satoshis. The mint `Bob` then responds with a Lightning invoice.
+To request the minting of tokens, a wallet `Alice` sends a `GET /mint?amount=<amount>` request with the requested amount `<amount>` in satoshis. The mint `Bob` then responds with a Lightning invoice.
```
thesimplekid, line 10: > with the requested amount `<amount>` in satoshis.

Since keyset is being changed to support multiple units will this request always be in satoshis?

dipunm: Being able to specify unit is interesting... it can allow denominations of millisats and maybe even (don't shoot me) bits.

02.md
callebtc: Yes, this is possible according to the TokenV3 encoding. Some wallets use this to export their entire balance from multiple mints with a single token. Not sure how useful that is, something we might deprecate for a next encoding version.

03.md
callebtc: > will this request always be in satoshis?

The new `GET /v1/mint` allows wallets to specify `amount` (example: `1234`), currency `symbol` (example: `USD`), and payment method `method` (example: `bolt11`) in the request body.

thesimplekid: > requested amount `<amount>` in satoshis.

In that case "in satoshis" should be removed I think.

01.md
```diff
@@ -35,11 +35,33 @@ Response of `Bob`:
...
+    "keys": {
+        "1": "03a40f20667ed53513075dc51e715ff2046cad64eb68960632269ba7f0210e38bc",
+        "2": "03fd4ce5a16b65576145949e6f99f445f8249fee17c606b688b504a849cdc452de",
+        "4": "02648eccfa4c026960966276fa5a4cae46ce0fd432211a4f449bf84f13aa5f8303",
+        "8": "02fdfd6796bfeac490cbee12f778f867f0a2c68f6508d17c649759ea0dc3547528",
+        ...
+    }
+  ]
 }
 ```
```
thunderbiscuit, line 67: This PR is called the great cleanup so I'll point this out: the files include links that are not valid, and I think that could be cleaned up here. My preference would be to add links at the bottom as needed in the document above instead of pre-emptively adding an arbitrary number (20) of them, for which many are non-existent NUTs.

02.md
thunderbiscuit: I agree that the "multiple mints per token" makes things more complicated than they could be, particularly because the use case for it is unclear. If you export your entire balance from multiple mints in a single token... at this point you're just exporting a data bundle; I think the word token looses its meaning and strength if it is too wide of a definition. Just my 2 cents. But like is stated above, this doesn't need to be solved here of course. Might just be something to think about for future refactorings.

02.md
```diff
@@ -56,72 +62,77 @@ Here, `MAX_ORDER` refers to the order of the maximum token value that the mint s
...
 ### Storing the keyset ID in a token
 
 A wallet can use the keyset ID in a token to recognize a mint it was issued by. For example, a wallet might store the `MINT_URL` together with the `keyset_id` in its database the first time it receives a keyset from that mint. That way, a wallet can know which mint to contact when it receives a token with a `keyset_id` of a mint that it has interacted with before.
 
 Note: V3 tokens (see [NUT-00][00]) include the `MINT_URL` to enable the first contact when a wallet receives a token from a mint it has never met before.
 
+### Keyset ID version
+
+Keyset IDs have a version byte (two hexadecimal characters). The currently used version byte is `00`. 
+
```
thunderbiscuit, line 76: I don't think the examples in other files are showcasing this version byte (they don't start with `00`). I think maybe adding a line as to why this version byte is useful might be good. Or you could just state it in the PR maybe so that we have a reference for later as per the thinking behind it.

02.md
```diff
@@ -56,72 +62,77 @@ Here, `MAX_ORDER` refers to the order of the maximum token value that the mint s
...
 
 ```
-1 - sort keyset by amount
-2 - concatenate all (sorted) public keys to one string
+1 - sort public keys by their amount in ascending order
+2 - concatenate all public keys to one string
 3 - HASH_SHA256 the concatenated public keys
-4 - take the first 12 characters of the hash
+4 - take the first 16 characters of the hex-encoded hash
+5 - prefix it with a keyset ID version byte
```
thunderbiscuit, line 86: 1. Are we hashing the string of numbers and letters or the byte array here? 
2. This reminds me that this PR should also udpate the test vectors for the small breaking changes like this one.

My preference would be to hash the concatenated bytes of all public keys. That way it's all bytes until the very end, where you pull the first 8 bytes of the hash and use their hex-encoded format.

02.md
```diff
@@ -56,72 +62,77 @@ Here, `MAX_ORDER` refers to the order of the maximum token value that the mint s
...
 ```python
-def derive_keyset_id(keys: Dict[int, PublicKey]):
+def derive_keyset_id(keys: Dict[int, PublicKey]) -> str:
 	"""Deterministic derivation keyset_id from set of public keys."""
 	pubkeys_concat = "".join([p.serialize().hex() for _, p in sorted_keys.items()])
-	return base64.b64encode(
-	    hashlib.sha256(pubkeys_concat.encode("utf-8")).digest()
-	).decode()[:12]
+	return "00" + hashlib.sha256(pubkeys_concat.encode("utf-8")).hexdigest()[:14]
```
thunderbiscuit, line 96: Should this be `.hexdigest()[:16]` instead?

02.md
```diff
@@ -56,72 +62,77 @@ Here, `MAX_ORDER` refers to the order of the maximum token value that the mint s
...
 ```http
-GET https://mint.host:3338/keys/L3zxxRB_I8uE
+GET https://mint.host:3338/keys/9bb9d58392cd823e
 ```
 
 With curl:
 
 ```bash
-curl -X GET https://mint.host:3338/keys/L3zxxRB_I8uE
+curl -X GET https://mint.host:3338/keys/9bb9d58392cd823e
```
thunderbiscuit, line 115: Should these use the `v1/` paths? Wondering if communication with the mint using the old non-v1 paths are expected to stay with the base64 and only the v1 paths are expected to migrate to the new keyset ids.

02.md
```diff
@@ -56,72 +62,77 @@ Here, `MAX_ORDER` refers to the order of the maximum token value that the mint s
...
+    "symbol": "sat",
+    "keys": {
+        "1": "03a40f20667ed53513075dc51e715ff2046cad64eb68960632269ba7f0210e38bc",
+        "2": "03fd4ce5a16b65576145949e6f99f445f8249fee17c606b688b504a849cdc452de",
+        "4": "02648eccfa4c026960966276fa5a4cae46ce0fd432211a4f449bf84f13aa5f8303",
+        "8": "02fdfd6796bfeac490cbee12f778f867f0a2c68f6508d17c649759ea0dc3547528",
+        ...
+    }
+  ]
 }
```
thunderbiscuit, line 133: Because this is the response to the `keys/<keysetid>` GET request, I expected this response to not be an array of `"keysets"` since there can only be one (IIUC). Is there a reason for it to be that way? I'm thinking maybe this would work just as well:

```json
{
  "id": "9bb9d58392cd823e",
  "symbol": "sat",
  "keys": {
    "1": "03a40f20667ed53513075dc51e715ff2046cad64eb68960632269ba7f0210e38bc",
    "2": "03fd4ce5a16b65576145949e6f99f445f8249fee17c606b688b504a849cdc452de",
    "4": "02648eccfa4c026960966276fa5a4cae46ce0fd432211a4f449bf84f13aa5f8303",
    "8": "02fdfd6796bfeac490cbee12f778f867f0a2c68f6508d17c649759ea0dc3547528",
    ...
  }
}
```

03.md
```diff
@@ -7,7 +7,7 @@ NUT-03: Request mint
...
 
 ## Example
```
thunderbiscuit, line 12: I think it would be good to add an example of the v1 request body.

03.md
```diff
@@ -27,14 +27,20 @@ Response of `Bob`:
...
+  "request": "lnbc100n1p3kdrv5sp5lpdxzghe5j67q...",
+  "method": "bolt11",
+  "symbol": "sat",
+  "amount": 1000,
+  "id": "67d1d9ea6ada225c115418671b64a..."
 }
 ```
 
-with `pr` being the bolt11 payment request and `hash` is a **random** hash generated by the mint to internally look up the invoice state. Note: `hash` **MUST NOT** be the bolt11 invoice hash as it needs to be a secret between mint and user. This is to prevent a third party who knows the bolt11 invoice from stealing the tokens when the invoice is paid (See [NUT-04][04]). `hash` **MUST** be url-safe since we use it in a URL as described in [NUT-04][04].
+with `method` being the payment method, `request` being the payment request and `id` is a **unique and random** id generated by the mint to internally look up the payment state. Note: `id` **MUST** remain a secret between user and mint and **MUST NOT** be derivable from the payment request or otherwise a third party who knows the payment request can stealing the tokens that this operation mints (See [NUT-04][04]).
```
thunderbiscuit, line 38: Typo: ~stealing~ _steal_.

03.md
```diff
@@ -27,14 +27,20 @@ Response of `Bob`:
...
 
-A wallet **MUST** store the `hash` and `amount_sat` in its database to later request the tokens upon paying the invoice. A wallet **SHOULD** then present the payment request (for example via QR code) to the user such that they can pay the invoice with another Lightning wallet. After the user has paid the invoice, a wallet **MUST** continue with [NUT-04][04] (minting tokens).
+A wallet **MUST** store the `id` and `amount` in its database to later request the tokens upon paying the invoice. A wallet **SHOULD** then present the payment request (for example via QR code) to the user such that they can pay the invoice with another Lightning wallet. After the user has paid the invoice, a wallet **MUST** continue with [NUT-04][04] (minting tokens).
+
+### Payment methods
+Currently,
```
thunderbiscuit, line 43: I think this part was forgotten. Did you mean to finish this or remove it?

04.md
```diff
@@ -16,7 +16,7 @@ Note: In order to preserve privacy around the amount that a client might want to
 Request of `Alice`:
 
 ```http
-POST https://mint.host:3338/mint?hash=67d1d9ea6ada225c115418671b64a
+POST https://mint.host:3338/mint?id=67d1d9ea6ada225c115418671b64a
```
thunderbiscuit, line 19: This might be a v1 path instead of the old one. Or does that change apply to the old requests as well?

04.md
```diff
@@ -57,12 +57,12 @@ If the invoice was successfully paid, `Bob` responds with a `PostMintResponse` w
 "promises":
   [
     {
-    "id": "DSAl9nvvyfva",
+    "id": "9bb9d58392cd823e",
```
thunderbiscuit, line 60: Again just a reminder that the keyset ids don't have the `00` prefix mentioned in NUT-02.

thunderbiscuit (commented):
A few overall comments: 
- The PR description states that the new v1 routes are to be used and the old ones are deprecated. Not sure how to handle that elegantly in the NUTs. Should examples for both be given? Or all examples be migrated to the v1? If so, most of those are not yet updated.
- Breaking changes means we should update the test vectors 
- How should tokens handle v1? Are there corner cases there that could create issues? What if you bundle a token with a ton of inputs, some v1 and some not? Haven't thought about this deeply, just writing down thoughts as they come to mind...

Overall some great improvements in there (also lots of new code to write for us! 😅). I suggest that if there is intention to eventually move forward with a solution to #54, it should be included here as part of this fairly wide-ranging set of breaking changes.

thunderbiscuit:
Great discussions over the dev call the other day. I figured I'd add a small todo list of things that might need to be addressed here before I forget. Note that not all of those might be required, and I'm going off of memory so hopefully not missing anything.

- [ ] Add note on standardized secrets (32 bytes of randomness), most likely transmitted as lowercase hex and why wallet should strive to follow the standard. Closes #54
- [ ] Make sure the `/info/` endpoint is future-proof and potentially adds messaging around breaking changes and supported features
- [ ] Consider the idea of using different endpoints for different units (additional NUTs for additional units would define new endpoints). If so, decide how to handle the default bolt11 `mint/` and `melt/` endpoints and their definition in the NUTs (are they good as is or do they now require their own spec file?)
- [ ] Add `unit` field to `Proof` object, requiring a new token format version (`cashuB`)
- [ ] Update all examples to the `v1` routes
- [ ] Address comments from reviewers above

02.md
thesimplekid: > My preference would be to hash the concatenated bytes of all public keys. That way it's all bytes until the very end, where you pull the first 8 bytes of the hash and use their hex-encoded format.

This is my preference as well but the example code below it looks like it is concatenating the string of hex encoded pubkey. 

> This reminds me that this PR should also udpate the test vectors for the small breaking changes like this one.

+1

02.md
```diff
@@ -56,72 +62,76 @@ Here, `MAX_ORDER` refers to the order of the maximum token value that the mint s
...
 def derive_keyset_id(keys: Dict[int, PublicKey]):
-	"""Deterministic derivation keyset_id from set of public keys."""
-	sorted_keys = dict(sorted(keys.items()))
-	pubkeys_concat = "".join([p.serialize().hex() for _, p in sorted_keys.items()])
-	return base64.b64encode(
-	    hashlib.sha256(pubkeys_concat.encode("utf-8")).digest()
-	).decode()[:12]
+    sorted_keys = dict(sorted(keys.items()))
+    pubkeys_concat = b"".join([p.serialize() for _, p in sorted_keys.items()])
+    return "00" + hashlib.sha256(pubkeys_concat).hexdigest()[:14]
```
thesimplekid, line 95: @thunderbiscuit commented this previously but this it was removed when the change to bytes was pushed. Step 4 states to take the first 14 characters of the hex encoded hash but the example implementation seems to take the first 14.

02.md
thesimplekid: This is my understanding as well, and I would prefer this response unless there is a reason for the list.

04.md
```diff
@@ -83,7 +83,7 @@ A list of `Proof`s is called `Proofs` and has the form:
 "proofs" :
```
thesimplekid, line 83: Can't comment on the correct line as it wasn't changed in this PR, but on line 77 [NUT-0] should be [NUT-00}

02.md
```diff
@@ -5,28 +5,28 @@ NUT-02: Keysets and keyset ID
 
 ---
 
-A keyset is a set of public keys that the mint `Bob` generates and shares with its users. The mint **MUST** use the [compressed public key format](https://learnmeabitcoin.com/technical/public-key#public-key-format). The keyset refers to the set of public keys that each correspond to the amount values that the mint supports (e.g. 1, 2, 4, 8, ...) respectively.
+A keyset is a set of public keys that the mint `Bob` generates and shares with its users. It refers to the set of public keys that each correspond to the amount values that the mint supports (e.g. `1, 2, 4, 8, ...`) respectively.
 
 ## Requesting mint keyset IDs
 
-A mint can have multiple keysets at the same time but **MUST** have only one *current* keyset (see [NUT-01][01]). Tokens from all other keysets are still accepted but must be redeemed for tokens from only the active keyset.
+A mint can have multiple keysets at the same time but **MUST** have at least one `active` keyset (see [NUT-01][01]). Tokens from inactive keysets are still accepted but new tokens **MUST** be redeemed from `active` keysets.
```
thesimplekid: Should a mint have only one active key set per unit?

thesimplekid: Should a mint have only one active key set per unit?

callebtc:
I've completed most of the Todos! I would suggest deferring these remaining suggestions to later since the cleanup in this round focusses more on breaking changes in the API.
> * Make sure the `/info/` endpoint is future-proof and potentially adds messaging around breaking changes and supported features

I think these can be added as we go. What we should probably do already now is to indicate which (method, unit) pairs a mint supports.

> * [ ]  Add `unit` field to `Proof` object, requiring a new token format version (`cashuB`)

We can do it in cashuA and add a `unit` field in the JSON without breaking the format
> * [ ]  Potentially add errors enum with codes

I would love this but I'm afraid it needs to be in another PR


Thanks for the detailed comments and many errors you've found!

05.md
```diff
@@ -5,104 +5,133 @@ NUT-05: Melting tokens
...
 
-## Checking Lightning fees
+In the second request, the wallet includes the `quote` id and provides `inputs` that sum up to `amount+fee_reserve` in the first response. For the method `bolt11`, the wallet can also include `outputs` in order for the mint to return overpaid Lightning fees (see [NUT-08][08]). The mint responds with a payment status `paid` and a `proof` of payment. If the request included `outputs`, the mint may respond with `change` for the overpaid fees (see [NUT-08][08]).
+
+We limit this document to mint quotes of `unit="sat"` and `method="bolt11"` which requests a bolt11 Lightning payment (typically paid by the mint from its Bitcoin reserves).
+
+
+# Melt quote
 
 The mint expects `Alice` to include `Proofs` of *at least* `total_amount = amount + fee_reserve` where `amount` is the amount to be paid with the Lightning invoice and `fee_reserve` is the maximum potential Lightning network fees for the payment. A wallet should first call the `/checkfees` endpoint to ask the mint what the expected maximum fee is by making a `CheckFeesRequest` POST request:
```
thesimplekid, line 19: Is a call to the check fees endpoint still needed as the `fee_reserve` is included in the quote?

08.md
```diff
@@ -1,7 +1,7 @@
 NUT-08: Lightning fee return
```
thesimplekid, line 1: I wonder if this should be made more general. I think most payment methods with fees will work in a similar way where the request is over paid and the change is returned as in cashu

09.md
```diff
@@ -42,7 +40,7 @@ With the mint's response being of the form `GetInfoResponse`:
 ```
 
 - `name` is the name of the mint and should be recognizable. 
-- `pubkey` is the hex pubkey of the mint derived from its private key. 
+- `pubkey` is the hex pubkey of the mint.
```
thesimplekid, line 43: How is this derived?

01.md
```diff
@@ -9,37 +9,64 @@ This describes the basic exchange of the public mint keys that the wallet user `
...
+Wallet user `Alice` receives public keys from mint `Bob` via `GET /v1/keys`. The set of all public keys for a set of amounts is called a *keyset*.
 
-Mint `Bob` responds with his *active* [keyset][02]. The active keyset is the keyset a mint currently uses to sign promises with. The active keyset can change over time, for example due to key rotation. A mint MAY support older keysets indefinitely. Note that a mint can support multiple keysets at the same time but will only respond with the active keyset on the endpoint `GET /keys`. A wallet can ask for the keys of a specific (non-active) keyset by using the endpoint `GET /keys/{keyset_id}` (see #2 [TODO: Link #2]).
+The mint responds only with his `active` keysets. Keyset are `active` if the mint will sign promises with. The mint will accept tokens from inactive keysets as inputs but will not sign with them for new outputs. The `active` keysets can change over time, for example due to key rotation. A list of all keysets, active and inactive, can be requested separately (see [NUT-02][02]).
```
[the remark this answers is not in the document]
moonsettler: yep. more than one just splits the anon set.

special case could be multiple currencies in the future (even with sidechains it's technically a floating exchange rate)

moonsettler:
One general observation in any context that is not obviously a keyset structure naming the keyset_id id is not intuitive at all.
ksid or keyset or keyset_id would probably be more appropriate. the other field names are pretty self explanatory except for the single letter stuff.

00.md
```diff
@@ -213,27 +141,28 @@ This token format has the `[version]` value `A`. Here, `List[Proof]` is identica
...
           "amount": 2,
           "secret": "EhpennC9qB3iFlW8FZ_pZw",
           "C": "02c020067db727d586bc3183aecf97fcb800c3f4cc4759f69c626c9db5d8f5b5d4"
         },
         {
-          "id": "DSAl9nvvyfva",
+          "id": "009a1f293253e41e",
           "amount": 8,
           "secret": "TmS6Cv0YT5PU_5ATVKnukw",
           "C": "02ac910bef28cbe5d7325415d5c263026f15f9b967a079ca9779ab6e5c2db133a7"
```
thesimplekid, line 153: Think the example proofs here should be updated to the recommended 32 byte hex secret, even though that is not enforced.

03.md
```diff
@@ -1,40 +1,85 @@
...
 ```bash
-curl -X GET https://mint.host:3338/mint?amount=1000
+curl -X POST https://mint.host:3338/v1/split -d \
+{
+  "inputs": 
+    [
+      {
+        "id": "009a1f293253e41e",
+        "amount": 2,
+        "secret": "S+tDfc1Lfsrb06zaRdVTed6Izg",
```
thesimplekid, line 54: Same comment as above, secret should be the recommended 32 byte hex

ngutech21:
>Hexadecimal keyset IDs
...
> 4 - take the first 16 characters of the hex-encoded hash
The descriptions says to take the first 16 characters, but in the python implementation the first 14 characters are used.

02.md
```diff
@@ -5,123 +5,148 @@ NUT-02: Keysets and keyset ID
 
 ---
 
-A keyset is a set of public keys that the mint `Bob` generates and shares with its users. The mint **MUST** use the [compressed public key format](https://learnmeabitcoin.com/technical/public-key#public-key-format). The keyset refers to the set of public keys that each correspond to the amount values that the mint supports (e.g. 1, 2, 4, 8, ...) respectively.
+A keyset is a set of public keys that the mint `Bob` generates and shares with its users. It refers to the set of public keys that each correspond to the amount values that the mint supports (e.g. `1, 2, 4, 8, ...`) respectively.
```
ngutech21, line 8: Both endpoints /keys and /keysets are very similar. What was the general idea of having two endpoints that almost return the same data? Now /keys and /keysets both return `unit` and `id` and can therefore drift apart. In my opinion it would be easier to just have one endpoint that returns all key related data. If the client is only interested in active keysets, this could be accomplished by a query parameter. Having just one endpoint would be easier to maintain and result in less redundant code.

03.md
```diff
@@ -1,40 +1,85 @@
-NUT-03: Request mint
+NUT-06: Split tokens
 ==========================
 
 `mandatory` `author: calle`
 
 ---
 
-Minting tokens is a two-step process: requesting a mint and minting the tokens. Here, we describe the first step. A wallet requests the minting of tokens in exchange for paying a bolt11 Lightning invoice (typically generated by the mint to add funds to its reserves, and typically paid with another Lightning wallet).
+The split operation is the most important component of the Cashu system. A split operation consists of multiple inputs (`Proofs`) and outputs (`BlindedMessages`). Mints verify and invalidate the inputs and issue new promises (`BlindSignatures`). These are then used by the wallet to generate new `Proofs` (see [NUT-00][00]).
```
ngutech21, line 8: I think the term split doesn't fit well anymore since we got rid of the amount field. This endpoint should be renamed to /swap, because it is less specific. Not every call to /split is indeed a split, but it transforms the inputs to outputs.

ngutech21 (changes requested):

thunderbiscuit (commented):
Tons of great work in here. I left some more comments. I see you have a few more todo items in your PR description so I don't want to approve it before you're actually done, but this is looking good IMO.

One thing I _think_ doesn't cause problem but I want to make sure I ask: are the v1 routes in any way impacting the way the tokens were before and are now built? As in, I don't _think_ mixing up proofs from old routes and proofs and new one from this newer cashu workflow can cause trouble inside a single cashu token, but can you confirm this to be true?

Lastly, I don't know where this might fit but there is a high-level mental model that took me a few weeks to really grasp I don't know why, and that was the key for me to not get lost in the sauce with all that new jargon/verbiage. Here is how I would write it (again don't know where it fits or if it's even needed, maybe just split into parts at the top of the split/melt/mint files?)

The Cashu protocol defines 3 types of interactions that can happen between a client and a mint, where the client can exchange:
1. ecash tokens for ecash tokens (called a **split** operation)
2. bitcoin for ecash tokens (called a **mint** operation)
3. ecash tokens for bitcoin (called a **melt** operation)

05.md
```diff
@@ -5,104 +5,133 @@ NUT-05: Melting tokens
...
+  "quote": <str>,
+  "inputs": <Array[Proof]>
 }
 ```
 
-Here, `pr` is the bolt11 invoice to be paid with an amount `amount` and `proofs` are the proofs with a total value of at least `total_amount := amount + fee_reserve` (see above).
+Here, `pr` is the bolt11 invoice to be paid with an amount `amount` and `proofs` are the proofs with a total amount of at least `amount + fee_reserve` (see previous melt quote response).
 
-With curl:
+The mint `Bob` then responds with a `PostMeltBolt11Response`:
```
ngutech21, line 89: How does the mint return overpaid fees? I think we should add an additional `change` field to the response like we did before the cleanup

05.md
```diff
@@ -5,104 +5,133 @@ NUT-05: Melting tokens
...
 
-```bash
-curl -X POST https://mint.host:3338/melt -d \
+```json
 {
-"proofs" : 
-  [
+  "paid": <bool>,
+  "proof": <str>
```
ngutech21, line 94: The term proof is ambigous in this context, because it doesn't refer to a cashu Proof but a bolt11 payment preimage. `payment_preimage` might be a better name since this is a specific bolt11 response we can use the lightning terminology.

05.md
```diff
@@ -5,104 +5,133 @@ NUT-05: Melting tokens
...
+}
+```
+`paid` is a boolean indicating whether the payment was successful, and `proof` being the bol11 payment preimage.
```
ngutech21, line 97: typo: bolt11 payment preimage

11.md
```diff
@@ -121,7 +121,7 @@ If the `locktime` is in the past and a tag `refund` is present, the `Proof` is s
 
 #### Complex Example
 
-This is an example `secret` that locks a `Proof` with a Pay-to-Pubkey (P2PK) condition that requires 2-of-3 signatures from the first public key in the `data` field, and the second public key in the `pubkeys` tag. If the `timelock` is passed, the `Proof` becomes spendable with a single signature from the public key in the `refund` tag. The signature flag `sigflag` indicates that signatures are necessary on the `Proofs` and the `BlindMessages` of a transaction.
+This is an example `secret` that locks a `Proof` with a Pay-to-Pubkey (P2PK) condition that requires 2-of-3 signatures from the public keys in the `data` field amd the `pubkeys` tag. If the `timelock` has passed, the `Proof` becomes spendable with a single signature from the public key in the `refund` tag. The signature flag `sigflag` indicates that signatures are necessary on the `inputs` and the `outputs` of a transaction.
```
thesimplekid, line 124: > field amd

typo: and

12.md
```diff
@@ -53,26 +53,26 @@ The mint produces these DLEQ proofs when returning `BlindedSignature`'s in the r
...
   "secret": <str>,
   "C": <str>,
-  "dleq": {# New: DLEQ proof
+  "dleq": { <-- New: DLEQ proof
     "e": <str>,
     "s": <str>,
     "r": <str>
   }
 }
 ```
```
thesimplekid, line 70: Just to stay consistent `json`

callebtc:
All Todos are now closed and with that the PR is officially ready for review (and thanks for all reviews already 🙏)

Rocket emoji!

thunderbiscuit (approved):
ACK e59f284f0b786eb7b239715469f9e0dd9cd4b6ea. 

Major improvement of the protocol. I'm happy to see that we have a few implementations that were able to put this into code already, and looking forward to seeing what gets built on top of this. 🚀

ngutech21 (approved):
ACK e1568fdadc9e3e99d6a071e1de17a8a79dd658bf

This is a great improvement of the cashu protocol.

00.md
```diff
@@ -36,145 +38,84 @@ NUT-00: Notation and Models
...
 
-A signature on the `BlindedMessage` is sent from `Bob` to `Alice` after [minting tokens][04] or after [splitting tokens][06]. A `BlindedSignature` is also called a `promise`.
+A `BlindedSignature` is sent from `Bob` to `Alice` after [minting tokens][04] or after [swapping tokens][03]. A `BlindedSignature` is also called a `promise`.
 
 ```json
 {
   "amount": int,
   "C_": hex_str,
-  "id": str | None
+  "id": hex_str
```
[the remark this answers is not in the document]
thunderbiscuit: I was wrong, it does break in small ways the tests (not the validity of the tokens but their serialization).

For example note the example at the end of NUT-00 which has the order "id, amount, secret, C" (different from the model for `Proof` described above):
```json
"proofs": [
    {
        "id": "009a1f293253e41e",
        "amount": 2,
        "secret": "407915bc212be61a77e3e6d2aeb4c727980bda51cd06a6afc29e2861768a7837",
        "C": "02bc9097997d81afb2cc7346b5e4345a9346bd2a506eb7958598a72f0cf85163ea"
    },
    {
        "id": "009a1f293253e41e",
        "amount": 8,
        "secret": "fe15109314e61d7756b0f8ee0f23a624acaa3f4e042f61433c728c7057b931be",
        "C": "029e8e5050b890a7d6c0968db16bc1d5d5fa040ea1de284f6ec69d61299f671059"
    }
]
```

This means that a library that implemented the order described in the model section:
```json
{
  "amount": int, 
  "id": hex_str,
  "secret": str,
  "C": hex_str,
}
```

Will not be able to reproduce your resulting serialization in base64 (even though both tokens would be valid).

I suggest the order be "id, amount, secret, C" for both the model and the example.

07.md
```diff
@@ -43,21 +43,19 @@ With the data being of the form `CheckStateRequest`:
 }
 ```
 
-`Proofs` is a list (array) of `Proof`s (see [NUT-0][00]). `Alice` **CAN** provide a full `Proof` but **MUST** provide at least the `secret` (which is the only thing that `Bob` needs to check whether the token has been spent).
+`Proofs` is a list (array) of `Proof`s (see [NUT-00][00]). `Alice` **CAN** provide a full `Proof` but **MUST** provide at least the `secret` (which is the only thing that `Bob` needs to check whether the token has been spent).
```
thesimplekid, line 46: Can this be used by the mint to finger print a wallet? If some wallets are checking by sending the full proof and others are only sending the secret? Should a recommendation be made to do one or the other similar to the ordering of the proofs to avoid fingerprinting?

06.md
```diff
@@ -1,87 +1,72 @@
...
+    "5": {
+      "methods": [
+        ["bolt11", "sat"]
+      ]
+    },
+    "7": {"supported": true},
+    "8": {"supported": true},
+    "9": {"supported": true},
+    "10": {"supported": true},
+    "12": {"supported": true},
```
thesimplekid, line 49: The trailing comma should be removed.

PR #60, opened 2023-10-19
fix  a typo for nut08

biluohc:


thunderbiscuit (approved):
ACK 998cb8de3a11b6bb3f739a6ac3c9127b6b02ac53. I had noted the same thing a few weeks ago.

We should make sure this change is disseminated widely however, because it is a breaking change in the serialization of the `PostMeltRequest` object.

callebtc:
Thanks for the PR but we can't merge this. This isn't a typo but renaming a field and would break the protocol.

@thunderbiscuit is right that we want to rename this field to a more generic name `request` which can harbor lightning invoices, on-chain addresses, or whatever else we might encounter along the way.

biluohc:
https://github.com/cashubtc/nuts/blob/main/08.md#example-1

So what is `Everything here is the same as in [NUT-05](https://github.com/cashubtc/nuts/blob/main/05.md) except for outputs.` ?

thesimplekid:
> So what is Everything here is the same as in [NUT-05](https://github.com/cashubtc/nuts/blob/main/05.md) except for outputs. ?

Yes, it allows for change to be returned if the routing fee is over paid.

PR #66, opened 2023-12-19
V1 small fixes

thunderbiscuit:
This PR simply cleans up some of the examples and field names/types, and does not introduce breaking changes.

1. In some of the type descriptions, type annotations for fields that contain arrays/lists is done as such: `<Array[BlindedMessage]>`. This is explicit and easy to parse, and was not applied to all type descriptions. I think I refactored them all here (for example NUT-03 currently uses this notation: `"inputs": <Proofs>, "outputs": <BlindedMessages>` which is less clear).
2. In NUT-03, the example for `PostSwapRequest` has proofs with fields in the wrong order according to the model for `Proof` in NUT-00 (`amount` comes before `id`). Same for NUT-05 and NUT-08.
3. In NUT-03, the example for `PostSwapRequest`'s `outputs` field has a blinded message with missing field `id` according to the new model for `BlindedMessage` in NUT-00.
4. The spec used `BlindSignature` and `BlindedSignature` interchangeably. These have now all be streamlined to `BlindSignature`.
5. In NUT-08, acknowledge the fact that the `payment_preimage` field in the `PostMeltBolt11Response` can be null.


callebtc (approved):
LGTM!

PR #69, opened 2023-12-28
Unify notation on NUT 07

gandlafbtc:


thunderbiscuit (approved):
ACK 3d8e2e35355c2c9b41fc8ba5bf801ff5aea5b043.

Note that the notation mismatch happens in a few other files as well and is fixed in #66 ([this commit]( https://github.com/cashubtc/nuts/pull/66/commits/255415d0f81b6ca7faa66fe3131c9b794d85940d) more particularly) but I had not fixed the wording of the sentence under it; once you fix the notation I think that removing the sentence indeed makes things cleaner 👍, and thus this is not a duplicate of #66 and ready to merge IMO.

callebtc (approved):

PR #70, opened 2023-12-29
Add Array[...] to model description

gandlafbtc:
I saw in most other nuts its notation is like that, let's change it here!


thesimplekid:
@thunderbiscuit has these included in #66

gandlafbtc:
Thanks, I'll close it then

PR #71, opened 2023-12-31
`/v1/checkstate`: new request and response models

callebtc:
### Request

`PostCheckStateRequest`:

```json
{
  "secrets": [
    <secret_str>,
    ...
  ]
}
```

### Response

Change from: 
```json
{
  "spendable": Array[bool],
  "pending": Array[bool]
}
```
to

New `PostCheckStateResponse` 

```json
{
  "states": [
    {
      "secret": <str>,
      "state": <str_enum[STATE]>,
      "witness": <str>
    },
    ...
  ]
}
```

`secret` now indicates which proof the returned data belongs to (provided by the client in the request), and `witness` is a new field that corresponds to the (potentially present) witness data when the proof was spent.

`state` is an enum string field:
```python
class SpentState(Enum):
    unspent = "UNSPENT"
    spent = "SPENT"
    pending = "PENDING"
```


thesimplekid:
ACK [80750b2](https://github.com/cashubtc/nuts/pull/71/commits/80750b272433b2210bee64c067c3693ba4c3eddd)

thesimplekid (approved):

07.md
```diff
@@ -65,14 +65,19 @@ curl -X POST https://mint.host:3338/v1/check -d \
...
+    },
+    ...
+  ]
 }
 ```
 
-Where `[bool]` is an array of booleans indicating whether the provided `Proof` is still spendable or pending. 
-
-**Important:** The list of booleans **MUST** be in the same order as the proofs provided by `Alice` in the request.
+Where `secret` belongs to the provided `Proof` to check in the request, `spendable` and `pending` are booleans indicating whether is still spendable or pending, and `witness` is the witness data that was potentially provided in a previous spend operation.
```
thunderbiscuit, line 80: missing word(s): `indicating whether is still spendable or pending` -> `indicating whether the proof is still spendable or pending`

thunderbiscuit (changes requested):
Small nit mentioned below.

I also approve of the idea of transforming `spendable` and `pending` into an enum as proposed by @callebtc, whether in this PR on another one.

07.md
callebtc: Changed to enum instead of two lists!

callebtc:
@thesimplekid @thunderbiscuit I decided to rename the endpoint to `/v1/checkstate` instead of `/v1/check` because it expresses better what we are checking. Are you ok with that?

thunderbiscuit (approved):
ACK. 094b5b5f68f0d6a9b3cc71820f6d748a9de040f6.

thesimplekid:
ACK. https://github.com/cashubtc/nuts/commit/094b5b5f68f0d6a9b3cc71820f6d748a9de040f6. Rename is good with me

Issue #61, opened 2023-10-29
NUT-11: Proof JSON seralization 


thesimplekid:
I am having trouble with the proof serialization in NUT-11, I do not think the example proof JSON in the NUT is valid. The `'` should be replaced with a `"` and the witness field is missing `{}`.

```json
{
   "amount":8,
   "secret":"[\"P2PK\", {\"nonce\": \"5d11913ee0f92fefdc82a6764fd2457a\", \"data\": \"026562efcfadc8e86d44da6a8adf80633d974302e62c850774db1fb36ff4cc7198\",}]",
   "C":"02250a37a56b78e66674f7f063e6abd3d9345f8761fb90cac0293108910a8c27a3",
   "id":"1cCNIAZ2X/w1",
   "witness":[
      {
         "signatures":[
            "c43d0090be59340a6364dc1340876211f2173d6a21c391115adf097adb6ea0a3ddbe7fd81b4677281decc77be09c0359faa77416025130e487f8b9169eb0c609"
         ]
      }
   ]
}

```

https://github.com/cashubtc/nuts/blob/aefec2e80c108cbf794e592dc6ae3b8bdc3fc6f8/11.md?plain=1#L31-L45

thesimplekid:
https://github.com/cashubtc/nuts/pull/77#issue-2100323353

PR #74, opened 2024-01-19
Nut 11 fixes

thesimplekid:
* Witness encapsulated as an object 
* Escape inner quotes in secret
* Remove extra bracket on link to nuts
* Unify Array notation signatures
* Remove dead link to NUT09

TODO: 
* There is an extra trailing , in the secret that should likely be removed.
* The ID should be updated to a v1 keyset id.

Note: 
Going to leave this as a draft in case I find anything else while implementing. 

fixes: https://github.com/cashubtc/nuts/issues/61#issue-1966893496

### DIFF
--- a/11.md
+++ b/11.md
@@ -31,17 +31,17 @@ The recipient who owns the private key of the public key `Secret.data` can spend
 
 ```json
 {
-  "amount": 8, 
-  "secret": '["P2PK",{"nonce": "5d11913ee0f92fefdc82a6764fd2457a","data": "026562efcfadc8e86d44da6a8adf80633d974302e62c850774db1fb36ff4cc7198",}]',
-  "C": "02250a37a56b78e66674f7f063e6abd3d9345f8761fb90cac0293108910a8c27a3",
-  "id": "1cCNIAZ2X/w1",
-  "witness": {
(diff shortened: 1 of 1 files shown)


thunderbiscuit:
Is this one still a draft @thesimplekid? Changes look good to me, except that some are also implemented in #79. Happy to give final review once you're happy with it and set it as ready for review.

thesimplekid:
@thunderbiscuit Thanks for having a look. 

So the issue with adding the escape chars is it changes the secret, so the witness needs to be updated. I ran into this issue when implementing nostr nip57, where if the secret goes through any serialization or de-serialization, it wont match the original secret the signature is on, so the signature check needs to be on the original secret string. However we still need to be able to  de-serialize the secret and but this can be difficult if the json is malformed since the rust crate i use for json, serde, is quite strict with json formatting for example I can't de-serialize this secret because of the trailing comma. So I would prefer to be as strict as possible with json formatting at least in the spec and test-vectors. 

I'll rebase this to remove the duplicated fixes and leave the json formatting up for discussion, but it shouldn't be merged as is since the signature is not valid.

thesimplekid:
close for #96

PR #75, opened 2024-01-21
Create 09.md

gandlafbtc:
Work in progress for nut-09


trbouma:
I hope this is optional. For a mint to keep track of BlindedMessages it produces is a potential surveillance issue. It would be possible to correlate an isssuance event with a redemption event. I like the idea, but the alarm bells are ringing… I like the fact that a mint has no idea what is circulating, other than what it is liable for issuing.

callebtc:
> I hope this is optional.

It is.

> For a mint to keep track of BlindedMessages it produces is a potential surveillance issue. It would be possible to correlate an isssuance event with a redemption event.

That's not possible: the point of blind signatures is that the BlindedMessages can not be correlated to the signatures. 

We try to operate on the assumption that the mint is fully observable by anyone and privacy shouldn't rely on whether the mint chooses to log something or not.

09.md
```diff
@@ -0,0 +1,121 @@
...
+
+1. get `HDKey` from master seed
+2. construct the derivation path for both `secrets` and blinding factors (`r`)
+
+```javascript
+secret_derivation_path = `m/129372'/0'/${keysetID}'/${counter}'/0`
+r_derivation_path = `m/129372'/0'/${keysetID}'/${counter}'/1`
+```
+
+note that `m/129372'/0'/` is the standard derivation path for cashu wallets, and all wallets should use this path.
```
AngusP, line 47: haha nice

AngusP: Maybe this needs some clarification, if I was implementing this I'm not sure if I'd use `1'/` for testnet or still `0'/`? (I know you said BIP32 not BIP44 but still, idiot-proofing 😉)

```suggestion
Note that `m/129372'/0'/` is the standard derivation path for cashu wallets, and all wallets should use this path. Note the `0'/` is *not* a BIP-44 coin type indicating Bitcoin -- Cashu tokens in other denominations or currencies should still always use this path.
```

**--OR--** but probably not worth the extra complexity:

```suggestion
Note that `m/129372'/` is the standard derivation path for cashu wallets, and all wallets should use this path.

`m/129372'/0'/` is the standard for Bitcoin, `m/129372'/1'/` for testnet coins. Other `coin_type`s should respect the [SLIP-0044 : Registered coin types for BIP-0044](https://github.com/satoshilabs/slips/blob/master/slip-0044.md) values.
```

There is some ?mess that there are coins that don't have a SLIP-44 type (USDT is the most obvious one)

09.md
```diff
@@ -0,0 +1,121 @@
...
+[03]: 03.md
+[04]: 04.md
+[05]: 05.md
+[06]: 06.md
+[07]: 07.md
+[08]: 08.md
+[09]: 09.md
+[10]: 10.md
+[11]: 11.md
+[12]: 12.md
```
AngusP, line 121: These are unused?

09.md
gandlafbtc: thanks for the great suggestions! i think we can go with the simple one, since the "coin type"/ unit has to be defined in the keyset anyway.

09.md
gandlafbtc: yes, unused, Don't know why they are there tbh. Thanks!

callebtc:
Continued in #87

PR #78, opened 2024-01-25
NUT-00/NUT-11: clarify types for `secret` and signatures on inputs and outputs

callebtc:
NUT-00: Short clarification that `Proof.secret` is a UTF-8 string. The recommendation to use a hex string can be confusing since it may suggest that the byte data is signed (whereas we actually sign the string). 

NUT-11: Clarify that `Proof.secret` is signed as a string whereas `output.B_` is signed as bytes.


Egge21M:
I wonder if the protocol should be specific about the secret. For example dictate that secret SHOULD be either a random 32-byte hex-string or serialised JSON according to NUT-10, in order to avoid fingerprinting.

thesimplekid:
ACK [9799455](https://github.com/cashubtc/nuts/pull/78/commits/9799455132cbe975e3ec4d9313c3a29ab0c0f038) happy with this change or the stronger wording or `SHOULD` suggested by @Egge21M .  However, I think its better not to reference optional nuts like 10 in mandatory ones.

thesimplekid (approved):

Egge21M (approved):
LGTM

ngutech21 (approved):
ACK

gandlafbtc:
lgtm

PR #82, opened 2024-02-07
NUT-4/NUT-5: Method-unit pair is object with settings

callebtc:
This PR changes the "settings" for NUT-4/NUT-5 (minting and melting) that are shown in the `/v1/info` response.

The unit pairs are changed from an array `["method", "unit"]` to an object that now also carries settings for the endpoints.

Example:
```json
{
  "method": "bolt11",
  "unit": "sat",
  "min_amount": 0,
  "max_amount": 10000        
}
```

### DIFF
--- a/05.md
+++ b/05.md
@@ -178,6 +178,7 @@ The settings for this nut indicate the supported method-unit pairs for melting. 
   "max_amount": <int|null>
 }
 ```
+`min_amount` and `max_amount` indicate the minimum and maximum amount for an operation of this method-unit pair.
 
 Example `MeltMethodSetting`:
 
--- a/05.md
+++ b/05.md
@@ -157,14 +157,38 @@ Response of `Bob`:
 The settings for this nut indicate the supported method-unit pairs for melting. They are part of the info response of the mint ([NUT-06][06]) which in this case reads 
 ```json
 {
-  "5": {
+  "4": {
     "methods": [
-      ["bolt11", "sat"]
+      <MeltMethodSetting>,
+      ...
     ],
+    "disabled": <bool>
   }
 }
 ```
-Here, `methods` is an array of arrays with `[method, unit]` pairs.
+
+`MeltMethodSetting` indicates supported `method` and `unit` pairs and additional settings of the mint. `disabled` indicates whether this melting is disabled.
+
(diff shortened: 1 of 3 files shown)


ngutech21:
Update: In the nut the struct is described as settings but in the json it's called method. Having method or payment_method twice feels redundant. 

How about this:
```json
 "4": {
      "settings": [
        {
          "payment_method": "bolt11",
          "unit": "sat",
          "min_amount": 0,
          "max_amount": 10000        
        }
      ],
      "disabled": false
    },
```

05.md
```diff
@@ -157,14 +157,39 @@ Response of `Bob`:
 The settings for this nut indicate the supported method-unit pairs for melting. They are part of the info response of the mint ([NUT-06][06]) which in this case reads 
 ```json
 {
-  "5": {
+  "4": {
     "methods": [
-      ["bolt11", "sat"]
+      <MeltMethodSetting>,
```
thunderbiscuit, line 162: Again I'd call it an array of "settings" rather than an array of "setting", so the object IMO could be named `MeltMethodSettings`

thunderbiscuit (commented):
I think it makes sense to have a more complex object for settings, and might even be a place where mints can add custom things that are useful for them/their users without breaking anything, because the wallet parsing could simply discard any field it does not know (I don't do that in my lib yet and the parsing errors out if there are extra fields on the response object but might turn that on if it's useful for people).

A few comments below.

ngutech21:
Since cashu-ts is migrating to v1 api and this is only a small change, I think we should get this PR ready to merge before the wallets adopt the new api. Any thoughts on this? @gandlafbtc  @callebtc

gandlafbtc:
looks good to me

gandlafbtc (approved):

thesimplekid (approved):
Agree with tb's small points but overall LGTM

callebtc:
> In the nut the struct is described as settings but in the json it's called method. Having method or payment_method twice feels redundant.

It feels appropriate to me that there is an array "methods" with entries that have a "method". Since we speak of "method-unit" pairs in the spec, I think the labels are ok or at least clear enough as is.

I have added all remaining suggestions and I think we are ready to merge this.

PR #83, opened 2024-02-07
Add `unit` root-level field on the mint info response

thunderbiscuit:
This PR moves the `unit` field currently used in the individual nut settings to a root-level `unit` field. The info response would look roughly like this:

```json
 {
  "name": "Bob's Cashu mint",
  "pubkey": "0283bf290884eed3a7ca2663fc0260de2e2064d6b355ea13f98dec004b7a7ead99",
  "version": "Nutshell/0.15.0",
  "description": "The short mint description",
  "description_long": "A description that can be a long piece of text.",
  "unit": "sat"
  "nuts": {
    "4": {
      "methods": [
        "bolt11"
      ],
      "disabled": false
    },
    "5": {
      "methods": [
        "bolt11"
      ]
    },
  }
}
```

I think this makes sense for a few reasons:
1. It simplifies the object
2. It makes explicit the unit the mint issues ecash against
3. Differences between what you'd have at the root-level and the `nuts` settings level I think doesn't make sense:
    a. **NUT-04 (mint)**: it doesn't make sense to have the mint issue another unit than the one true unit of the mint; the keysets are associated with a given unit (a note for 8 has to be 8 "of something"). This something _must_ be the same as the overall unit of the mint defined at the root of the endpoint. You can't say "my keysets for [1, 2, 4, 8, 16] are good for satoshis, but I can mint (produce signatures with these keysets) for USD tokens.
    b. **NUT-05 (melt)**: it doesn't make sense to have the wallet pay with a different unit than the one associated with the keysets, because those are the ones that the mint signed for in the first place. It does make sense to ask the mint to pay invoices using different units/tokens, and that's captured by the `methods` field (`"bolt11"` in this case). For example you might spend your ecash (which is denominated in sat) and ask the mint to pay not a bolt11 invoice but rather an onchain payment, or a USD invoice of some kind. But the unit you'll use to "pay" for this melt will always be the unit of the tokens issued by the mint.
4. We could not allow discrepancies between these fields (say your nut.4 and nut.5 fields had different units), because that implies your keysets are not uniquely associated with a unit. So rather than forcing the check between the inner fields, it's easier to write a global/root-level field that captures that data in one place.

Let me know if you see something I've missed or if there was a reason to have those units defined individually in the nuts' settings.


thunderbiscuit:
Note that the PR also adds a requirement that the `PostMintQuoteBolt11Request` and `PostMeltQuoteBolt11Request` requests issued by the wallet use the same `unit` as the one in the info endpoint returned by the mint. This is a nice redundancy and ensures that the wallet really means to mint/melt the same tokens that the mint issues, but could also be considered superfluous. It's implied that the correct unit is known by the wallet because of the unit the mint issues (and returns on the info endpoint).

In any case I figured it provided enough value to keep it, but let me know if you think it's better to remove it.

ngutech21:
Why do you want to limit the mint to only issue one currency, when the cashu protocol offers the possibility to offer at least 2 different currencies (sat/usd) at the moment? We could easily add e.g. euro as another currency by writing a new nut and defining a new keyset-id prefix e.g. 02.

I had another look at the nuts and figured we didn't define the supported currencys and corresponding keyset-id prefixes. There should be a section in the nut-03 defining that the keyset-id prefix 00 == sat and keyset-id prefix 01 == usd.

thunderbiscuit:
I guess it depends on how you interpret the idea of a mint. My feeling is that it is cleaner to think of a mint as an entity that handles tokens tied to one specific unit. A single company/org/person could manage multiple mints (they could even be on subdomains of a single domain name for example), but a "mint" itself would only handle 1 unit. This makes it easier for wallets to handle, because they don't need to do a check on every call of the mint for the specific unit they mean to interact with. It also ensures that you only ever purchase a single "token type" from a mint (gets complicated if you can ask the mint to pay you half in usd and half in sats; instead, why not make two separate calls to the two "sub-mints"?)

The idea of using the first byte of the keyset as an identifier for the unit is something interesting we should discuss in an issue for sure. My only worry there is that you'd have a non-complete structure because people wanting to offer different units would somehow need to register themselves with the spec maybe? (see [BIP-44](https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki) and the registered coins at [SLIP-44](https://github.com/satoshilabs/slips/blob/master/slip-0044.md)). It gets a little weird, and in Cashu you'd have a lot of units that cannot be registered (say with the vouchers idea where units are specific and unique to a particular situation/app). How should those be handled? Anyway good discussion ahead I'm sure.

thunderbiscuit:
Note: I just noticed that this idea of a single-unit mint clashes with the NUT-02 section:

```md
#### Currency unit
Mints **MUST** generate a keysets for each `unit` they support.
```

I guess this PR is opening up a bigger discussion around what a mint "is"...

callebtc:
This PR conflicts with how we manage keysets. I understand that most applications will only require a single unit so I would suggest to simply run a mint that supports a single unit and signal that via the keysets. 

Closing because of stale discussion.

PR #88, opened 2024-03-04
Add a description for h2c

gudnuf:
I came across this addition to the spec in both nutshell, and in cashu-ts (v1). Maybe this is not ready to get added to the spec, but I figured I would submit a PR anyway for fun.

Also, when I first implemented the BDHKE flow a couple of months ago, the hash_to_curve function took me a while to figure out what it was. Maybe this is because I'm a noob, but I figured a clear explanation would be helpful for future nutty noobs.

### DIFF
--- a/00.md
+++ b/00.md
@@ -32,6 +32,16 @@ This document details the notation and models used throughout the specification 
 - `T` blinded message
 - `Z` proof (unblinded signature)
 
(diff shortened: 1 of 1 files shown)


AngusP (approved):

callebtc (approved):
Beautiful addition, thank you.

PR #90, opened 2024-03-05
NUT-14: Optional WebSocket Communications

Egge21M:
NUT-14 defines an optional communication layer for Mints and Apps that is built on WebSocket. It strives to be highly compatible and flexible.

I will flesh out the details of the next hours and day, but wanted to get this public as draft as quick as possible to get as much feedback as possible

### DIFF
--- a/14.md
+++ b/14.md
@@ -1,0 +1,39 @@
+# NUT-14: Optional WebSocket communication
+
+`optional`
+
+---
+
+This NUT describes an additional communication protocol between wallets / apps and mints based on WebSocket, that allows for live and bi-directional interaction.
+
+## Signaling Support via NUT-06
+
+NUT-14 is optional. Mints can signal support via NUT-06 and it's `/info` endpoint. Furthermore mints might choose to support only certain endpoints on the optional WebSocket protocol.
+The `nuts` key of `GetInfoResponse` for NUT-14 looks as follows17:
+
+```json
+"nuts": {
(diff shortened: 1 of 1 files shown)


gudnuf:
I've been thinking about using nostr wallet connect to communicate with a mint. For example, you could include tokens in an nwc payment request along with the invoice to pay. Nut-14 seems like it could lay the groundwork for getting mints to start using web sockets. 

Do you see the idea of adding nwc support to mints as a separate nut?
Or, what's the advantage to having a direct web socket connection to the mint vs. using nostr to communicate?

Egge21M:
> I've been thinking about using nostr wallet connect to communicate with a mint. For example, you could include tokens in an nwc payment request along with the invoice to pay. Nut-14 seems like it could lay the groundwork for getting mints to start using web sockets. 
> 
> 
> 
> Do you see the idea of adding nwc support to mints as a separate nut?
> 
> Or, what's the advantage to having a direct web socket connection to the mint vs. using nostr to communicate?

I love the idea of mints that support NWC, but I think this should be spec'd out else where. NUT-14 aims to provide a Cashu native communication layer, without any dependencies.

Nostr (and especially NWC) come with some overhead that is not necessary for basic, bidirectional communication between apps and mints.

Egge21M:
Closing this in favor of https://github.com/cashubtc/nuts/pull/98

Issue #91, opened 2024-03-06
NUT-07 – Allow token state check with `Y`


callebtc:
I would like to add a token spent state check using `Y` (`BlindedMessage`) instead of the `secret`. That way, someone other than the owner of a token can check the state of a token (if they know `Y`) without being able to spend it directly.

This could be added as an additional optional field in `PostCheckStateRequest` to the `/v1checkstate` endpoint  or as a new endpoint altogether. I tend towards the former option but I'm open for other preferences.

In the former case, `PostCheckStateRequest` would become

```json
{
  "secrets": [
    <secret_str>,
    ...
  ], 
  "Ys": [
    <Y_hex_str>,
    ...
  ] 
}
```

In that case, there would have to be a rule that states that either "secrets" OR "Ys" has to be populated for the request.

PR #93, opened 2024-03-07
NUT-07: Token state check with `Y`

callebtc:
Instead of using `secret` in the token state check, we use `Y`. 

This allows third parties to check the state of a token for which they have received the `Y` without being able to spend it. 

Wallets that know `secret` can compute `Y` by `Y = hash_to_curve(secret)`.

Closes: https://github.com/cashubtc/nuts/issues/91

### DIFF
--- a/07.md
+++ b/07.md
@@ -73,7 +73,7 @@ With curl:
 **Request** of `Alice`:
 
 ```bash
-curl -X POST https://mint.host:3338/v1/checkstate -H 'Content-Type: application/json' -d '{                                                                           cc@ddd
+curl -X POST https://mint.host:3338/v1/checkstate -H 'Content-Type: application/json' -d '{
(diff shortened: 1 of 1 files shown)


07.md
```diff
@@ -38,14 +38,14 @@ With the data being of the form `PostCheckStateRequest`:
 
 ```json
 {
-  "secrets": [
-    <secret_str>,
+  "Ys": [
+    <Array[hex_str]>,
     ...
   ]
 }
```
thesimplekid, line 45:
```suggestion
  "Ys": <Array[hex_str]>,
}
```
Bring notation inline with other nuts. [] brackets arent needed with the <Array[]>

callebtc: Thank you, didn't notice!

Issue #81, opened 2024-02-06
Discussion: consider removing libraries' links from spec


thunderbiscuit:
After speaking with @ngutech21 over a dev call last week, I decided to open an issue to discuss the value of the links to the libraries that implement the NUTs in the README.

My thinking is the following: links to individual libraries that implement the cashu protocol are important, but belong elsewhere than the spec (main website? awesome list? other places?). Here are a few reasons why I think over time it gets harder and harder to keep those in sync, and how they somehow complicate the commit history of the spec repo without reason.

1. It used to be just a few libraries, but now there are a lot more! It's a lot of links. Keeping them up to date is messy and likely to always be slightly behind.
2. Some of the libraries in the list _did_ implement certain NUTs, but might not have updated to the v1 endpoints. What does that mean for the links? Should we remove them? Until they implement them, and then bring them back?
3. What happens when libraries become unmaintained?
4. There is a feeling (for me at least) that the spec is a rather static document(s) that moves slowly and cautiously. For the links in the readme to be valuable, they need to be quite opposite this, and be evergreen/changing. The two might be better if they are separated. 

Now I think the readme should point to locations where developers would be able to find that sort of information, but that could be just a new section with some meta links, or maybe just a sentence like `You can find a list of libraries that implement the various NUTs in this spec at [this link] and [this link].`

Anyway just a thought as the spec matures.

thesimplekid:
Agree, think it makes sense to move them to https://github.com/cashubtc/awesome-cashu or similar and remove them from the spec repo.

PR #94, opened 2024-03-08
README Cleanup: Simplify mandatory table

AngusP:
* For **mandatory** NUTS, just say `all` for wallet and mint support to reduce duplication
* Move the list of Wallets and Mints that support Mandatory NUTS to after the table, to make the full set of options clearer (also sort alpha)
* Markdown table formatting fixes, fix broken lnbits link
* mandatory NUT table can be human-readable


README.md
```diff
@@ -6,25 +6,45 @@ These documents each specify parts of the Cashu protocol. Read the specification
...
+
+| NUT #    | Description                       | Wallets | Mints |
+|----------|-----------------------------------|---------|-------|
+| [00][00] | Cryptography and Models           | all     | all   |
+| [01][01] | Mint public keys                  | all     | all   |
+| [02][02] | Keysets and keyset IDs            | all     | all   |
+| [03][03] | Swapping tokens                   | all     | all   |
+| [04][04] | Minting tokens                    | all     | all   |
+| [05][05] | Melting tokens                    | all     | all   |
+| [06][06] | Mint info                         | [Nutshell][py], [eNuts][enuts] | [Nutshell][py], [cashu-rs-mint][cashu-rs-mint] |
```
AngusP, line 18: Left this row as it was given not all wallets+mints support it yet

README.md
```diff
@@ -6,25 +6,45 @@ These documents each specify parts of the Cashu protocol. Read the specification
...
+| --- | --- | --- | --- |
+| [07][07] | Token state check | [Nutshell][py], [Feni][feni], [Moksha][cashume], [Nutstash][ns], [cashu-ts][ts], [cashu-crab][cashu-crab] | [Nutshell][py], [Feni][feni], [LNbits], [Moksha][moksha], [cashu-rs-mint][cashu-rs-mint] |
+| [08][08] | Overpaid Lightning fees | [Nutshell][py], [Feni][feni], [Moksha][cashume], [Nutstash][ns], [cashu-ts][ts], [cashu-crab][cashu-crab] | [Nutshell][py], [LNbits], [Moksha][moksha], [cashu-rs-mint][cashu-rs-mint] |
 | [09][09] | Deterministic backup and restore | - | -
-| [10][10] | Spending conditions | [Nutshell][py] | [Nutshell][py]
-| [11][11] | Pay-To-Pubkey (P2PK) | [Nutshell][py] | [Nutshell][py]
-| [12][12] | DLEQ proofs | [Nutshell][py] | [Nutshell][py]
+| [10][10] | Spending conditions | [Nutshell][py] | [Nutshell][py] |
+| [11][11] | Pay-To-Pubkey (P2PK) | [Nutshell][py] | [Nutshell][py] |
+| [12][12] | DLEQ proofs | [Nutshell][py] | [Nutshell][py] |
```
AngusP, line 47: Just adds missing trailing `|`s

callebtc (changes requested):
Chef's kiss! 

Could you move the wallet and mint list below all the tables so that both tables are side by side and not interrupted by a long list?

thunderbiscuit:
As is, it's a NACK for me. I think the information on which wallets and mints support which NUTs is important, but does not belong in the spec (see #81 for my full opinion on this). As the project grows and more and more mints and wallet libraries are created, tracking all of them and deciding whether they are up to date enough to be part of the table is a fool's errand IMO (example: Feni has not seen a commit in 6 months and is completely out of date with the spec).

But a simplification of the readme is a great idea. I like the break down into two table, one mandatory and one optional. The cleanest way to move forward IMO is simply with the following, simplified table format:

```md
| NUT #    | Description                       |
|----------|-----------------------------------|
| [00][00] | Cryptography and Models           |
| [01][01] | Mint public keys                  |
| [02][02] | Keysets and keyset IDs            |
| [03][03] | Swapping tokens                   |
| [04][04] | Minting tokens                    | 
| [05][05] | Melting tokens                    |
```

thunderbiscuit:
> That makes sense, though can be one of those things that we stop doing when it does become a faff to keep up-to-date. Early on, it's helpful to have as otherwise discovery can be hard.

Yeah I see your point. Discovery is very important (which is why I like the [awesome list](https://github.com/cashubtc/awesome-cashu), and [main website](https://docs.cashu.space/), and think we should use/promote them more). I do think we're past the early stages tough, and the way I see it, this page is now doing counter-discovery: there are a lot of cashu wallets/libraries/mints that are not in this list (and Feni which should be removed). If someone thinks of this list as some sort of canonical list coming from the devs of Cashu (what does that even mean? this repo is a spec!), they'll miss out on a lot. Again just my opinion, but this is I think creating more confusion and problems than it's worth, and will have to be removed no matter what at some point in the future. My opinion is, do it now.

callebtc:
Thank you this is much better! I agree, will remove Feni from the list, I think the dev is no contributing to GoNuts, which I added here.

callebtc (approved):

PR #95, opened 2024-03-08
Fix - total amount is not 40 but 64 sats

titusz:


AngusP (approved):

callebtc (approved):
ACK [b94bb2f](https://github.com/cashubtc/nuts/pull/95/commits/b94bb2f8869b72d10056e6e9fbf83b42d6709e12)

Thank you for catching this error.

PR #97, opened 2024-03-21
Nut-05: fix small typo

ngutech21:
The settings for Nut-05 in the info-endpoint should be under the key 5 and not 4


callebtc (approved):
Thank you for catching this!

PR #98, opened 2024-03-21
NUT-17: WebSocket state updates

Egge21M:
This is my second take on an optional WebSocket protocol to allow bidirectional messages between clients and mints. WebSockets allow mints to push updates to clients without clients constantly polling them.

The proposal uses ids to keep track of requests, which removed the need to close and open sockets all the time and also allows for multiple request on a single socket connection.


XX.md
```diff
@@ -0,0 +1,164 @@
...
+```
+
+This example shows a Mint signaling support of WebSocket only for `check_quote`
+
+## Specifications
+
+All requests sent from apps to mints over Websocket MUST look like this
+
+```json
+["REQ", "<REQ ID>", "<COMMAND>", "<PARAMS>"]
```
bumi, line 33: would it make sense to use the JSON-RPC format here?

Egge21M: I don't have strong options about this. The format is already very close to the RPC spec so I wouldn't mind. What would be the benefit of adhering to it?

AngusP: > What would be the benefit of adhering to it?

I'd guess mainly existing client libraries for using JSON-RPC rather than having to write your own

Egge21M:
I wonder if this should include a standardised endpoint as well? E.g. wss://<hostname>/v1/ws

callebtc (approved):
ACK [d99a6f1](https://github.com/cashubtc/nuts/pull/98/commits/d99a6f10bd3b1880a7891c9b1ecdf550b74b4e93)

17.md
```diff
@@ -0,0 +1,241 @@
...
+[
+  "20385fc7245...", "d06667cda9b...", "e14d8ca96f..."
+]
+```
+
+Note that `id` and `subId` are unrelated. The `subId` is the ID for each subscription, whereas `id` is part of the JSON-RPC spec and is an integer counter that must be incremented for every request sent over the websocket.
+
+**Important:** If the subscription is accepted by the mint, the mint MUST first respond with the *current* state of the subscribed object and continue sending any further updates to it. 
+
+For example, if the wallet subscribes to a `Proof.Y` of a `Proof` that has not been spent yet, the mint will first respond with a `ProofState` with `state == "UNSPENT"`. If the wallet then spends this `Proof`, the mint would send a `ProofState` with `state == "PENDING"` and then one with `state == "SPENT"`. In total, the mint would send three notifications to the wallet.
```
thesimplekid, line 90: Is it required that the mint sends 3 notifications. In the event the proof is spent via a swap there isn't really a `PENDING` state. As signing the outputs either succeeds and inputs are spent or fails and the inputs should not be `SPENT`. Of course could use the `PENDING` while the verifying and signing happen just not sure how much benefit that adds?

thesimplekid (approved):
ACK [d99a6f1](https://github.com/cashubtc/nuts/pull/98/commits/d99a6f10bd3b1880a7891c9b1ecdf550b74b4e93)

17.md
callebtc: Good points, maybe we can relax this condition.

pablof7z:
just floating this idea, too little too late but wonder if this is something that would have been of interest; what if the mint spoke nostr REQs and acted just like a nostr relay? would be kinda neat to be able to reuse all the nostr comms libraries we've got to talk to mints

Don't know, maybe it's dumb and the protocol is too different, but was curious.

PR #104, opened 2024-03-25
NUT-15: HTLCs

callebtc:
Todo:
- [x] needs comment on dependency of /check otherwise witness can not be retrieved.

### DIFF
--- a/15.md
+++ b/15.md
@@ -7,7 +7,7 @@ NUT-13: Hashed Timelock Contracts (HTLCs)
 
 This NUT describes the use of Hashed Timelock Contracts (HTLCs) which defines a spending condition based on [NUT-10][10]'s well-known `Secret` format. Using HTLCs, ecash tokens can be locked to the hash of a preimage or a timelock. This enables use cases such as atomic swaps of ecash between users, and atomic coupling of an ecash spending condition to a Lightning HTLC.
 
-`HTLC` spending conditions can be thought of as an extension of `P2PK` locks [NUT-11][11] but with a hash lock in `Secret.data` and a new `Proof.witness.preimage` witness in the locked inputs to be spent. The `preimage` that was used to spend a locked token can be retrieved using [NUT-07][07].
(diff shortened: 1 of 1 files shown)


starbackr-dev:
All good with the content. the title shows NUT-13. but the PR shows as NUT-15. Do you want to change it to NUT-14 since NUT-13 is there already?

15.md
```diff
@@ -0,0 +1,71 @@
...
+        "pubkeys",
+        "02698c4e2b5f9534cd0687d87513c759790cf829aa5739184a3e3735471fbda904",
+      ],
+      [
+        "locktime",
+        "1689418329"
+      ],                   
+      [
+        "refund",
+        "033281c37677ea273eb7183b783067f5244933ef78d8c3f15b1a77cb246099c26e"
```
thesimplekid, line 36: These should be the 32 byte keys, as we've discussed for NUT11 https://github.com/cashubtc/nuts/pull/96

nerdyrugbyguy:
References to "hashlock" tag should be removed.  Instead, the data field is what contains the hash of the preimage.

nerdyrugbyguy:
Need to caution users that spending conditions become anyone can spend if not supported by the issuing mint.  When receiving and sending ecash as part of a HTLC swap, it is crucial to verify via NUT-06 that both mints support NUT-15.

15.md
callebtc: Resolved: they should be 33 bytes.

callebtc:
> Need to caution users that spending conditions become anyone can spend if not supported by the issuing mint. When receiving and sending ecash as part of a HTLC swap, it is crucial to verify via NUT-06 that both mints support NUT-15.

Thank you, this is a good comment. This should also be added as a note to NUT-10, as it applies to potentially all spending conditions.

15.md
```diff
@@ -0,0 +1,73 @@
+NUT-13: Hashed Timelock Contracts (HTLCs)
```
thesimplekid, line 1:
```suggestion
NUT-15: Hashed Timelock Contracts (HTLCs)
```

thesimplekid (approved):
LGTM

PR #106, opened 2024-03-26
NUT-XX: Authentication

starbackr-dev:
Currently, access to mint endpoints are open to anyone. Since Cashu tokens are IOUs, the mint should have an optional way of restricting access to some end points if needed. We propose a new feature to provide optional access control for specific endpoints. This will address the need to restrict sensitive operations like minting and melting to authorized parties. This NUT is completely optional and may not be supported by mints and wallets. 

We have developed a (NUT) specification for implementing authentication through bearer tokens. This approach is easy to integrate into wallet provdiders, allowing mints to control access to their endpoints.

Selective Endpoint Authentication

Mints have the flexibility to authenticate certain endpoints selectively. For example, they could restrict minting while allowing unrestricted swapping and melting.

Ofcourse, this will compromise privacy. Both the mint and wallet should be explicit in informing the users about this. 

Example Scenario

Consider a mint such as the one I am currently running (https://stablenut.umint.cash) connected to Strike, a platform enforcing KYC for users transacting with its API. By enabling authentication, the mint can limit minting and melting only to KYC-compliant users. Meanwhile, tokens minted through this restricted channel can continue to be swapped between users freely without restrictions. (Calle can still post them on Nostr/twitter)

Each mint will clearly indicate its access policy on their /info endpoint, empowering wallet providers with the necessary information to guide their users' decisions.

Registration Process is out of scope for this NUT - The registration process for authorized public keys is scoped out of this NUT. Mints can implement their own authentication mechanisms and include authed pubkeys table inside mint database

We invite your review and comments on this proposal.

### DIFF
--- a/16.md
+++ b/16.md
@@ -1,9 +1,9 @@
-## NUT-16: Authentication with Schnorr Signatures
+## NUT-16: Authentication 
 ==========================
 
-`optional` `use with: NUT-04, NUT-05`
+`optional` 
 
-In this document, we describe a mechanism for obtaining time-bounded bearer tokens using Schnorr signatures for accessing protected endpoints for a mint. The registration process for authorizing public keys is outside the scope of this NUT and should be implemented separately by the mint. The auth tokens will be time bounded with an explict expiry time, after which the tokens should be destroyed by the mint.  
+In this document, we describe a mechanism for obtaining time-bounded bearer tokens using user derived public/private keys for accessing protected endpoints for a mint. The registration process for authorizing public keys is outside the scope of this NUT and should be implemented separately by the mint. We derive a unique public/private key pair unique to each mint by hashing the mint URL. This provides additional privacy, as the user's public key cannot be linked across mints.
 
 NOTE: The authentication of a user through the utilization of a public key adopts an accounting model that may compromise privacy.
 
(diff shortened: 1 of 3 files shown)


starbackr-dev:
Thanks for the great feedback and discussion during last month's Dev call! I've made some updates based on the suggestions provided.

1, I've taken out the references to Mint and Melt and instead implemented authentication for any endpoint. This will make it more flexible and allow for easier integration.

2  @trbouma's suggestion to look at LUD-04 LNURL Auth was super helpful, and I was able to incorporate some of the implementation logic to simplify and enhance the privacy. Specifically, I've implemented the linking key concept from LNURL-Auth to create private/pub keys based on the mint URL. This means that each user will have a unique key for each mint, preventing any tied keys to multiple mints. The wallet can still calculate the keys easily from the seed.

Finally, I've introduced an "action" enum to determine the type of action the user is authorized to perform. This will help us keep track of the user's permissions and ensure they only have access to the appropriate actions.

There were some other minor cleanups and code optimization, but these are the main updates. I'm excited about the progress and looking forward to the nextDev call to get more feedback and continue improving the protocol 🚀

16.md
```diff
@@ -0,0 +1,118 @@
...
+POST https://bob.com/v1/auth
+```
+``` json
+
+Post Data:
+{ 
+  action:"mint", 
+  k1:"8278e1a48e61c261916791dabb6af760488e4f01932e11fe7054f59754e3de6e"
+  signature:c568f78e4b234a5f7d8c3b2a679e48d1234567890abcdef
+  linkingKey:7345786068584cd33000582ba87a9ddf77db5377c67910ab59d7e9a5f44
```
thesimplekid, line 61:
```suggestion
  "action":"mint", 
  "k1":"8278e1a48e61c261916791dabb6af760488e4f01932e11fe7054f59754e3de6e"
  "signature": "c568f78e4b234a5f7d8c3b2a679e48d1234567890abcdef"
  "linking_key": "7345786068584cd33000582ba87a9ddf77db5377c67910ab59d7e9a5f44"
```

16.md
```diff
@@ -0,0 +1,118 @@
...
+
+- Secure storage of user private keys is crucial for the security of this authentication mechanism.
+- The mint must implement robust signature verification and token management processes.
+- Client applications must handle bearer tokens securely and respect their expiry times.
+- If a mint choose to implement this NUT, then add this information to /info endpoint as per NUT-06 spec.
+
+```json
+{
+    "nuts": {
+    "15": {
```
thesimplekid, line 105:
```suggestion
    "16": {
```

16.md
```diff
@@ -0,0 +1,118 @@
...
+``` json
+
+Post Data:
+{ 
+  action:"mint", 
+  k1:"8278e1a48e61c261916791dabb6af760488e4f01932e11fe7054f59754e3de6e"
+  signature:c568f78e4b234a5f7d8c3b2a679e48d1234567890abcdef
+  linkingKey:7345786068584cd33000582ba87a9ddf77db5377c67910ab59d7e9a5f44
+}
+
```
thesimplekid, line 63: Missing a ``` here

16.md
```diff
@@ -0,0 +1,118 @@
...
+          "restricted": true,
+          "registration": "/v1/auth/register"   
+        }
+      ],
+      "disabled": false
+    },
+
+}
+
+```
```
thesimplekid, line 118:
```suggestion

```

README.md
```diff
@@ -28,7 +28,7 @@ Wallets and mints `MUST` implement all mandatory specs and `CAN` implement optio
 | [11][11] | Pay-To-Pubkey (P2PK) | [Nutshell][py] | [Nutshell][py] |
 | [12][12] | DLEQ proofs | [Nutshell][py] | [Nutshell][py] |
 | [13][13] | Deterministic secrets | [Nutshell][py], [Moksha][cashume], [cashu-ts][ts] | - |
-
+| [15][16] |Authentication with Schnorr Signatures |
```
thesimplekid, line 31:
```suggestion
| [16][16] | Authentication with Schnorr Signatures |
```

README.md
```diff
@@ -74,3 +74,4 @@ Wallets and mints `MUST` implement all mandatory specs and `CAN` implement optio
 [11]: 11.md
 [12]: 12.md
 [13]: 13.md
+[15]: 15.md
```
thesimplekid, line 77:
```suggestion
[16]: 16.md
```

16.md
```diff
@@ -0,0 +1,118 @@
...
+
+Response:
+
+```json
+HTTP/1.1 200 OK
+
+{
+  "access_token": "9876543210fedcba",
+  "token_type": "Bearer",
+  "expires_in": 3600
```
thesimplekid, line 72:
```suggestion
  "expiry": 1711917473
```

I think this should be the unix time the token expires at more in line with mint/melt quote expiry feilds

16.md
```diff
@@ -0,0 +1,118 @@
...
+# Authentication Flow
+
+**1. Client initiates authentication:** The client application initiates the authentication process by redirecting the user to the designated authorization endpoint. This endpoint is provided by the mint.
+
+Alice wants to access a protected resource on Bob's mint. Her client application initiates the authentication process by redirecting her to Bob's mint authorization endpoint with the required action as `tag` :
+
+```http
+GET https://bob.com/v1/auth/challenge?&tag=<(action enums)>
+```
+Bob's mint generates a challenge `k1` consisting of randomly generated 32 bytes, for example:
```
thesimplekid, line 19:
```suggestion
Bob's mint generates a challenge `k1` consisting of a 64 character hex string generated from 32 random bytes, for example:
```

16.md
```diff
@@ -0,0 +1,118 @@
...
+
+
+`tag` enums meaning:
+- `register`: service is requesting to register a user's `linkingKey`.
+- `mint` user is authed to mint tokens.
+- `melt` user is authed to melt tokens.
+- `...`
+
+
+This k1 is displayed to Alice. Alice's client application then uses her `linkingPrivKey` private key to sign the challenge over the secp256k1 curve using the `linkingPrivKey` private key
```
thesimplekid, line 36:
```suggestion
This k1 is displayed to Alice. Alice's client application then uses her `linkingPrivKey` private key to sign the challenge over the secp256k1 curve using the `linkingPrivKey` private key. We use `libsecp256k1's` serialized 64 byte Schnorr signatures on the SHA256 hash of the message to sign. The message to sign is the field `k1`.
```

16.md
thesimplekid: Can we update this to a valid 64 byte signature.

16.md
```diff
@@ -0,0 +1,118 @@
...
+- `...`
+
+
+This k1 is displayed to Alice. Alice's client application then uses her `linkingPrivKey` private key to sign the challenge over the secp256k1 curve using the `linkingPrivKey` private key 
+
+## `linkingKey` derivation
+
+ - `linkingKey` This is a public key derived from `linkingPrivKey`
+ - `linkingPrivKey` To keep the derived public key unique to each mint, the `linkingPrivKey` should be derived as follows
+ -  Use the derivation path m/138'/0 from the master key to get `hashingkey`
```
davidcaseria, line 42: Is it necessary to specify the exact derivation path used by the client?

davidcaseria:
Is the access token intended to be stateful? Ideally, the mint server would not keep a complete list of active tokens in its database. An alternative could be to specify the claims in a JWT.

Additionally, has something like an HMAC signing been considered instead of an access token? The HMAC secret can be an ECDH exchange with the user’s linking key and the mint’s key, published in the info endpoint. The user’s public key and a timestamp would be required as requested headers for the served to validate the request and prevent replay attacks.

PR #107, opened 2024-03-29
NUT-XX and NUT-XX+1: Mint / Melt Bitcoin On-Chain

ngutech21:
First draft for minting and melting tokens On-Chain.


gandlafbtc:
looks great so far, nice work!

question about the minting

How do we know when the payment is complete? Do we just try to proceed with mint, and see if it works? Or is there a way to look up the `paid` status of a mintQuote?

Also, is there a way to let the client know, how many more confs are required, before the mint can take place?

Last question, what happens if the tx confirmation time exceeds the expiry?

17.md
```diff
@@ -0,0 +1,201 @@
...
+
+## Settings
+The settings for this nut indicate the supported method-unit pairs for minting and whether minting is supported or not. They are part of the info response of the mint ([NUT-06][06]) which in this case reads 
+```json
+{
+  "14": {
+    "supported": true,
+    "methods": [
+      {
+        "payment_method": "btconchain",
```
callebtc, line 179: Should be `method`
```suggestion
        "method": "btconchain",
```

17.md
```diff
@@ -0,0 +1,201 @@
...
+        "unit": "sat",
+        "min_amount": 1000,
```
callebtc, line 181: Might be better to use a realistic minimal amount for on-chain here.
```suggestion
        "min_amount": 10000,
```

18.md
```diff
@@ -0,0 +1,209 @@
...
+   "description": <str>,
+   "amount": <int>,
+   "fee": <int>,
+   "paid": <bool>,
+   "expiry": <int>
+  }
+]
+```
+The mint can return multiple `PostMeltQuoteBtcOnchainResponse` with different `fees` and `expiry` dates. The wallet can choose which one to pay and the other ones will expire.  Where `quote` is the quote ID, `amount` the amount that needs to be provided, and `fee` the additional fee that is required. The mint expects `Alice` to include `Proofs` of *at least* `total_amount = amount + fee`. `paid` indicates whether the request as been paid and `expiry` is the Unix timestamp until which the melt quote is valid.
+
```
callebtc, line 46: I'm wondering why this is necessary. Why shouldn't the sending wallet be free to choose the fee as they wish? The mint wouldn't have to respond with multiple quotes. Multiple quotes also means that the mint would have to generate multiple receive addresses (right?).

What happens when multiple quotes get paid (by accident for example?)

I would suggest that the mint should return only one address and amount, not care about the fee, and only issue ecash after 10 or 12 confirmations.

callebtc, line 46: I think there are two ways of handing the variable fee rate problem here. 

One option could be that the wallet chooses a fee in `PostMeltQuoteBtcOnchainRequest` this would be a bit like a Lightning node choosing the `max_fee` for a payment. The mint could then respond with a response or an error depending on whether it agrees. I think this would have to accompanied with some info setting where the mint announces it's accepted fee range. 

The other option, as you've proposed, would be for the mint to limit the users's choices by returning multiple melt quotes. 

Did you consider the first option as well? I wonder if it has any benefits. Happy to hear thoughts from others on this as well.

callebtc:
> How do we know when the payment is complete? Do we just try to proceed with mint, and see if it works? Or is there a way to look up the `paid` status of a mintQuote?

You can use the following endpoint as per `17.md`:
```
GET https://mint.host:3338/v1/mint/quote/btconchain/{quote_id}
```

> Also, is there a way to let the client know, how many more confs are required, before the mint can take place?

This would be cool but I suspect it won't be generally possible to determine this number with every on-chain backend. Take for example a service like Strike or Blink. Obviously, this would be possible with a bitcoin core, lnd, cln, ... on-chain wallet.

> what happens if the tx confirmation time exceeds the expiry?

That's a very good question. While LN invoices can have a clear expiry, on-chain is a bit more tricky. It might be worth to consider removing this from onchain mints. I think melt expiry doesn't have this problem.

17.md
```diff
@@ -0,0 +1,201 @@
...
+ with the requested `amount` and the `unit`. 
+ 
+ The mint `Bob` then responds with a `PostMintQuoteBtcOnchainResponse`:
+
+```json
+{
+  "quote": <str>,
+  "address": <str>,
+  "paid": <bool>,
+  "expiry": <int>
```
callebtc, line 35: This expiry field might be problematic (cf @gandlafbtc's comment).

callebtc:
Awesome NUT, feels complete! Looking forward to implementing it.

18.md
```diff
@@ -0,0 +1,209 @@
...
+# Settings
+The settings for this nut indicate the supported method-unit pairs for melting. They are part of the info response of the mint ([NUT-06][06]) which in this case reads 
+```json
+{
+  "15": {
+    "supported": true,
+    "methods": [
+      {
+        "payment_method": "btconchain",
```
thesimplekid, line 187:
```suggestion
        "method": "btconchain",
```

18.md
ngutech21: Interesting idea. No haven't thought about it yet.

ngutech21:
> 
> > what happens if the tx confirmation time exceeds the expiry?
> 
> That's a very good question. While LN invoices can have a clear expiry, on-chain is a bit more tricky. It might be worth to consider removing this from onchain mints. I think melt expiry doesn't have this problem.

Good point. I think it's problematic in both way: A wallet could choose a fee that is to low to get in the next block before the quote expires. 
On the other side it might be problematic to have a quote that never expires. The way I see a quote is an offer or a contract between the mint and the wallet. If the quote would never expire, the offer is still valid and it would not be possible for a mint to switch to a different onchain backend. I know this is a very rare case. Maybe it would be a good idea to make the expiry optional and recommend a long expiry like 14 days as default for the mint. Any other suggestions?

thesimplekid:
>  Maybe it would be a good idea to make the expiry optional and recommend a long expiry like 14 days as default for the mint.

Worth noting that for sat denominated quotes a long expiry is fine, but for units with an exchange rate (ie sat/usd) it introduces some risk to the mint where wallets could attempt to arbitrage the exchange rate since its up to them to execute the quote or not.

17.md
```diff
@@ -0,0 +1,201 @@
...
+  "C": "029e8e5050b890a7d6c0968db16bc1d5d5fa040ea1de284f6ec69d61299f671059"
+  }
+]
+```
+
+## Settings
+The settings for this nut indicate the supported method-unit pairs for minting and whether minting is supported or not. They are part of the info response of the mint ([NUT-06][06]) which in this case reads 
+```json
+{
+  "14": {
```
thesimplekid, line 175:
```suggestion
  "17": {
```

18.md
```diff
@@ -0,0 +1,209 @@
...
+{
+  "paid": false
+}
+```
+
+# Settings
+The settings for this nut indicate the supported method-unit pairs for melting. They are part of the info response of the mint ([NUT-06][06]) which in this case reads 
+```json
+{
+  "15": {
```
thesimplekid, line 183:
```suggestion
  "18": {
```

18.md
```diff
@@ -0,0 +1,209 @@
...
+  },
+  {
+    "quote": "OewtRaqe...",
+    "description": "5 sat per vbyte",
+    "amount": 10,
+    "fee": 10,
+    "paid": false,
+    "expiry": 1701704757
+  }
+]
```
thesimplekid, line 79: Do you have other use cases for the general `description` field? If its always fee rate maybe a specific fee rate as an int would be better?

18.md
```diff
@@ -0,0 +1,209 @@
...
+```
+
+Here, `quote` is the melt quote ID to be paid and `inputs` are the proofs with a total amount of at least `amount + fee` (see previous melt quote response).
+
+The mint `Bob` then responds with a `PostMeltBtcOnchainResponse`:
+
+```json
+{
+  "paid": <bool>,
+  "txid": <str>
```
thesimplekid, line 122:
```suggestion
  "txid": <str|null>
```

Can be null if not paid yet

17.md
```diff
@@ -0,0 +1,201 @@
...
+}
+```
+ with the requested `amount` and the `unit`. 
+ 
+ The mint `Bob` then responds with a `PostMintQuoteBtcOnchainResponse`:
+
+```json
+{
+  "quote": <str>,
+  "address": <str>,
```
thesimplekid, line 33: What if we change this from an address to a [BIP-21 URI](https://github.com/bitcoin/bips/blob/master/bip-0021.mediawiki)? This would enable the mint to optionally support [BIP-78 payjoins](https://github.com/bitcoin/bips/blob/master/bip-0078.mediawiki).

callebtc: IUC, BIP-21 is meant for clickable links, additional information such as comments, amount etc, should not be part of the `address` field as it would introduce ambiguity. I think it's fair to assume that the backend would be able to pay most common address formats, and should error (for melt quote) if it doesn't support a particular address format.

elnosh:
should there be a way for mints to charge a fee upfront in a mint quote request?

> How do we know when the payment is complete? Do we just try to proceed with mint, and see if it works? Or is there a way to look up the `paid` status of a mintQuote?
> 
> Also, is there a way to let the client know, how many more confs are required, before the mint can take place?

regarding these questions, would it help or are there any concerns if the mint responded with the number of confirmations it requires to consider the mint quote paid?

18.md
```diff
@@ -0,0 +1,209 @@
...
+GET https://mint.host:3338/v1/melt/btconchain/{tx_id}
+```
+
+The mint `Bob` responds with a `GetMeltBtcOnchainResponse`.
+
+```json
+{
+  "paid": false
+}
+```
```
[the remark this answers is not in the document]
thesimplekid: I think this should be removed. I don't see any benefit over just using `GET /v1/melt/quote/btconchain/{quote_id}`. 

A related open question I have is should the mint broadcast the transaction as soon as it receives valid proofs for the melt, or can the mint decide to wait and include it in a batch of multiple melts?

thesimplekid:
> regarding these questions, would it help or are there any concerns if the mint responded with the number of confirmations it requires to consider the mint quote paid?

The number a confs a mint needs in order to consider a transaction as final should be added to either the quote or the `info` endpoint under the on chain nut settings.

18.md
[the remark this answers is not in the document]
elnosh: > I think this should be removed. I don't see any benefit over just using `GET /v1/melt/quote/btconchain/{quote_id}`.

yeah


> A related open question I have is should the mint broadcast the transaction as soon as it receives valid proofs for the melt, or can the mint decide to wait and include it in a batch of multiple melts?

interesting. That will require some changes to the response of the quote state call

18.md
ngutech21: Other use cases are "next block" or "5 blocks" if the backend uses target blocks instead of sat per vbyte. But I think in most cases it would be sats per vbyte and therefore the description should be an optional field. What do you think?

thesimplekid: Yeah I agree I don't think the description should be required. If a wallet wants to display some estimation like confirmation time it can choose an block explorer to get that from.

18.md
```diff
@@ -0,0 +1,209 @@
...
```
thesimplekid, line 177:
```suggestion

```

EthnTuttle:
No idea if this is relevant but was reviewing it and thought of this thread/PR/NUT.
 > The deposit flow in fedimint places the responsibility on the client to prove that a deposit is spendable by the mint. The client achieves this by generating a tweak applied to the mint's peg-in descriptor and sends funds to the tweaked address. Once the client observes enough confirmations, it will prove to the mint that the coins are spendable with the tweak and the merkle proof that the tx was included in a block. If the proof is valid, the mint will issue ecash for the deposit. With this flow, the mint won't be able to correlate user deposit activity if a new deposit address is used.
https://github.com/fedimint/fedimint/pull/5473#issuecomment-2181019654

17.md
```diff
@@ -0,0 +1,201 @@
...
+Response of `Bob`:
+
+```json
+{
+  "quote": "DSGLX9kevM...",
+  "address": "bc1qkyfgd7mus7ykfd7qkwakq75qsf7rtm...",
+  "paid": false,
+  "expiry": 1701704757
+}
+```
```
thesimplekid, line 60:
```suggestion

```json
{
  "quote": "DSGLX9kevM...",
  "address": "bc1qkyfgd7mus7ykfd7qkwakq75qsf7rtm...",
  "paid": UNPAID,
}
```
```

File: 17.md [hunk 14]
```diff
@@ -0,0 +1,201 @@
...
+ The mint `Bob` then responds with a `PostMintQuoteBtcOnchainResponse`:
+
+```json
+{
+  "quote": <str>,
+  "address": <str>,
+  "paid": <bool>,
+  "expiry": <int>
+}
+```
```
thesimplekid, line 37:
```suggestion
```json
{
  "quote": <str>,
  "address": <str>,
  "state": <str_enum[STATE]>,
}
```
```

File: 17.md [hunk 15]
```diff
@@ -0,0 +1,201 @@
...
+
+Where `quote` is the quote ID and `address` is the payment request to fulfill. `paid` indicates whether the request as been paid and `expiry` is the Unix timestamp until which the mint quote is valid.
+
+Note: `quote` is a **unique and random** id generated by the mint to internally look up the payment state. `quote` **MUST** remain a secret between user and mint and **MUST NOT** be derivable from the payment request. A third party who knows the `quote` ID can front-run and steal the tokens that this operation mints.
```
thesimplekid, line 41:
```suggestion
Where `quote` is the quote ID and `address` is the payment request to fulfill. 

`state` is an enum string field with possible values `"UNPAID"`, `"PAID"`, `"PENDING"`, `"ISSUED"`:
- `"UNPAID"` means that the quote's request has not been paid yet.
- `"PAID"` means that the request has been paid.
- `"PENDING"` means that the quote is currently being issued.
- `"ISSUED"` means that the quote has already been issued.

Note: `quote` is a **unique and random** id generated by the mint to internally look up the payment state. `quote` **MUST** remain a secret between user and mint and **MUST NOT** be derivable from the payment request. A third party who knows the `quote` ID can front-run and steal the tokens that this operation mints.
```

18.md
```diff
@@ -0,0 +1,209 @@
...
+
+```json
+{
+  "paid": <bool>,
+  "txid": <str|null>
+}
+```
+`paid` is a boolean indicating whether the payment was successful, and `txid` is the Bitcoin on chain transaction id of the transmitted transaction. 
+
+If `paid==true`, `Alice`'s wallet can delete the `inputs` from her database (or move them to a history). If `paid==false`, `Alice` can repeat the same request again until the payment is successful.
```
thesimplekid, line 127:
```suggestion
```json
{
  "state": <bool>,
  "txid": <str|null>
}

`txid` is the Bitcoin on chain transaction id of the transmitted transaction. 

`state` is an enum string field with possible values `"UNPAID"`, `"PENDING"`, `"PAID"`:
- `"UNPAID"` means that the request has not been paid yet.
- `"PENDING"` means that the request is currently being paid.
- `"PAID"` means that the request has been paid successfully.

If `state==PAID`, `Alice`'s wallet can delete the `inputs` from her database (or move them to a history). If `paid==PENDING`, `Alice` can repeat the same request again until the payment is successful.
```

18.md
```diff
@@ -0,0 +1,209 @@
...
+  {
+   "quote": <str>,
+   "description": <str>,
+   "amount": <int>,
+   "fee": <int>,
+   "paid": <bool>,
+   "expiry": <int>
+  }
+]
+```
```
thesimplekid, line 44:
```suggestion
```json
[
  {
   "quote": <str>,
   "description": <str>,
   "amount": <int>,
   "fee": <int>,
   "state": <str_enum[STATE]>,
  }
]
```
```

File: 18.md [hunk 18]
```diff
@@ -0,0 +1,209 @@
...
+  "fee": 2,
+  "paid": false,
+  "expiry": 1701704757  
+  },
+  {
+    "quote": "OewtRaqe...",
+    "description": "5 sat per vbyte",
+    "amount": 10,
+    "fee": 10,
+    "paid": false,
```
thesimplekid, line 76:
```suggestion
     "state": <str_enum[STATE]>,
```

18.md
```diff
@@ -0,0 +1,209 @@
...
+Response of `Bob`:
+
+```json
+[
+  {
+  "quote": "TRmjduhIsPxd...",
+  "description": "1 sat per vbyte",
+  "amount": 10,
+  "fee": 2,
+  "paid": false,
```
thesimplekid, line 68:
```suggestion
   "state": <str_enum[STATE]>,
```

thesimplekid (changes requested):

17.md
```diff
@@ -0,0 +1,201 @@
...
+```http
+POST https://mint.host:3338/v1/mint/quote/btconchain
+```
+
+The wallet of `Alice` includes the following `PostMintQuoteBtcOnchainRequest` data in its request:
+
+```json
+{
+  "amount": <int>,
+  "unit": <str_enum["sat"]>
```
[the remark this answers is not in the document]
ngutech21: Why should we remove the unit-field from the PostMintQuoteOnchainRequest? If the mint uses blink as a backend for handling the onchain transactions it can support onchain snd/receive with usd as a currency. Hopefully in the future there will be similar services. https://dev.blink.sv/api/usd-onchain-send

18.md
ngutech21: The description is now optional

ngutech21:
> > > what happens if the tx confirmation time exceeds the expiry?
> > 
> > 
> > That's a very good question. While LN invoices can have a clear expiry, on-chain is a bit more tricky. It might be worth to consider removing this from onchain mints. I think melt expiry doesn't have this problem.
> 
> Good point. I think it's problematic in both way: A wallet could choose a fee that is to low to get in the next block before the quote expires. On the other side it might be problematic to have a quote that never expires. The way I see a quote is an offer or a contract between the mint and the wallet. If the quote would never expire, the offer is still valid and it would not be possible for a mint to switch to a different onchain backend. I know this is a very rare case. Maybe it would be a good idea to make the expiry optional and recommend a long expiry like 14 days as default for the mint. Any other suggestions?

So what should we do? I think it would be best to change the expiry for mint-requests to optional so a mint can use a expiry for usd based mints and no expiry for sat based mint requests? Any other suggestions @gandlafbtc @callebtc  @thesimplekid

17.md
ngutech21: What is the difference between issued and pending? Is issued the initial state?

18.md
ngutech21: Thanks for the suggestion. I replaced the boolean flag paid with the enum

17.md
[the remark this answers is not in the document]
thesimplekid: This is a question of scoping this NUT. To support other units where a conversion rate is involved how expiration is handled as well as how confirmation time is defined is much more important as the conversion rate could change outside an acceptable slippage.  So in order to support other units how confirmation and expiration is defined needs to be included as well.

thesimplekid:
The issue with on chain is there is no way to expire an address unlike ln, so how is a deposit after an expiration handled is it considered a donation to the mint? If that is that case I think that needs to be made explicit. I think this maybe okay since even in the case of removing expiration that is effectively the case if a deposit is made to a mint that is no longer in operation (if the mint operator still has the keys), so having the explicit handling of expiration might be an improvement. 

Of course there is the opportunity for a benevolent mint operator to settle a quote paid after out of band if they are known or have contact info set in the NUT-06. I noticed this is how the [strike api](https://docs.strike.me/walkthrough/receiving-payments#4-receive-the-payment) handles it, though that has its own trade offs. 

As i mentioned above definition of confirmation needs to be included as well with the expiration to signal what needs to happen before the expiration in order for the quote to be settled, num blocks conf, transaction broadcast. This could even be split for example transaction needs to be broadcast before x time to get the exchange rate, but ecash wont be issued until x confirmations.

ngutech21:
> should there be a way for mints to charge a fee upfront in a mint quote request?
> 
> > How do we know when the payment is complete? Do we just try to proceed with mint, and see if it works? Or is there a way to look up the `paid` status of a mintQuote?
> > Also, is there a way to let the client know, how many more confs are required, before the mint can take place?
> 
> regarding these questions, would it help or are there any concerns if the mint responded with the number of confirmations it requires to consider the mint quote paid?

Yes, good point. I have added min_confirmations to the info-endpoint

18.md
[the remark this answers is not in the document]
ngutech21: I removed the redundant GET https://mint.host:3338/v1/melt/btconchain/{tx_id} endpoint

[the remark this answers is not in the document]
ngutech21: > I think this should be removed. I don't see any benefit over just using `GET /v1/melt/quote/btconchain/{quote_id}`.
> 
> A related open question I have is should the mint broadcast the transaction as soon as it receives valid proofs for the melt, or can the mint decide to wait and include it in a batch of multiple melts?

IMO the mint can decide if it wants to batch transaction or not. This is same as centralized btc exchanges handle this. We could add a property to the info-endpoint to signal that a mint might delay the melt for batching, but I don't think this is necessary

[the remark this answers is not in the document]
thesimplekid: > IMO the mint can decide if it wants to batch transaction or not. This is same as centralized btc exchanges handle this. We could add a property to the info-endpoint to signal that a mint might delay the melt for batching, but I don't think this is necessary

I think we do need to signal this and it needs to be more concrete then the mint may delay. I would suggest we add a time component to the `MeltQuoteResponse` this way if the user wants the withdrawal more quickly they will need to pay a higher fee if they are willing to wait and allow the mint to batch they can select a `Quote`  with a lower fee and wait longer. 

Allowing the mint to delay for batching but not defining what a delay means will lead to a bad user experience where they don't know when their transaction will be process and they may need it immediately and be willing to pay for that.

18.md
```diff
@@ -0,0 +1,208 @@
...
+
+Response of `Bob`:
+
+```json
+{
+  "quote": "DSGLX9kevM...",
+  "address": "bc1qkyfgd7mus7ykfd7qkwakq75qsf7rtm...",
+  "state": "UNPAID",
+  "expiry": 1701704757
+}
```
davidcaseria, line 65: Is it worth adding the txid here?

thesimplekid: It might though this question made me think what if a a quote is payed with multiple transaction, I think that should still be valid? So to cover that it would need to be a list of txid? Unless we explicitly say it **MUST** be paid in one tx.

18.md
```diff
@@ -0,0 +1,208 @@
...
+```
+
+The wallet **MUST** store the `amount` in the request and the `quote` id in the response in its database so it can later request the tokens after paying the request. After payment, the wallet continues with the next section.
+
```
thesimplekid, line 69: I think it would be worth adding a note here that wallets **SHOULD** display the quote as a [BIP21](https://github.com/bitcoin/bips/blob/master/bip-0021.mediawiki) URI or BIP21 URI encoded as a QR code to the user. This would help reduce errors of underpaying or overpaying accidentally. 

NOTE: There is a draft to replace BIP21 https://github.com/bitcoin/bips/pull/1555 but as its backwards compatible i suggest we say bip21 for now and can update later.

prusnak:
I wonder if we want to allow minting/melting on Liquid chain and if yes, then whether it should be part of deez nuts or a new pair of nuts. :D

davidcaseria:
> I wonder if we want to allow minting/melting on Liquid chain and if yes, then whether it should be part of these two nuts or a new pair of nuts.

I think it makes more sense to keep separate NUTS, especially since its support should be signaled differently by NUT-06.

prusnak:
> especially since its support should be signaled differently by NUT-06.

Ah yes, that is indeed a good reason.

davidcaseria:
It does not necessarily impact the spec, but something like this should be the recommended way to handle address generation: https://blog.zaprite.com/optimizing-address-usage-for-the-gap-limit/

19.md
```diff
@@ -0,0 +1,200 @@
...
+   "quote": <str>,
+   "description": <str|null>,
+   "amount": <int>,
+   "fee": <int>,
+   "state": <str_enum[STATE]>,
+   "expiry": <int>
+  }
+]
+```
+The mint can return multiple `PostMeltQuoteBtcOnchainResponse` with different `fees` and `expiry` dates. The wallet can choose which one to pay and the other ones will expire.  Where `quote` is the quote ID, `amount` the amount that needs to be provided, and `fee` the additional fee that is required. The mint expects `Alice` to include `Proofs` of *at least* `total_amount = amount + fee`. `paid` indicates whether the request as been paid and `expiry` is the Unix timestamp until which the melt quote is valid.
```
thesimplekid, line 45: It maybe better if we express fee in addition to amount in sats. But also with a time component. This would allow the mint to batch transactions. For example fee of x sats will be paid within the hour, fee of y sats within the day etc. This leaves it up to the mint to actually set the final sats per vbyte to meet whatever time period they agreed to in the quote.

18.md
```diff
@@ -0,0 +1,208 @@
...
+ The mint `Bob` then responds with a `PostMintQuoteBtcOnchainResponse`:
+
+```json
+{
+  "quote": <str>,
+  "address": <str>,
+  "state": <str_enum[STATE]>,
+  "expiry": <int>
+}
+```
```
thesimplekid, line 37: If unit is included here don't we need to add an amount in sats to the quote response so the mint can signal the exchange rate. Otherwise if it is non bitcoin unit the wallet will not know how much bitcoin to the address to fill the quote.

That being said because of the imprecise nature of sending a bitcoin onchain transaction I would be for removing the unit and limiting it to sat denominated quotes.

davidcaseria: I agree. I support keeping it simple to start and only supporting sats as the unit.

18.md
davidcaseria: I think it should be explicit only one transaction is supported.

18.md
```diff
@@ -0,0 +1,208 @@
...
+
+Where `quote` is the quote ID and `address` is the payment request to fulfill. 
+
+`state` is an enum string field with possible values `"UNPAID"`, `"PAID"`, `"PENDING"`, `"ISSUED"`:
+- `"UNPAID"` means that the quote's request has not been paid yet.
+- `"PAID"` means that the request has been paid.
+- `"PENDING"` means that the quote is currently being issued.
```
davidcaseria, line 44: Is this when the transaction is still confirming?

davidcaseria, line 44: Is this when the transaction is still confirming?

18.md
davidcaseria, line 65: Should we return the number of confirmations the transaction currently has?

19.md
davidcaseria: Should the client specify how fast they want the transaction confirmed in the quote request (e.g., with a `target_blocks` field)?

thesimplekid: As its currently written the mint returns a list of quotes with different fees for different speed of confirmation. That could be flipped and the client tells the mint what it wants and the mint returns only one quote for that target. 

I lean towards how its currently written where the mints returns a set of options and then the client picks the one that best suits it. But as i comments i think we need a time component in addition to the fee so that could be expressed in block time.

davidcaseria: Should the spec offer guidance to the mint implementors on what fee rates should be returned so wallet experiences can be similar even though the wallet isn't in control of requesting its fee rate?

davidcaseria: Should the spec offer guidance to the mint implementors on what fee rates should be returned so wallet experiences can be similar even though the wallet isn't in control of requesting its fee rate?

starbackr-dev:
All,  Is there an implementation of this on-chain mint/melt? I'm thinking of starting one with nutshell. Checking to make sure I'm not duplicating work.

thesimplekid:
> All,  Is there an implementation of this on-chain mint/melt? I'm thinking of starting one with nutshell. Checking to make sure I'm not duplicating work. 

PR for cdk is here https://github.com/cashubtc/cdk/pull/172 haven't seen anything for nutshell.

starbackr-dev:
> > All,  Is there an implementation of this on-chain mint/melt? I'm thinking of starting one with nutshell. Checking to make sure I'm not duplicating work.
> 
> PR for cdk is here [cashubtc/cdk#172](https://github.com/cashubtc/cdk/pull/172) haven't seen anything for nutshell.

great. thanks. Will start to work on Nutshell.

18.md
AngusP: Feels simplest to say

1. You can only mint BTC on-chain (sats only, no other units)
2. You can then swap sats ecash for ecash in a different unit with the mint

So if you wanted to get e.g. "USD eCash by making an onchain tx", it's a two-step process. Has the advantage of a swap being atomic, much faster and less flaky than a mint depending on N confirmations

18.md
AngusP: It is always going to be *possible* to make multiple transactions paying the same address, so someone will do that... so how to handle? Multiple transactions also make reorg risk more complex and add consolidation cost for the Mint.

Feels resonable that the requirement should be for a single transaction with one output paying the mint, anything else and the Mint can 'fail' the mint request and refund whatever it got *somehow* -- would need the user to provide a refund address, or the mint would have to issue an ecash claim on it (but then, consolidation cost etc., would be a 'mint with on-chain dust and then drain the mint's lightning channels' attackable behaviour)

19.md
AngusP: > what fee rates should be returned

Surely the mint should be able to pick whatever fee it wants? Needs to be flexible to feerate market swings, but also some mints might want to disincentivise on-chain melts by adding a premium, others might offer it at cost

PR #109, opened 2024-04-05
NUT-00: CBOR-encoded TokenV4

callebtc:
This PR proposes a new TokenV4 binary encoding format using CBOR that reduces token size by around 40%. 

It is implemented as a WIP and  POC in https://github.com/cashubtc/nutshell/pull/502

Todo:
- [x] example and test cases


00.md
```diff
@@ -196,6 +196,52 @@ When serialized, this becomes:
...
+```json
+{
+  "t": [
+    {
+      "m": str, # mint URL
+      "p": [ # proofs
+        {
+          "i": bytes, # keyset ID
+          "a": uint, # amount
+          "s": str, # secret
```
davidcaseria, line 220: Should the secret be of type `bytes`?

callebtc: The secret is a UTF-8 encoded string, so converting it to bytes doesn't save space in this case (in my tests with CBOR).

00.md
```diff
@@ -196,6 +196,52 @@ When serialized, this becomes:
...
+cashu[version][base64_token_cbor]
+```
+
+##### Token format
+
+The deserialized `base64_token_cbor` is a JSON of the same form as a TokenV4 but with shorter keys and data represented as binary data (`bytes`) instead of hex strings (`hex_str`). Note that we have expanded what is called `Proofs` in TokenV3 (called `p` here with TokenV4) showing that its values are also different from the TokenV3 serialization.
+
+```json
+{
+  "t": [
```
AngusP, line 213: Nit: When compressing keys to a single char, if possible it's nice to have no collisions to avoid confusion, and also makes expanding to readable names trivial with a dict (no context needed). Here it's only `"m"` and `"s"` that can have two meanings

* I'd suggest `"m"` for `memo` becomes `"o"` (memO)
* `"s"` in DLEQ could be something else, not sure what it stands for... could be `"g"` from siGnature (I think it's a Schnorr sig along with `e`?), though that may be confused with the `g` generator point, so it could also be `"p"` from resPonse

callebtc: In my view, I think it's fine that labels are duplicated since this object isn't meant for human consumption anyway...

00.md
```diff
@@ -196,6 +196,52 @@ When serialized, this becomes:
...
+  "u": str <optional>, # unit
+  "m": str <optional> # memo
+}
+```
+
+When serialized, a TokenV4 looks like this:
+
+```
+cashuBoWF0gaJhbXVodHRwOi8vbG9jYWxob3N0OjMzMzhhcIKkYWlIAIiC2b1wghxhYQJhc3hAMGY5OTJmYmNmZjQ1MGI2YzJmNWNiYWMwMmQ2OWE4ZDUwOGIwOTEyYzEwZTI1NDVlYmEyZGVmNTU0M2FhNDVlMWFjWCECWAksUObH-oRm-Oa1whlFLElCyz5i_r1zyC1etH_Kmc6kYWlIAIiC2b1wghxhYQhhc3hAYTg5NGZjNDljZWYxMmM3MGFjMDhhZmQ2YTRkZTY4ZmQ2M2JiOWJjYmY5MzAyZTc3OTdiZTQ1ZDc2MGNiOGFmY2FjWCEC4rg3LKRNqFEtqV-xqtqX2WypEzSFhDAd7WHdiSQsZ6c=
+```
```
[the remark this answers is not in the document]
callebtc: I'll work on this!

thesimplekid (commented):
LGTM, maybe just add a test vector for it

AngusP (approved):

00.md
```diff
@@ -196,6 +196,52 @@ When serialized, this becomes:
...
+        },
+        ...
+      ]
+    },
+    ...
+  ],
+  "u": str <optional>, # unit
+  "m": str <optional> # memo
+}
+```
```
thesimplekid, line 236:
```suggestion
```json
{
 "m": str, # mint URL
  "p": [ # proofs
      {
        "i": bytes, # keyset ID
        "a": uint, # amount
        "s": str, # secret
        "c": bytes, # signature
        "d": { # DLEQ proof
          "e": bytes,
          "s": bytes,
          "r": bytes
           }
       }
        ...
      ],
  "u": str <optional>, # unit
  "e": str <optional> # memo
}
```
```

Following some discussion in other channels. I think its best to remove the support for mulimint tokens in V4. It adds unnecessary complexity to the token.

File: 00.md [hunk 5]
```diff
@@ -196,6 +196,52 @@ When serialized, this becomes:
...
+        {
+          "i": bytes, # keyset ID
+          "a": uint, # amount
+          "s": str, # secret
+          "c": bytes, # signature
+          "d": { # DLEQ proof
+            "e": bytes,
+            "s": bytes,
+            "r": bytes
+          }
```
elnosh, line 226: should it say that the DLEQ field is optional?

callebtc: Agreed! Please feel free to PR!

elnosh:
```suggestion
          "d": { # DLEQ proof (optional) 
            "e": bytes,
            "s": bytes,
            "r": bytes
          }
```

callebtc:
Done [ed29eb2](https://github.com/cashubtc/nuts/pull/109/commits/ed29eb240aa06d3d9ea58c083664e7a35241f852)

thesimplekid (approved):
ACK d9b5499a703b66049ce7368d2a8ea26b406ff4ff

PR #112, opened 2024-04-22
Update 11.md

callebtc:
typos

### DIFF
--- a/11.md
+++ b/11.md
@@ -38,10 +38,9 @@ The recipient who owns the private key of the public key `Secret.data` can spend
   "witness": {
     "signatures":
         [
-          "c43d0090be59340a6364dc1340876211f2173d6a21c391115adf097adb6ea0a3ddbe7fd81b4677281decc77be09c0359faa77416025130e487f8b9169eb0c609"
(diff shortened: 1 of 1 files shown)


11.md
```diff
@@ -38,18 +38,17 @@ The recipient who owns the private key of the public key `Secret.data` can spend
   "witness": {
     "signatures":
         [
-          "c43d0090be59340a6364dc1340876211f2173d6a21c391115adf097adb6ea0a3ddbe7fd81b4677281decc77be09c0359faa77416025130e487f8b9169eb0c609"
+"c43d0090be59340a6364dc1340876211f2173d6a21c391115adf097adb6ea0a3ddbe7fd81b4677281decc77be09c0359faa77416025130e487f8b9169eb0c609"
         ]
     }
-  ]
 }
```
AngusP, line 44: Spacing here a bit weird but meh

AngusP (approved):

thesimplekid (approved):
ACK 9a6b102316d570de52b2ee9e970a39a03d7934c4

I think we should also fix a few other things in this nut either here or in a different PR.

- The secret should be serialized with escape chars, both nutshell and CDK do now but its not shown in the example in then nut.  
- The keys should be specified to be the 33 byte pubkeys
- Test vectors should be added.

I've updated #96 with the escaped secret and 33 byte pubkeys in the example and test vectors but maybe worth adding some explanation around it

PR #117, opened 2024-05-04
NUT-6: update contact info

gohumble:
This PR updates the contact information structure. 

The main change is in the representation of contact information, which is now a structured object instead of an array. This allows for a more explicit and readable representation of contact types and values.

### DIFF
--- a/06.md
+++ b/06.md
@@ -25,11 +25,6 @@ With the mint's response being of the form `GetInfoResponse`:
   "description": "The short mint description",
   "description_long": "A description that can be a long piece of text.",
   "contact": [
-    ["email", "contact@me.com"],
-    ["twitter", "@me"],
-    ["nostr" ,"npub..."]
-  ],
-  "contact_info": [
     {
       "type": "email",
       "info": "contact@me.com"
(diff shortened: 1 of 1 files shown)


callebtc:
LGTM, people have been complaining about this for a long time.

callebtc (approved):

thesimplekid (approved):
ACK [6de3a61](https://github.com/cashubtc/nuts/pull/117/commits/6de3a61f93eaf762142c0957a89596252a11218a)

Much prefer the object to the array

Egge21M:
Actually I don't see any benefit to this, but I don't have a strong opinion on it either

callebtc:
> Actually I don't see any benefit to this

We've had many examples in the spec that started off as an array and turned into an object later because arrays suck. I hope we won't add any arrays anymore just to turn them into objects half a year later.

callebtc:
What about a backwards compatible update? We could add the new contact field as a new key, e.g. `contact_info`, such that we don't break existing implementations with it.

callebtc:
ACK [cd9d6ad](https://github.com/cashubtc/nuts/pull/117/commits/cd9d6ad362eb69bfd81326f553764c52acd1ff7b)

callebtc:
I would suggest we implement this in mints and clients before we merge the PR.

thesimplekid:
ACK cd9d6ad362eb69bfd81326f553764c52acd1ff7b

Implemented in CDK in https://github.com/cashubtc/cdk/pull/117

06.md
```diff
@@ -28,16 +28,30 @@ With the mint's response being of the form `GetInfoResponse`:
     ["email", "contact@me.com"],
     ["twitter", "@me"],
     ["nostr" ,"npub..."]
```
thesimplekid, line 30: Following up on the dev call. I would prefer to remove the old contact and make this a breaking change to avoid having to support the old style going forward. From the call it seemed like many wallets don't currently parse this so it wouldn't break current implementations and would be a simple upgrade to support it. 
```suggestion
    ["email", "contact@me.com"],
    ["twitter", "@me"],
    ["nostr" ,"npub..."]
```

06.md
```diff
@@ -28,16 +28,30 @@ With the mint's response being of the form `GetInfoResponse`:
...
-  ],  
-  "motd": "Message to display to users.",  
+  ],
```
thesimplekid, line 31: Following up on the dev call. I would prefer to remove the old `contact` and make this a breaking change to avoid having to support the old style going forward. From the call it seemed like many wallets don't currently parse this so it wouldn't break current implementations and would be a simple upgrade to support it.

callebtc:
> Implemented in CDK in [cashubtc/cdk#117](https://github.com/cashubtc/cdk/pull/117)

Updated, with new field now called "contact" and thus being a breaking change. @thesimplekid we renamed the `type` field to `method` now, which fits a bit better.

PR #124, opened 2024-05-26
NUT-18: Payment Requests

Egge21M:
Cashu transactions are initiated by the sender and can be pushed to the receiver without any additional help. However for many use cases requesting a payment may be more desirable (e.g. point-of-sales). This proposal introduces a standardised format for payment requests, that supply a sending wallet with all information necessary to complete the transaction.


xx.md
```diff
@@ -0,0 +1,103 @@
...
+```
+
+while Transport is defined as
+
+```go
+type Transport struct {
+    T  string `type`
+    Ta string `target`
+}
+```
```
callebtc, line 38: Please use JSON with the type annotations we use in other documents.

xx.md
```diff
@@ -0,0 +1,103 @@
...
+4. Receiver receives the token and finalises the transaction
+
+## Payment Request
+
+A Payment Request is defined as follows
+
+```go
+type PaymentRequest struct {
+  A int         `amount (optional)`
+  U string      `unit`
```
davidcaseria, line 23: Can the unit also be optional and default to sats?

xx.md
```diff
@@ -0,0 +1,103 @@
...
+    T  string `type`
+    Ta string `target`
+}
+```
+
+- amount: The amount of the requested token (payment amount)
+- unit: The unit of the requested token (e.g. sats)
+- mint: The mint to use to mint the token
+- description: A human readable discription that the sending wallet will display after scanning the request
+- transport: The method of transport chosen to transmit the created token to the sender (can be multiple, sorted by preference)
```
davidcaseria, line 44: Is there a use case where transport is empty? Or should there be a minimum of one element in the list?

thesimplekid:
What do you think of an optional condition that could specify spending conditions. For example if the receiver wants it locked using P2PK?

Egge21M:
> What do you think of an optional condition that could specify spending conditions. For example if the receiver wants it locked using P2PK?

Yes, this should be included. I will add it later doday

xx.md
Egge21M: I was thinking that as well, however I can receivers run into compatibility issues quickly if they let users choose the unit.

callebtc:
> What do you think of an optional condition that could specify spending conditions. For example if the receiver wants it locked using P2PK?

Very cool. The receiver could specify the entire `secret` as well.

xx.md
```diff
@@ -0,0 +1,81 @@
...
+## Transport
+
+Transport is an important part of Cashu Payment Requests. Receivers can choose what kind of transport they want to support and where they want to receive the token. By making this clear in the payment requests, wallets can handle payment requests accordingly (or decline them if they do not support the transport layer). A transport consists of a type and a target.
+
+```json
+{
+  "t": str,
+  "ta": Target
+}
+```
```
thesimplekid, line 53: NIT

```suggestion
```json
{
  "t": <str>,
  "ta": <Target>
}
```
```

File: xx.md [hunk 5]
```diff
@@ -0,0 +1,81 @@
...
+```json
+{
+  "a": int <optional>,
+  "u": str,
+  "r": str <optional>,
+  "d": str <optional>,
+  "m": str <optional>,
+  "l": str <optional>,
+  "t": Transport
+}
```
thesimplekid, line 29: Bit of a nit but other nuts are this format

```suggestion
```json
{
  "a": <int|null>,
  "u": <str>,
  "r": <str|null>,
  "d": <str|null>,
  "m": <str|null>,
  "l": <str|null>,
  "t": <Transport>
```

xx.md
```diff
@@ -0,0 +1,81 @@
...
+A Payment Request is defined as follows
+
+```json
+{
+  "a": int <optional>,
+  "u": str,
+  "r": str <optional>,
+  "d": str <optional>,
+  "m": str <optional>,
+  "l": str <optional>,
```
thesimplekid, line 27: What is the format of the lock script?

Egge21M: I was thinking to just make it the well known secret format. Basically what will become Proof.secret.

However thinking about it, that will not work because of the nonce. Do you have something in mind? Otherwise we just have to come up with another Type

xx.md
thesimplekid: Ah yeah looks like both are used in different nuts. So can just leave it for this one and we can make another PR later with just formatting changes if we want to make them all consistent

xx.md
thesimplekid: Maybe just the well known secret without a nonce and the sender can add the nonce?

xx.md
```diff
@@ -0,0 +1,81 @@
...
+  - target: `{num: "phone number"}`
+
+## Encoded Request
+
+Cashu Payment Requests can be encoded and displayed/sent as string or qr code. They underlying object is serialised using CBOR and finally base64 encoded and prefixed with a version byte and standard prefix.
+
+_Example_
+
+```sh
+cashrqAaVhQRVhVWNzYXRhTXgiaHR0cHM6Ly9taW50Lm1pbmliaXRzLmNhc2gvQml0Y29pbmFEeCNQbGVzYXNlIHBheSB0aGUgdmVyeSBmaXJzdCBjYXNodSBwcmFUgaJhVGVub3N0cmJUYXhGbnByb2ZpbGUxcXFzZG11cDZlMno2bWNwZXVlNno2a2wwOGhlNDloY2VuNXhucmMzdG5wdncwbWRndGplbWgwc3V4YTBrag
```
callebtc, line 80: should this be a `cashureq` prefix? 

can we add a version byte like with keysets? It would be good to add a paragraph explaining the payment request serialization like with tokenv3:

```
cashureq[version][request object]
```

callebtc: Maybe it'd be better to use a prefix that doesn't look so visually similar to a token: `creq`?

xx.md
callebtc, line 53:
```suggestion
```json
{
  "t": str,
  "a": Target
}
```
```
can we save a byte here?

File: xx.md [hunk 8]
```diff
@@ -0,0 +1,81 @@
...
+
+```json
+{
+  "t": str,
+  "ta": Target
+}
+```
+
+- t: type of Transport
+  - ta: target of Transport
```
callebtc, line 56:
```suggestion
  - a: target of Transport
```

xx.md
```diff
@@ -0,0 +1,81 @@
...
+The sending wallet MUST honor all parts of a payment request that are specified
+
+## Transport
+
+Transport is an important part of Cashu Payment Requests. Receivers can choose what kind of transport they want to support and where they want to receive the token. By making this clear in the payment requests, wallets can handle payment requests accordingly (or decline them if they do not support the transport layer). A transport consists of a type and a target.
+
+```json
+{
+  "t": str,
+  "ta": Target
```
callebtc, line 51:
```suggestion
  "a": Target
```
can we save a byte here?

xx.md
Egge21M: Yes, good point. The first byte after the prefix is indeed a version byte. I'll update the text to be more clear about that

Egge21M: I have added a slightly different notation that makes it clear what exactly is concatianated and encoded when. Let me know what you think

xx.md
```diff
@@ -0,0 +1,89 @@
...
+
+while Transport is defined as
+
+- a: The amount of the requested token (payment amount)
+- u: The unit of the requested token (e.g. sats)
+- r: The mint to use to mint the token
+- d: A human readable description that the sending wallet will display after scanning the request
+- m: The memo of the requested token (can be used for accounting purposes)
+- l: The lock script of the requested token
+- t: The method of transport chosen to transmit the created token to the sender (can be multiple, sorted by preference)
```
ngutech21, line 40: Having multiple transports and letting the user choose which one they prefer makes sense to me. The json example you provided does not allow providing multiple transports. I think it should be an array of Transport
```json
"t": <Array[Transport]>
```

xx.md
```diff
@@ -0,0 +1,89 @@
...
+{
+  "t": str,
+  "a": Target
+}
+```
+
+- t: type of Transport
+- a: target of Transport
+
+There can be many transport layers, but here are some recommendations:
```
ngutech21, line 58: I think it would be better to have a specific list of transport-types cashu supports in general instead of allowing everything. This list can be upgraded in the future without breaking the api. Otherwise a receiver could just add a new transport-type the wallet does not know how to handle.

xx.md
```diff
@@ -0,0 +1,89 @@
...
+
+- nostr
+  - type: "nostr"
+  - target: "<nprofile...>"
+- post
```
ngutech21, line 63: What is the transport-type post? Do you mean http with the method post? If so I would prefer the term http as the transport-type. Post could be misinterpreted as snailmail

xx.md
Egge21M: That wouldn't be an issue though. If the PR does not include a transport that the wallet can handle, the wallet can not pay it and will let the user know. 

I think we should curate a list of recommended / common transports though.

ngutech21:
Does this nut provide any information through the info-endpoint other than the supported-flag? 
It could make sense to provide the supported currencies and min/max amounts like we do in nut-4 and nut-5. Or is this nut using the same values as nut-5?

xx.md
```diff
@@ -0,0 +1,89 @@
...
+  - type: "email"
+  - target: "<email address>"
+- sms
+  - type: "sms"
+  - target: "<phone number>"
+
+Regardless of the transport layer the payload should always (unless specified differently) be a JSON serialised object as follows:
+
+```json
+{ "token": "cashuAy..." }
```
prusnak, line 76: Why not just the plaintext?

Sending JSON makes sense only for HTTP but for all other transports (e.g. SMS or E-mail) it is awkward.

Egge21M: I agree. I thought it would be good to define a reasonable standard that could work in any transport.

I an definitely open for suggestions

xx.md
```diff
@@ -0,0 +1,89 @@
...
+
+## Payment Request
+
+A Payment Request is defined as follows
+
+```json
+{
+  "a": int <optional>,
+  "u": str,
+  "r": str <optional>,
```
davidcaseria, line 24: Should more than 1 mint be allowed?

xx.md
```diff
@@ -0,0 +1,89 @@
...
+```
+
+- t: type of Transport
+- a: target of Transport
+
+There can be many transport layers, but here are some recommendations:
+
+- nostr
+  - type: "nostr"
+  - target: "<nprofile...>"
```
davidcaseria, line 62: Should the nostr transport follow NIP-61?

xx.md
davidcaseria: I favor removing this requirement and allowing the plaintext encoded token to be used.

xx.md
```diff
@@ -0,0 +1,89 @@
...
+
+```json
+{ "token": "cashuAy..." }
+```
+
+## Encoded Request
+
+Cashu Payment Requests can be encoded and displayed/sent as string or qr code. They underlying object is serialised using CBOR and finally base64 encoded and prefixed with a version byte and standard prefix.
+
+`"creq" + base64(versionByte + cbor(PaymentRequest))`
```
davidcaseria, line 83: I don't have a strong opinion, but should we consider keeping the cashu prefix and using the version byte to differentiate a token and a payment request?

Also, NUT-00 doesn't encode the version byte in the base64; it is included in the prefix. We should keep the spec consistent with this treatment.

xx.md
callebtc: I think it's better to use a structured object because: 
1) We will break things once we decide to add stuff later
2) There is a new payment `id` field that the requester can ask for in the payment payload.

Egge21M:
@callebtc thanks for adding the suggestions. This does indeed LGTM!

xx.md
```diff
@@ -0,0 +1,96 @@
...
+  "t": Transport
+}
+```
+
+while Transport is defined as
+
+- `i`: Payment id to be included in the payment payload
+- `a`: The amount of the requested payment
+- `u`: The unit of the requested payment (MUST be set if `a` is set)
+- `s`: Whether the payment request is for single use
```
davidcaseria, line 37: Should be `r` to match above

xx.md
```diff
@@ -0,0 +1,96 @@
...
+  - type: `nostr`
+  - target: `<nprofile>`
+  - tags: `[["1", "NIP-04"], ["2", "NIP-17"], ["3", "NIP-61"]]`
+- post
+  - type: `post`
+  - target: `<endpoint url>`
+
+## Payment payload
+
+Regardless of the transport layer, the payload sent to the receiver is a JSON serialized object as follows:
```
davidcaseria, line 70: How would this work with NIP-61, where the nostr event is the payment, which has its own format and should be signed by the sender?

Egge21M:
Are we ready to remove NIP-04, because the support for NIP-17 might be too limited still. Afaik nostr-tools does not have it yet.

But I agree that we probably should be looking forward and not carry stuff around that's already considered harmful

callebtc (approved):
LGTM

18.md
```diff
@@ -0,0 +1,111 @@
+# NUT-XX: Payment Requests
```
ok300, line 1:
```suggestion
# NUT-18: Payment Requests
```

18.md
```diff
@@ -0,0 +1,111 @@
...
+
+`optional`
+
+---
+
+This NUT introduces a standardised format for payment requests, that supply a sending wallet with all information necessary to complete the transaction. This enables many use-cases where a transaction is better initiated by the receiver (e.g. point of sale).
+
+## Flow
```
ok300, line 9: IMO it would be helpful to indicate in the Flow at which point user interaction or confirmation is needed, and if there is anything the app must / should show to the user before confirmation, similar to LNURL: https://github.com/lnurl/luds/blob/luds/03.md

18.md
```diff
@@ -0,0 +1,111 @@
...
+3. Sender sends the token according to the transport specified in the payment request
+4. Receiver receives the token and finalises the transaction
+
+## Payment Request
+
+A Payment Request is defined as follows
+
+```json
+{
+  "i": str <optional>,
```
ok300, line 22: It would help to add a short comment for each of these, to indicate 1) what they are, 2) what data type, and 3) if there are any limitations (length, range, etc)

18.md
```diff
@@ -0,0 +1,111 @@
...
+  "r": bool <optional>,
+  "m": Array[str] <optional>,
+  "d": str <optional>,
+  "t": Array[Transport]
+}
+```
+
+while Transport is defined as
+
+- `i`: Payment id to be included in the payment payload
```
ok300, line 34: Are these the fields of `Transport` (as the line just above this says)?

Or is the `Transport` only
```json
{
  "t": str,
  "a": str,
  "g": Array[Array[str, str]] <optional>
}
```

as the section below says?

18.md
```diff
@@ -0,0 +1,111 @@
...
+}
+```
+
+- t: type of Transport
+- a: target of Transport
+- g: optional tags for the Transport
+
+### Transport types
+
+The supported transport types are:
```
ok300, line 60: Which are the transport types? Looks like a list was intended here, but missing.

18.md
callebtc: lol thanks 🙏

prusnak:
typo fix in https://github.com/cashubtc/nuts/pull/178

PR #125, opened 2024-05-26
fix possible typo: remove b in b"Secp256k1_HashToCurve_Cashu_"

melvincarvalho:
if it is a typo?

see: https://t.me/CashuBTC/25742


AngusP (commented):
Not a typo, it's python syntax for a byte string - `b"ascii"` is the same-ish as a byte array of chars, whereas the native string type `"hell😮"` is UTF-8

```python
>>> s = "Hell😮"
>>> s
'Hell😮'
>>> s.encode("utf-8")
b'Hell\xf0\x9f\x98\xae'  # note the `b'` at the start
```

PR #126, opened 2024-05-27
Add fees to NUT-02

callebtc:
Proposing to add fees to the mandatory part of the spec as I expect most public mints will require fees at some point. If the fees are not set, we should assume it as being 0.

Moved around some existing text. The new part is:

### Fees (parts per thousand)

Keysets indicate the fee `input_fee_ppk` that is charged when a `Proof` of that keyset is spent as an input to a transaction. The fee is given in parts per thousand (ppk) per input measured in the `unit` of the keyset and the sum is rounded up to the next larger integer. 

As an example, we construct a transaction spending 3 inputs (`Proofs`) from a keyset with unit `sat` and `input_fee_ppk` of `100`. A fee of `100 ppk` means `0.1 sat` per input. The sum of the fees would be 300 ppk for this transaction and the mint would charge `1 sat` in fees (`ceil(0.3) == 1`). The fees for spending 1-10 inputs is 1 sat, 11-20 inputs is 2 sat and so on.

...

#### Wallet input and output construction
When constructing a transaction with ecash inputs (example: `/v1/swap` or `/v1/melt`), wallets **MUST** add fees to the inputs (or subtract from the outputs) if they spent ecash from a keyset with fees. The mint checks the following equation:

```python
sum(inputs) - sum(fees) == sum(outputs)
```

The `fees` are calculated for each input individually (by summing the fee from the keyset they are from) and then rounded up to the next integer.

### DIFF
--- a/02.md
+++ b/02.md
@@ -58,15 +58,17 @@ Response `GetKeysetsResponse` of `Bob`:
 {
   "keysets": [
     {
-      "id": <keyset_id_hex_str>,
-      "unit": <currency_unit_str>,
+      "id": <hex_str>,
+      "unit": <str>,
       "active": <bool>,
-      "input_fee_ppk": <int>,
+      "input_fee_ppk": <int|null>,
     },
     ...
   ]
 }
 ```
+
+Here, `id` is the keyset ID, `unit` is the unit string (e.g. "sat") of the keyset, `active` indicates whether new ecash can be minted with this keyset, and `input_fee_ppk` is the fee (per thousand units) to spend one input spent from this keyset. If `input_fee_ppk` is not given, we assume it to be `0`.
 
 ## Example response
 
--- a/02.md
+++ b/02.md
@@ -1,5 +1,4 @@
-NUT-02: Keysets and keyset ID
-==========================
+# NUT-02: Keysets and keyset ID
 
 `mandatory`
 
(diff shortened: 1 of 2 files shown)


02.md
```diff
@@ -1,29 +1,53 @@
...
+### Fees
+
+Keysets indicate the fee `input_fee_ppk` that is charged when a `Proof` of that keyset is spent as an input to a transaction. The fee is given in parts per thousand (ppk) per input measured in the `unit` of the keyset. The total fee for a transaction is the sum of all fees per input rounded up to the next larger integer (that that can be represented with the keyest).
+
+As an example, we construct a transaction spending 3 inputs (`Proofs`) from a keyset with unit `sat` and `input_fee_ppk` of `100`. A fee of `100 ppk` means `0.1 sat` per input. The sum of the individual fees are 300 ppk for this transaction. Rounded up to the next smallest denomination, the mint charges `1 sat` in total fees, i.e. `fees = ceil(0.3) == 1`. In this case, the fees for spending 1-10 inputs is 1 sat, 11-20 inputs is 2 sat and so on.
+
+#### Wallet transaction construction
+
+When constructing a transaction with ecash `inputs` (example: `/v1/swap` or `/v1/melt`), wallets **MUST** add fees to the inputs or, vice versa, subtract from the outputs. The mint checks the following equation:
+
```
elnosh, line 32: any reason for including melts? I thought mints could use the `fee_reserve` field in melts to charge fees

elnosh: Ok, I think I see part of it. The mint does not know the number of inputs so it can't calculate the fee per input. But it seems to me that the mint can still charge fees in melts although not in the same way as with `input_fee_ppk`. I see most mint implementations have some setting like `LIGHTNING_FEE_PERCENT` which is the percent they are currently using for the `fee_reserve` (or if not, let me know), so mints could add to that their desired fees for the melt. So that field will be `LIGHTNING_FEE_PERCENT` + `MELT_FEE_PERCENT`. It's by percent and not by input but still a way to charge fee.  

> the whole fee system could be circumvented by users simply using internal invoices instead

Maybe I'm missing something but I think mints today have ways to charge fees in both mints and melts requests to avoid this. If a mint receives a request for a mint quote for 1000 sats and it wants to charge a fee, it can return an invoice for 1002 sats (2 sats in fees). Similar in melt, if a wallet user wants a 1000 sat invoice paid then the mint can ask 1002 sat worth of proofs (a fee accounting for both lightning and melt).

callebtc: Using only the `fee_reserve` would also mean that everyone has to reserve the maximum possible fee, so most users who won't use 1000x1 sat inputs would have to reserve as much as those who do.

Regarding the check in the mint, what I do in nutshell is for a `/mint` is:

```python
# verify that the amount of the input proofs is equal to the amount of the quote
total_provided = sum_proofs(proofs)
total_needed = (
    melt_quote.amount
    + melt_quote.fee_reserve
    + self.get_fees_for_proofs(proofs)
)
if not total_provided >= total_needed:
    raise TransactionError(
        f"not enough inputs provided for melt. Provided: {total_provided}, needed: {total_needed}"
    )
```

elnosh: ok, prob a misunderstanding as I was proposing that this change does not affect melt operations and only do the input fee calculation for swaps. 

> This way it seems cleaner: the mint checks is the provided inputs are amount + fee + fee_reserve and can deduct the overpaid fee_reserve like they do now

I was thinking about this and agree that this could be cleaner to know the type of fee. But you could also end up with weird flows like paying fees in one way and the possibility of asking for back fees in another way

02.md
```diff
@@ -1,29 +1,53 @@
...
+```
+
+Here, `sum(inputs)` and `sum(outputs)` mean the sum of the amounts of the inputs and outputs respectively. `fees` is calculated from the sum of each input's fee and rounded up to the next larger integer:
+
+```python
+def fees(inputs: List[Proof]) -> int:
+  sum_fees = 0
+  for proof in inputs:
+    sum_fees += keysets[proof.id].input_fee_ppk
+  return ceil(sum_fees / 1000)
```
conduition, line 44: This implies wallets must use floating point operations to compute fees, which is a non-deterministic procedure. Wallets would need to be robust against off-by-one errors. I ran into this problem while working on fees for #128. 

If it were me, i'd define the fees with integer arithmetic only so that every implementation can agree. 

```suggestion
def fees(inputs: List[Proof]) -> int:
  sum_fees = 0
  for proof in inputs:
    sum_fees += proof.amount * keysets[proof.id].input_fee_ppk // 1000
  return sum_fees
```

callebtc: I had this concern too and thought that using an intermediary float for the last step would not include any ambiguity but thinking about it now, it might.

Using integer division `//` however will round down and not up, so I'm not sure whether your suggestion would produce the same result as using `ceil()`. Do you agree?

conduition: Oh sorry, i missed that. if you want to round up with integer division, try this:



```suggestion
def fees(inputs: List[Proof]) -> int:
  sum_fees = 0
  for proof in inputs:
    sum_fees += proof.amount * keysets[proof.id].input_fee_ppk
  return (sum_fees + 999) // 1000
```

02.md
```diff
@@ -142,12 +170,23 @@ Response of `Bob` (same as [NUT-01][01]):
...
-    }
+    },
   }, ...
   ]
 }
 ```
 
+## Wallet implementation notes
+
+Wallets can request the list of keyset IDs from the mint upon startup and load only tokens from its database that have a keyset ID supported by the mint it interacts with. This also helps wallets to determine whether the mint as added a new current keyset or whether it has changed the `active` flag of an existing one.
```
gandlafbtc, line 181:
```suggestion
Wallets can request the list of keyset IDs from the mint upon startup and load only tokens from its database that have a keyset ID supported by the mint it interacts with. This also helps wallets to determine whether the mint has added a new current keyset or whether it has changed the `active` flag of an existing one.
```

02.md
callebtc: Noticed another issue here: the fees do not depend on `proof.amount` but only on the number of input proofs.

callebtc: Addressed in [b7d64fd](https://github.com/cashubtc/nuts/pull/126/commits/b7d64fdcae5b9e4225181154c8bd07a979314a2f).

02.md
```diff
@@ -1,29 +1,87 @@
...
+sum(inputs) - fees == sum(outputs)
+```
+
+Here, `sum(inputs)` and `sum(outputs)` mean the sum of the amounts of the inputs and outputs respectively. `fees` is calculated from the sum of each input's fee and rounded up to the next larger integer:
+
+```python
+def fees(inputs: List[Proof]) -> int:
+  sum_fees = 0
+  for proof in inputs:
+    sum_fees += keysets[proof.id].input_fee_ppk
```
conduition, line 49: Still missing `proof.amount` here:

```suggestion
    sum_fees += proof.amount * keysets[proof.id].input_fee_ppk
```

02.md
elnosh: > Could you make an example?

I was hung up with having 2 different fee fields (input fees and lightning fees) so it was that scenario but seems that should be the way forward in order to be able to charge by input.

02.md
callebtc: > Still missing `proof.amount` here:

The fees are not amount-dependent.

thesimplekid (approved):

Egge21M (approved):

elnosh (approved):

lescuer97:
merged in nutmix.
https://github.com/lescuer97/nutmix/pull/74

PR #127, opened 2024-05-27
Add spending condition trees to multiplex NUT-10

conduition:
This proposes a NUT similar to TapRoot's BIP341, which allows the mint to enforce pre-committed sets of spending conditions while only requiring a single condition to be revealed.

PR #128, opened 2024-05-27
NUT for DLC execution

### LINKED ISSUE #122: NUT for Discreet Log Contract support
Today I published a first-draft of an idea for [discreet log contract](https://bitcoinops.org/en/topics/discreet-log-contracts/) integration for Chaumian Ecash mints.

https://conduition.io/cryptography/ecash-dlc/

Before I go to the trouble of writing up a NUT submission, is this even the right place for such an idea? Or would DLCs be outside the scope of Cashu? 

Naturally I don't think DLC support should be mandatory for any cashu implementation. But DLCs could be a highly desirable feature for ECash... something which might incentivize non ecash users to take the plunge. DLCs could offer a sort of programmable functionality on top of Ecash, outside of simple unconditional payments, without adding much complexity to the mint itself.

conduition:
Depends on https://github.com/cashubtc/nuts/pull/127

Closes #122 

Adds a NUT which enables DLC execution using a cashu mint as a blind intermediary. Based on [this proposal](https://conduition.io/cryptography/ecash-dlc/).


dlc.md
```diff
@@ -0,0 +1,520 @@
...
+
+The locking points `[K1, K2, ... Kn]` are elliptic curve points, encoded in SEC1 compressed binary format.
+
+The locking points are blinded by the participants, to obscure the nature of the DLC from the mint at settlement time. Blinded locking points are computed as:
+
+```python
+Ki_ = Ki + bi * G
+```
+
+...for some blinding secret `bi` known to all DLC participants. Each locking point SHOULD be allocated a _unique_ blinding secret.
```
conduition, line 36: I think it should be safe to use a single blinding secret `b` to mask every locking point, like this

```
Ki_ = b * Ki
```

This would be easier on the client, as the wallet only needs to store one blinding secret per DLC instead of `n` secrets. 

Can somebody check me on this?

dlc.md
```diff
@@ -0,0 +1,520 @@
...
+}
+```
+
+- `Payout.dlc_root` is a DLC merkle root hash.
+- `Payout.signature` is a [BIP-340] signature made on the `dlc_root` hash, which should verify against `Payout.pubkey`.
+- `Payout.outputs` is a set of blinded messages for the mint to sign.
+
+For each `Payout` object, the mint performs the following checks:
+
+1. Validate that `Payout.signature` is a valid [BIP-340] signature made by `Payout.pubkey` on `payout.dlc_root`
```
conduition, line 417: This could also be a hash preimage, or the dlog of `Payout.pubkey`.

dlc.md
```diff
@@ -0,0 +1,612 @@
...
+
+```json
+{
+  "atomic": <bool|null>,
+  "settlements": [
+    {
+      "dlc_root": "2db63c93043ab646836b38292ed4fcf209ba68307427a4b2a8621e8b1daeb8ed",
+      "outcome": {
+        "timeout": 1716777419,
+        "P": "{\"361cd8bd1329fea797a6add1cf1990ffcf2270ceb9fc81eeee0e8e9c1bd0cdf5\":\"10000\"}"
```
lollerfirst, line 416: Also here

dlc.md
```diff
@@ -0,0 +1,612 @@
...
+
+To register and fund a DLC on the mint, the funder issues a `POST /v1/dlc/fund` request to the mint, sending a request body of the following format.
+
+```json
+{
+  "atomic": <bool|null>,
+  "registrations": [
+    {
+      "dlc_root": "2db63c93043ab646836b38292ed4fcf209ba68307427a4b2a8621e8b1daeb8ed",
+      "funding_amount": <int>,
```
lollerfirst, line 227: So all the provided inputs must be signed by keys of the same denomination? Or is just funding_amount in sats?

dlc.md
```diff
@@ -0,0 +1,612 @@
...
+To mark the DLC as settled on the mint, the wallet issues a `POST /v1/dlc/settle` request with the following body format.
+
+```json
+{
+  "settlements": [
+    {
+      "dlc_root": "2db63c93043ab646836b38292ed4fcf209ba68307427a4b2a8621e8b1daeb8ed",
+      "outcome": {
+        "k": "8e935aec5668312be8f960a5ecc3c5dd290e39985970bfd093047df7f05cc9ec",
+        "P": "{\"361cd8bd1329fea797a6add1cf1990ffcf2270ceb9fc81eeee0e8e9c1bd0cdf5\":\"10000\"}"
```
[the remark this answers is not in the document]
conduition: It can be any positive integer. Since payout values are computed by relative weights, this number can be anything if there is only one recipient in the payout weights map.

dlc.md
conduition: input proofs can be in any combination of denominations, as per the usual rules of swap/melt operations.  

The new thing here is the `funding_amount` which is an explicit indicator to the mint, saying "i want to fund this DLC with exactly this much money". The value of all the input proofs will need to meet that threshold - or more if the mint charges fees. (see L254 for how fees are computed)

Note also that the `threshold` tag in the `DLC` secret is compared against the `funding_amount`

lollerfirst: Yes I understand this, but how do we know what denomination the funding amount is in? I, for now, just assumed it is always sats bit maybe it should be specified in each registration as an additional field

conduition: Oh sorry, I see what you mean. I forgot about multi-currency support. Yes, good idea, [I added a `unit` field to disambiguate that for the mint](https://github.com/cashubtc/nuts/pull/128/commits/e10d9f2b3e6cc13d2f931fb6affdbbaf8ed4ea0b). This is also something the mint needs to store (probably easier to just have different tables for DLCs funded under different `unit`s than to concretely store the unit as a string)

conduition:
Note to self todo:
- Remove the atomic flag. It adds an unnecessary database-rollback requirement to the implementation and realistically very few users will need to register or resolve multiple DLCs atomically.
- Instead of a single dedicated key for funding proof signatures, use the first key from the relevant unit's active keyset

dlc.md
```diff
@@ -0,0 +1,607 @@
...
+  ]
+}
+```
+
+`Settlement.outcome.timeout` is the timeout timestamp `t`. The steps for the mint to verify the settlement is the same as an outcome settlement (above), with one modification: `Settlement.merkle_proof` is verified differently.
+
+```python
+T = hash_to_curve(Settlement.outcome.timeout.to_bytes(4, 'big'))
+leaf_hash = SHA256(T || Settlement.outcome.P)
+assert merkle_verify(dlc_root, Settlement.merkle_proof, leaf_hash)
```
lollerfirst, line 428: Can't we just do:
```python
T = hash_to_curve(Settlement.outcome.timeout.to_bytes(4, 'big'))
P = hash_to_curve(Settlement.outcome.P.encode("utf-8"))
leaf_hash = sha256((T+P).serialize())
assert merkle_verify(dlc_root, Settlement.merkle_proof, leaf_hash)
``` ?

Also just noticed the unix epoch is cut to 4 bytes: this will bug in ~2035 or something like that when the epoch is a number bigger than 2^32

File: dlc.md [hunk 7]
```diff
@@ -0,0 +1,607 @@
...
+- `Settlement.merkle_proof` is a list of hashes, which must prove that `SHA256(ki_ * G || Pi)` is a leaf hash of the `dlc_root` merkle hash.
+
+The mint verifies each settlement object as follows:
+
+1. If `Settlement.outcome.P` fails to parse as a JSON dictionary mapping pubkeys to positive integers, or if the number of entries in that dictionary exceeds [the `max_payouts` setting](#mint-settings), return an error.
+1. If `Settlement.dlc_root` does not correspond to any known funded DLC, return an error.
+1. If `Settlement.dlc_root` corresponds to a DLC, but that DLC has already been settled, return an error.
+1. Verify `Settlement.merkle_proof`:
+
+```python
```
lollerfirst, line 392: See my other comment.

dlc.md
```diff
@@ -0,0 +1,607 @@
...
+leaf_hash = SHA256(Settlement.outcome.k * G || Settlement.outcome.P)
+assert merkle_verify(dlc_root, Settlement.merkle_proof, leaf_hash)
+```
+
```
lollerfirst, line 396: see my other comment on the timeout redemption

dlc.md
conduition: > Also just noticed the unix epoch is cut to 4 bytes: this will bug in ~2035 when the epoch is a number bigger than 2^32

Correction: it will be a bug in _2106,_ since 0xFFFFFFFF is an unsigned integer. Although it probably won't matter, i'd be open to using a 64-bit encoding. Can't hurt

conduition: > Can't we just do:


Possibly, yes, but I think my approach is faster (one fewer `hash_to_curve` invocations, and one less EC point addition op). 

It also lends itself better to formal proofs of security: The client's goal with this procedure is to prove to the mint that `Settlement.outcome.timeout` and `Settlement.outcome.P` were committed to at registration time (and also that they are one of the payees in `P`). 

If you define `leaf_hash = sha256(hash_to_curve(t)+hash_to_curve(P))`, then a formal proof of soundness for the server would need an extra assumption - namely that for any point $E$, it is infeasible to find two inputs $x$ and $y$ such that $H'(x) + H'(y) = E$ (where $H'$ is the `hash_to_curve` function). I'm not sure whether that statement can be proven. [Wagner's Algorithm](https://conduition.io/cryptography/wagner/#Wagner%E2%80%99s-Algorithm) could _possibly_ be applied to reduce the amount of work needed to forge a fraudulent proof. I don't think it would be possible/practical but still... it'd be safer and simpler to just do `sha256(T || P)` instead. That is pretty easy to prove secure.

conduition: I didn't mention this explicitly in the spec, but the reason I chose to do `hash_to_curve(timeout)` is because it allows implementations to type the leaves of the merkle tree as tuples of `(Point, String)`. A simpler approach, such as:
```
leaf_hash = sha256(Settlement.outcome.timeout.to_bytes() || Settlement.outcome.P)
```

...would require leaf hashes be expressed as a union of either `(Time, string)` or `(Point, string)`.

lollerfirst: > for any point $E$ it is infeasible to find two inputs $x$ and $y$ such that $H'(x) + H'(y) = E$
I think this is the assumption upon which the whole Chaumian e-cash scheme relies on, without the addition. Namely if you have a point $E$ of which you know the discrete logarithm ($eG = E$), and you can find an input $x$ such that $H'(x) = E$ then you break the scheme because you can forge any proof you want.

> Wagner algorithm
I am familiar with Wagner's algorithm.

Having said this, it's a non-issue. We can keep it like this.

conduition: Updated to use a 64-bit timestamp for futureproofness

conduition:
Added the following spec changes:
- No more 'atomic' flags https://github.com/cashubtc/nuts/pull/128/commits/ce22a20b4de9374d14f2dfb387c4564885727be6
- Timeout timestamps are now encoded as uint64 instead of uint32 https://github.com/cashubtc/nuts/pull/128/commits/b67e62a5686e468513f1f76e4181983efba5ade1 
- Add recommendations for clients to avoid being scammed by replay attacks https://github.com/cashubtc/nuts/pull/128/commits/6990d591e7cc1b844a6a52c392d34c609d79cebf
- Improve security of the `funding_proof` signature by the mint. It now commits to the `funding_amount` explicitly, and to the funding `unit` implicitly, by using a signing key from an active keyset instead of a dedicated key.

dlc.md
```diff
@@ -0,0 +1,675 @@
...
+For each `Payout` object, the mint performs the following checks:
+
+1. Authenticate `Payout.witness`:
+  - If present, validate `Payout.witness.secret` is the discrete log (private key) of `Payout.pubkey` (either parity).
+  - If present, validate `Payout.witness.signature` as a [BIP-340] signature made by `Payout.pubkey` on `payout.dlc_root`
+  - If either of the above fields are invalid, or if neither are present, return an error.
+1. If `Payout.dlc_root` does not correspond to any known funded DLC, return an error.
+1. If `Payout.dlc_root` corresponds to a known DLC, but that DLC has not been settled, return an error.
+1. If `Payout.pubkey` is not a key in the `debts` map, return an error.
+1. If `sum([out.amount for out in Payout.outputs]) != debts[Payout.pubkey]`, return an error.
```
lollerfirst, line 539: I think here it should be `sum([out.amount for out in Payout.outputs]) != debts[Payout.pubkey]` where `eligible_amount` is calculated as follows:
```python
denom = sum(debts.values())
nom = debts[payout.pubkey]
eligible_amount = int(nom / denom * funding_amount)
``` 
Also, I was thinking we can discriminate based on whether the requested amount is equal or less than the eligible amount:
* if it's equal, we drop the `pubkey` entry from the debts map
* if it's less, we adjust the debts map so that the `pubkey` entry displays a weight value calculated such that `new_eligible_amount == old_eligible_amount - requested_amount`

conduition: > I think here it should be `sum([out.amount for out in Payout.outputs]) != eligible_amount` where eligible_amount is calculated as follows:
> 
> ```
> denom = sum(debts.values())
> nom = debts[payout.pubkey]
> eligible_amount = int(nom / denom * funding_amount)
> ```


Earlier in the document, we define `debts` as follows:

```python
weights = json.loads(Settlement.outcome.P)
weight_sum = sum(weights.values())
debts = dict(((pubkey, funding_amount * weight // weight_sum) for pubkey, weight in weights.items()))
```

You can view `debts` as a mapping from `weights[pubkey]` to `weights[pubkey] / sum(weights.values()) * funding_amount`. Thus we can assert:

```python
sum(debts.values()) ~= funding_amount
``` 

(approximately, due to integer division rounding)



In your suggestion you have:

```python
eligible_amount = int(debts[payout.pubkey] / sum(debts.values()) * funding_amount)
```

But as i just showed, `sum(debts.values())` is approximately equivalent to `funding_amount`. Simplify your definition of `eligible_amount` and you'll find it's the same as `debts[payout.pubkey]`:

```
debts[payout.pubkey] / sum(debts.values()) * funding_amount
debts[payout.pubkey] / funding_amount * funding_amount
debts[payout.pubkey]
```

conduition: > Also, I was thinking we can discriminate based on whether the requested amount is equal or less than the eligible amount:

Interesting. Is there any reason (other than a computational error) why a wallet wouldn't want to claim its full payout at once?

lollerfirst: Aw, man. I think I might have misinterpreted/read too quickly that part of the spec. I thought the debts map was just the `Settlement.outcome.P` once it was validated (BTW that's also what I had implemented).

Any reason why it can't be?

lollerfirst: > > Also, I was thinking we can discriminate based on whether the requested amount is equal or less than the eligible amount:
> 
> Interesting. Is there any reason (other than a computational error) why a wallet wouldn't want to claim its full payout at once?

It's just something that can happen. Do we let the mint just rug the claimant in that case? I thought it might be better not to.

conduition: > Any reason why it can't be?

Technically you could, but you'd also have to save the `weight_sum`. It's simpler to calculate the debts at settlement time rather than at claim time while you still have access to the full `weights` map.

> It's just something that can happen. Do we let the mint just rug the claimant in that case? I thought it might be better not to.


If the claimant's client is buggy for some reason and won't compute the correct debt amount on its own, then a claimant could just use `GET /v1/dlc/status/{dlc_root}` to see the correct `debts[pubkey]` value and withdraw directly using a different client. But integer arithmetic is deterministic regardless of programming language, so this shouldn't happen as long as clients follow the spec. 

As for specifically why i'm trying to avoid this, it's because it would allow one more vector for clients to DoS the mint (by consuming space with pubkeys for cheap). This is something I was hoping to clean up once we had a working implementation we can test against.

lollerfirst: > This is something I was hoping to clean up once we had a working implementation we can test against.

Right now it should do everything. It surely has a million problems but you can test it if you want.

lollerfirst: > As for specifically why i'm trying to avoid partial claims, it's because it would allow one more vector for clients to DoS the mint (by consuming space with pubkeys for cheap). This is something I was hoping to clean up once we had a working implementation we can test against.

Right now if the DLC is overpaid, the mint automatically adjusts the funding amount. So one clients might provide an amount which doesn't exactly match their eligible amount, thereby failing the transaction.

lollerfirst (approved):

dlc.md
conduition: > So one client might unknowingly provide an amount which doesn't exactly match their eligible amount, thereby failing the transaction.

If a DLC is overfunded, all clients involved in the DLC must know about it. The `funding_amount` is committed to in the `DlcFundingProof` that the funder uses to prove DLC registration. If for some reason a funder goes silent after registering and overfunding a DLC, their peer clients can simply use `GET /v1/dlc/status/{dlc_root}` to see the true funding amount. This endpoint will also explicitly return the `debts` map once the DLC is settled, allowing faulty clients to recover gracefully if the `POST /v1/dlc/payout` request fails due to a mismatching output sum. 

I just don't see the point in allowing partial payouts when all these options are available. It feels like an invitation to a class of "missing money" bugs in client implementations where clients withdraw some money and leave the rest on the table unknowingly. It also allows one more way to DoS the mint which we'll have to clean up later.

conduition: it's your call if you'd like to implement partial withdrawals (it's your code after all), but I would recommend against it.

lollerfirst: Ok ok. I'll change it back.

dlc.md
```diff
@@ -0,0 +1,675 @@
...
+
+```python
+Ki_ = Ki + b * G
+```
+
+...for some blinding secret `b` known to all DLC participants. The blinding secret should be randomly selected by any participant. It should NOT be derived deterministically from the oracle announcement.
+
+## Payout Structures
+
+Payout structures are serialized dictionaries which map `xonly_pubkey -> weight`.
```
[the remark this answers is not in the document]
conduition: The use of 33-byte pubkeys for BIP340 signatures was a mistake. See https://github.com/cashubtc/nuts/issues/133. New NUTs should use xonly pubkeys when validating BIP340 signatures

thesimplekid:
Closing as there is no active work. Please reopen if work continues.

PR #129, opened 2024-05-30
Fix incorrect test vector

Kukks:
Based on https://github.com/cashubtc/cdk/blob/main/crates/cdk/src/dhke.rs#L219, I was finally able to figure out why my seemingly very simple crypto was not working.


thesimplekid (approved):

callebtc:
🫡

PR #141, opened 2024-06-25
NUT-04: add new `state` enum, deprecate `paid`

callebtc:
## What
 
This change to NUT-04 deprecates the `paid` field to `PostMintQuoteBolt11Response` and replaces it with a new `state` field that is a string enum with four possible values: `UNPAID`, `PAID`, and `ISSUED`. 

### Mints

- add new `state` field to `PostMintQuoteBolt11Response`
- keep the `paid` field around until all wallets update

### Wallets

- replace the `paid` field with `state`

As long as `paid` is kept around, wallets can still function the same way as before, also when the `POST /v1/mint/bolt11` is changed (since JSON is "backwards compatible" to new fields).

Tracking progress:
- [x] nutshell
- [x] CDK
- [x] cashu-ts
- [x] nutmix
- [x] gonuts
- [ ] ...


elnosh:
when would a quote be pending?

callebtc:
> when would a quote be in a `PENDING` state?

When it's currently being issued (i.e. in a mint transaction). The pending state can be used by the mint to prevent race conditions.

lescuer97:
Implemented in [Nutmix](https://github.com/lescuer97/nutmix/pull/50)

callebtc:
Pending state removed in [9e86b9e](https://github.com/cashubtc/nuts/pull/141/commits/9e86b9e5de821c4b600350741cace6d83ae17bb6)

thesimplekid (approved):
ACK [9e86b9e](https://github.com/cashubtc/nuts/pull/141/commits/9e86b9e5de821c4b600350741cace6d83ae17bb6)

elnosh (approved):

callebtc:
LFG

Issue #110, opened 2024-04-16
NUT-04: Add optional memo for `PostMintQuoteBolt11Request`


callebtc:
There have been multiple [requests](https://github.com/cashubtc/nutshell/issues/322) to be able to control the `memo` of an invoice. The availability of this option could depend on the backend used.

NUT-04 should have an optional `memo` field that a mint can can choose to honor (if their Lightning backend supports setting it).

gudnuf:
Along with the option for `memo` and `expiry` it would be good to have `description_hash`. [LUD 06](https://github.com/lnurl/luds/blob/luds/06.md) requires the invoice to be generated with a description hash.

callebtc:
> Along with the option for `memo` and `expiry` it would be good to have `description_hash`. [LUD 06](https://github.com/lnurl/luds/blob/luds/06.md) requires the invoice to be generated with a description hash.

Most backends don't allow you to set the expiry, so I'm reluctant to make this user configurable. As for the description hash, there has been an [ongoing debate](https://github.com/lnurl/luds/pull/234) whether that should be removed or not. The majority of wallets don't care about it and most backends also don't let you set it (or in the case of CLN require you to know the entire description to be able to pay).

I think memo makes sense, description hash maybe too if the need is there (although I think it doesn't matter anymore, so I'd also ignore it), but for the expiry I'm doubtful that it should be a user choice.

gudnuf:
Agreed that `expiry` doesn't need to be in there if its not well supported.

If description hash gets removed from LUD 06, then I also don't think description hash is necessary. As it is now very few lightning address providers seem to validate it anyways.

AngusP:
> As it is now very few lightning address providers seem to validate it anyways.

That matches my experience, I *think* there might be some that will hash the full description and check that instead if there's a `d` tag but not `h`

thesimplekid:
I'm not sure this should added it creates an opportunity for the mint to learn more about the user and/or their intentions then is required to mint ecash which i think should be discouraged.

callebtc:
> I'm not sure this should added it creates an opportunity for the mint to learn more about the user and/or their intentions then is required to mint ecash which i think should be discouraged.

Good point but I think it should still be optional. The mint can also read all the memos of the invoices they are paying.

thesimplekid:
> Good point but I think it should still be optional so that mints can provide the same user experience as any other lightning wallet.

[...]

PR #144, opened 2024-06-27
NUT-04: Add optional description for PostMintQuoteBolt11Request

minibits-cash:
This PR extends PostMintQuoteBolt11Request by optional description field. Description is then ought to be passed to the lightning backend by the mint and included in the returned bolt11 invoice.

Considered two options for param name: memo and description, leaning to the latter as it is more generic and decoded bolt11 invoice uses the same term.


thesimplekid:
As mentioned on the dev call since this may not be supported by all backends it would be best to add signalling for it on the MintInfo endpoint. A boolean field could be added to the [Mint Settings](https://github.com/cashubtc/nuts/blob/main/04.md#settings) something like `set_description`. I also think it would be a good idea for the mint to return an error if a description is included and it is not supported.

thesimplekid:
closes https://github.com/cashubtc/nuts/issues/110

callebtc (approved):
LGTM. We should add Error codes but that doesn't block this PR

PR #146, opened 2024-07-02
NUT-08: mint fee return any overpaid amount

thesimplekid:
Change NUT08 to reflect that should return any over paid amount on a melt even if it is greater then the `fee_reserve` this avoids wallets having to `swap` before a `melt` to get the correct amount. 

- [x] Nutshell
- [x] CDK
- [ ] Nutmix
- [ ] gonuts
- [ ] ...


ngutech21:
Just one thing to consider: This is changing the existing behavior of a mint. So I think it would be a good idea to signal this change to the wallet by adding a flag to the info-endpoint like supports_overpaying: true or something like that. So the wallet knows exactly how the mint will react. 
Otherwise it will be hard for a wallet to figure out what the current behavior of the mint is. The mint is signaling that it supports NUT-08 but not how it will treat an overpayment.

thesimplekid:
> Just one thing to consider: This is changing the existing behavior of a mint. 

I don't think is actually changing the behavior of the mint and this is how nutshell implements it already @callebtc could you confirm?

05.md
```diff
@@ -9,7 +9,7 @@ Melting tokens is the opposite of minting tokens (see [NUT-04][04]). Like mintin
...
 
-In the second request, the wallet includes the `quote` id and provides `inputs` that sum up to `amount+fee_reserve` in the first response. For the method `bolt11`, the wallet can also include `outputs` in order for the mint to return overpaid Lightning fees (see [NUT-08][08]). The mint responds with a payment status `paid` and a `proof` of payment. If the request included `outputs`, the mint may respond with `change` for the overpaid fees (see [NUT-08][08]).
+In the second request, the wallet includes the `quote` id and provides `inputs` that sum up to `amount+fee+fee_reserve` in the first response. For the method `bolt11`, the wallet can also include `outputs` in order for the mint to return overpaid Lightning fees (see [NUT-08][08]). The mint responds with a payment status `paid` and a `proof` of payment. If the request included `outputs`, the mint may respond with `change` for the overpaid fees (see [NUT-08][08]).
```
gudnuf, line 12: I was confused for a while about what `fee` is exactly. is that supposed to be the keyset fee as defined in NUT02? If so, would it be more clear to say:

+In the second request, the wallet includes the `quote` id and provides `inputs` that sum up to `amount+fee+fee_reserve` in the first response where `fee` is calculated from the keyset's `input_fee_ppk` as described in [NUT02](https://github.com/cashubtc/nuts/blob/main/02.md)
```

callebtc:
Merging this long overdue PR @thesimplekid 

@elnosh @lescuer97 ping

callebtc (approved):

PR #155, opened 2024-08-21
Nostr Broadcast (or a better name TBD)

lollerfirst:
This NUT aims to address [this privacy concern](https://primal.net/e/note1qync806vlhhuy0tyuz6pmkryh2t6pw69frktvg5z7xgtz2hrvhzsjmz0a5) when wallets check for token melt status, while also possibly facilitating future implementations of the famously proposed [PoL scheme](https://gist.github.com/callebtc/ed5228d1d8cbaade0104db5d1cf63939) by calle. This is because the mint is effectively building a Proof of Liabilities attestation (or report) composed by all the nostr notes containing all of the secrets/blinded signatures.


XX.md
```diff
@@ -0,0 +1,94 @@
...
+Conversely, when the Mint publishes notes about freshly spent e-cash, the notes are simply broadcast to anyone who is listening, without creating a direct connection between the mint and any specific wallet.
+Then, when a wallet wants to check if a particular e-cash was claimed, it can query the Nostr notes published by the mint, searching for the inclusion of specific e-cash it sent. This query is also anonymous, as the wallet is simply retrieving publicly available information.
+As a result, the mint never learns which wallet is querying the information.
+
+### PoL Scheme
+
+The proposed protocol also streamlines the creation of a *Proof of Liabilities Report* (see [PoL scheme](PoL)), as each note represents a signed contribution to the "Mint Proof" or "Burn Proof" attestations of the mint for a particular epoch.
+
+## Mint Nostr Identity
+The public key for the Mint's nostr identity is the same as advertised in the `pubkey` field of the `GetInfoResponse` object returned by the Mint upon receiving a `v1/info` request, as specified in [NUT-06](06).
```
davidcaseria, line 23: We should note how the `pubkey` field is a compressed public key and how Nostr public keys are x-only.

Also, this is unrelated to this NUT, but I assumed the `pubkey` field was for the LN node ID, but maybe that's not the case.

XX.md
```diff
@@ -0,0 +1,94 @@
...
+``` 
+
+Where:
+* `name` can be the Mint's name, reachable URL or empty
+* `about` is the Mint's short description or empty
+* `picture` is the URL to an image for the Mint or empty.
+
+## Posting Notes
+
+The Mint regularly (about every 5 seconds) broadcasts events of `kind: 30078`, which is reserved for *"Application Specific Data"*. The Mint must also specificy a `created_at` unix timestamp in seconds along with the `content` of the note, which is initially set as follows:
```
davidcaseria, line 51: Kind 30078, as defined in [NIP-78](https://github.com/nostr-protocol/nips/blob/master/78.md), needs a `d` tag for the context. However, I also think this is an issue because this event will be replaced every 5 seconds. I don't think that's the intention, right?

XX.md
```diff
@@ -0,0 +1,94 @@
...
+
+[00]: 00.md
+[06]: 06.md
+[07]: 07.md
+[PoL]: https://gist.github.com/callebtc/ed5228d1d8cbaade0104db5d1cf63939
+
+## Use Cases
+
+### Ecash Transactions
+When `Alice` prepares a token to be sent to `Carol`, she can mark these tokens in her database as *pending*. She can then, periodically or upon user input, query its connected relays for notes referencing the Mint's pubkey as the author and a timestamp for filtering out older notes. `Alice` can then scan the notes' content and look for the inclusion of the spent secrets (sent to `Carol`). If it finds any, the token has been redeemed by `Carol` already, i.e., is `SPENT`. If the proof is not spendable anymore (and, thus, has been redeemed by `Carol`), she can safely delete the proof from her database.
```
davidcaseria, line 89: In CDK, I believe we call this status "reserved." This is not directly related to this NUT, but it would be good to have standardized terminology across implementations as much as possible.

XX.md
```diff
@@ -0,0 +1,94 @@
...
+
+### Facilitating Proof of Liabilities Reports
+![](https://user-images.githubusercontent.com/93376500/249383182-ed572841-cd78-40ea-b171-c1f768cd13dc.png)
+
+Each note represents signed contributions to "Mint proofs" (issued ecash) and "Burn proofs" (redeemed ecash) for the epoch specified in `epoch` field. The notes can then be compiled into a succint Proof of Liabilities report for each epoch by anyone requesting the publicly available information about the mint through nostr relays.
```
davidcaseria, line 94: I can imagine some cool public dashboards from this data. :slightly_smiling_face:

XX.md
```diff
@@ -0,0 +1,94 @@
...
+However, this approach has a significant privacy drawback: when the recipient attempts to claim the e-cash, they will reveal the same `Y` to the Mint, allowing it to potentially associate that token with both the sender's and recipient's identity (i.e. IP addresses). This compromises the protocol's inteded privacy guarantees.
+Conversely, when the Mint publishes notes about freshly spent e-cash, the notes are simply broadcast to anyone who is listening, without creating a direct connection between the mint and any specific wallet.
+Then, when a wallet wants to check if a particular e-cash was claimed, it can query the Nostr notes published by the mint, searching for the inclusion of specific e-cash it sent. This query is also anonymous, as the wallet is simply retrieving publicly available information.
+As a result, the mint never learns which wallet is querying the information.
+
+### PoL Scheme
+
+The proposed protocol also streamlines the creation of a *Proof of Liabilities Report* (see [PoL scheme](PoL)), as each note represents a signed contribution to the "Mint Proof" or "Burn Proof" attestations of the mint for a particular epoch.
```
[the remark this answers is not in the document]
lollerfirst: I thought BURN would be appropriate to describe the burning of tokens in both the swap and melt events, while MINT is used to describe all events that create Blind Signatures.

XX.md
lollerfirst: > We should note how the pubkey field is a compressed public key and how Nostr public keys are x-only
True.

> Also, this is unrelated to this NUT, but I assumed the pubkey field was for the LN node ID, but maybe that's not the case.

It would make sense if it could be specified as an env setting. In Nutshell it's derived from the seed. I am not sure of this but i believe you are required to have a seed because a bunch of other things depend on it.

XX.md
lollerfirst: Totally forgot about the `d` tag

XX.md
lollerfirst: I have taken this passage from [NUT-06](06.md)'s gospel and readapted it.

XX.md
lollerfirst: No that's not the intention. I thought this event would be just like a text note but invisible to normal clients. I didn't think posting successive ones would replace the older ones. If that's the case, we need another event kind.

davidcaseria: It probably makes sense to write a NIP for a specific event kind.

XX.md
thesimplekid: In CDK we use `RESERVED` and `PENDING` for different cases. A `PENDING` proofs is one that is currently involved in a transaction at the mint, on a `melt` when the ln payment is in the process of being paid this is generally quick and the proof is not pending for long but if a payment is stuck for examples it can remain in this pending state until the payment goes through or fails. 

`RESERVED` is used by the wallet to indicate when a proof as been included in a created token this is to protect it from being accidentally double spent and included in another transaction before whoever received the created token claims the proof. The mint is unaware of this `RESERVED` state and only has 3 state `UNSPENT`, `PENDING` and `SPENT`` while the wallet has the 4th `RESERVED` state.

I think its worth making this distinction for wallets and would change this to be `reserved`

XX.md
```diff
@@ -0,0 +1,102 @@
...
+    "name": str,
+    "about": str,
+    "picture": str
+}
+``` 
+
+Where:
+* `name` can be the Mint's name, reachable URL or empty
+* `about` is the Mint's short description or empty
+* `picture` is the URL to an image for the Mint or empty.
```
thesimplekid, line 54: I don't think we should just use the `kind: 0` for this, instead i would recommend we use a [replaceable event](https://github.com/nostr-protocol/nips/blob/fade0164f52033314bf0a5ef9bd63c2483afae9b/01.md?plain=1#L96) with  the mints NUT-06 info in it. This would also serve as a way for a mint to announce themselves on nostr.

XX.md
```diff
@@ -0,0 +1,102 @@
...
+```json
+[
+    {
+        "epoch": int,
+        "event": "MINT"|"BURN",
+        "contents": [(int, str), ...]
+    },
+    ...
+]
+```
```
thesimplekid, line 69: Similarly to the above comment it should be a different kind then a `kind: 1` as this isn't a general text not and is specific. Could even be a different kind for `MINT` and `BURN` the `e` tag of the previous event should also be included.

XX.md
```diff
@@ -0,0 +1,102 @@
...
+Where:
+* `epoch`: is the current number of key rotations
+* `event`: specifies whether this entry is a mint or burn
+* `contents`: depends on `event`:
+  + if `event` is `BURN`: array of tuples, each mapping integer amounts to hex-encoded spent secrets.
+  + if `event` is `MINT`: array of tuples, each mapping
+  integer amounts to compressed, serialized and hex-encoded blind signature points.
```
thesimplekid, line 76: What is the need for the mapping integer, could it not be just a list of secrets or `Y`'s?

XX.md
```diff
@@ -0,0 +1,102 @@
...
+```
+
+Upon startup, the Mint must broadcast a `kind: 0` event, reserved for setting *"Metadata"*. The event must have its `content` field set to the JSON-serialization of the following object:
+```json
+{
+    "name": str,
+    "about": str,
+    "picture": str
+}
+```
```
thesimplekid, line 49: As David mentioned above iit should be a different kind then a `kind: 1` as this isn't a general text note and is specific. Could even be a different kind for `MINT` and `BURN` the `e` tag of the previous event should also be included.

XX.md
lollerfirst: The amounts are necessary to reconstruct total assets vs. total liabilities.

XX.md
lollerfirst: I assume you are referring to the timely "broadcasting" section, here. 
What event kind would you suggest?

> the `e` tag of the previous event should also be included.

Good idea.

XX.md
lollerfirst: Fixed

XX.md
lollerfirst: > The amounts are necessary to reconstruct total assets vs. total liabilities.

For checking the melt state it's irrelevant.

lollerfirst:
I am thinking we should probably add another event case for letting the Mint (or Mint operator, manually) publish the signed outstanding balance for that particular epoch, to be consistent with the PoL scheme (see the example picture i've included)

XX.md
thesimplekid: >  What event kind would you suggest?

There aren't any that would fit this so an unreserved kind and then propose it as a NIP?

davidcaseria:
With the current event structure, a client must query for all the mint events since sending the token, decode the contents, and check when it is claimed. This may be an overwhelming number of events for the client to process for a large mint or token that has not been claimed for a while. Can the event be structured to be partially filtered on the relay
while not leaking privacy?

lollerfirst:
@davidcaseria 
> While not leaking privacy

If they are filtered on the relay, and there is cooperation between the relay and the Mint (or they are operated by the same entity) then you are back to square one. The idea here is the wallet should not communicate `x` or `Y` to anybody else other than the receiver of the token.

What we COULD do is have the client request a relay the Bloom filter for a particular mint and epoch, where all the burnt secrets have been inserted. The relays calculate (or have it cached) the Bloom filter (with predefined parameters) for the specified epoch and then send it to the client.

Then looking for spent secrets in the Bloom filter is much easier for the client if the BURN event since timestamp have been many.

lollerfirst:
**TO BE CLEAR**:
The Mint *BATCHES* the `MINT` and `BURN` events to be broadcast.
I thought it was obvious from the wording of "regularly"  updates but apparently it's not.
We don't want to publish events on every client request the Mint processes.

lollerfirst:
I have opened a PR for the NIP: https://github.com/nostr-protocol/nips/pull/1526

AngusP (commented):
I don't think I agree conceptually with using a stream of Nostr events for this use -- you're right that *"when the recipient attempts to claim the e-cash, they will reveal the same `Y` to the Mint, allowing it to potentially associate that token with both the sender's and recipient's identity"*; However, adding a dependency on Nostr and streaming many notes for spends is a lot of overhead (and data) to address the problem.

> When a sender transfers e-cash to a recipient, they need to confirm that the recipient has claimed the payment in order to safely mark the proof in their wallet as `SPENT`.

This is of-course true, but I don't think it's an *immediate* concern -- it's OK if a token is in some `PROBABLY_SPENT` state for some-time™, as we only want to claw it back if the recipient never actually spent it. So we can check later in the next day or week and assume the payment 'failed' if the recipient hasn't spent it. Additionally, there's an open question of "how long should I wait for a recipient to claim a token before I take it back" as a legitimate recovery for a failed payment rather than basically stealing from someone who thinks you've paid them and was going to claim the token but hasn't yet, especially in a offline/poor-connectivity use-case. Ideally a recipient doesn't give you the stuff you're paying for before they've claimed the token, but *ideally* isn't guaranteed -- best real-world argument I can give for this is that merchants do this all the time with fiat payment cards.

If we expect that for some payments the recipient may claim hours after we sent them a token, then with Nostr events being published every ~5 seconds by the mint, that's also a lot of events and network requests a wallet may need to chew through to check if the recipient claimed.

That being the case, then a simple solution like the mint publishing a daily spend bloom filter (or similar construct) for all the tokens spent on a given day would allow the wallet to check for a spend without revealing to the mint which `Y` in-particular it's interested in (Bloom filters can have false-positives so we'd have to handle that edge-case). Or even better, if the mint publishes a [Proof of Liabilities](https://gist.github.com/callebtc/ed5228d1d8cbaade0104db5d1cf63939) daily, then you know the token is unspent by the recipient if there's a proof in the PoL for it, and is spent if there is not. I like how this also 'forces' PoL to be implemented and checked frequently by wallets, because it gives them an incentive to do so other than to check if they might be getting rugged 🤣.

Lastly, it's also 'kinda nice' that there isn't necessarily a public record of "when what token got spent" -- obviously the sender, recipient and mint could know, but otherwise it's private. If the mint publishes events to Nostr, then that's no-longer the case?

lollerfirst:
> If the mint publishes events to Nostr, then that's no-longer the case?
The information published only makes sense to the sender and receiver. Others can only see "some" token of "some" amount has been spent. 

But to be quite honest I don't believe this is a good idea anymore.

PR #131, opened 2024-06-01
error codes

thesimplekid:
Tracking progress:
- [ ] nutshell
- [x] CDK
- [ ] cashu-ts
- [ ] nutmix
- [ ] gonuts
- [ ] ...

PR #161, opened 2024-09-09
refactor: one error code file

thesimplekid:
Alternative to #131 

I found having having the error codes split across multiple files to be confusing since the same error can be used in multiple nuts. This change keep all the error codes in one file while noting what nut(s) each error code is relevant to.

Open to comments on which one others prefer.


callebtc (approved):
LGTM, I agree it's better in one file.

elnosh (approved):
I like this better as well

PR #170, opened 2024-10-04
Bolt12

thesimplekid:
- [ ] CDK https://github.com/cashubtc/cdk/pull/709
- [ ] Nutshell
- [ ] cashu-ts
- [ ] gonuts
- [ ] nutmix

### DIFF
--- a/23.md
+++ b/23.md
@@ -1,0 +1,243 @@
+# NUT-23: Mint tokens via BOLT12
+
+`optional`
+
+`depends on: NUT-20`
+
+---
+
+Similar to [NUT-04][04], which covers minting via BOLT11 invoices, minting via [BOLT12](https://github.com/lightning/bolts/blob/master/12-offer-encoding.md) is also a two-step process: requesting a mint quote and minting tokens. This document describes both steps, focusing on BOLT12-specific considerations.
+
+# Mint quote
+
+To request a mint quote, the wallet of `Alice` makes a `POST /v1/mint/quote/bolt12`.
+
+```http
+POST https://mint.host:3338/v1/mint/quote/bolt12
+```
+
+The wallet of `Alice` includes the following `PostMintQuoteBolt12Request` data in its request:
+
+```json
+{
+  "amount": <int|null>,
+  "unit": <str_enum["sat"]>,
+  "description": <str|null>,
+  "expiry": <int|null>,
+  "single_use": <bool>,
+  "pubkey": <str>
+}
+```
+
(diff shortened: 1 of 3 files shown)


19.md
```diff
@@ -0,0 +1,199 @@
...
+{
+  "19": {
+    "methods": [
+      <MeltMethodSetting>,
+      ...
+    ],
+    "disabled": <bool>
+  }
+}
+```
```
thesimplekid, line 170: Should these settings be under there own nip number or added to the 04/05 headings

lollerfirst:
Following up on your presentation at BTC++ we could make it so that a bolt12 quote has a different status enum:
* `EMPTY`
* `MINTABLE`
* `RETIRED`

The states can go from `EMPTY` to `MINTABLE` and back, but once the quote is `RETIRED` no more state changes.
One might always expect to save and re-use the same quote to mint ecash multiple times, as that is the purpose of the bolt12 (to be paid multiple times).

And then introduce a new endpoint through which one can retire the quote.

19.md
```diff
@@ -0,0 +1,174 @@
...
+  "request": <str>,
+  "expiry": <int>,
+  "amount_paid": <int>,
+  "amount_issued": <int>
+}
+```
+
+Where `quote` is the quote ID and `request` is the bolt12 offer. `expiry` is the Unix timestamp until which the mint quote is valid. `amount_paid` is the amount that has been paid to the mint via the bolt12 offer. `amount_issued` is the amount of ecash that has been issued for the given mint quote. `amount_paid` - `amount_issued` represents the amount of ecash a wallet can still mint.
+
+Note: `quote` is a **unique and random** id generated by the mint to internally look up the payment state. `quote` **MUST** remain a secret between user and mint and **MUST NOT** be derivable from the payment request. A third party who knows the `quote` ID can front-run and steal the ecash that this operation mints.
```
ok300, line 45: Would it make sense to specify the length and type of the `quote`? For example, 64 char hex string?

IMO that makes it clear for implementors and avoids quick-and-dirty (and-insecure) solutions like incrementing from 1, or current unix timestamp, etc.

This also avoids complications if some mints decide to try out exotic characters, which some wallets may not expect.

20.md
```diff
@@ -0,0 +1,191 @@
...
+  "expiry": <int>,
+  "amount_paid": <int>,
+  "amount_issued": <int>,
+  "pubkey": <str>
+}
+```
+
+Where `quote` is the quote ID and `request` is the bolt12 offer. `expiry` is the Unix timestamp until which the mint quote is valid. `amount_paid` is the amount that has been paid to the mint via the bolt12 offer. `amount_issued` is the amount of ecash that has been issued for the given mint quote. `amount_paid` - `amount_issued` represents the amount of ecash a wallet can still mint.
+
+Note: `quote` is a **unique and random** id generated by the mint to internally look up the payment state. `quote` **MUST** remain a secret between user and mint and **MUST NOT** be derivable from the payment request. A third party who knows the `quote` ID can front-run and steal the ecash that this operation mints.
```
thesimplekid, line 53:
```suggestion
```

thesimplekid:
> Should we pick this back up?

Yes that's my plan in CDK it should be most of the way there. Should be a fun rebase. I just rebased this pr on main.

thesimplekid:
Superseded by https://github.com/cashubtc/nuts/pull/259

PR #173, opened 2024-10-09
feat: add amount to nut05

thesimplekid:
In order to support paying an amount-less invoice, add amount to quote request.

### DIFF
--- a/05.md
+++ b/05.md
@@ -28,10 +28,11 @@ The wallet `Alice` includes the following `PostMeltQuoteBolt11Request` data in i
 {
   "request": <str>,
   "unit": <str_enum["sat"]>
+  "amount": <int|null>
 }
 ```
 
-Here, `request` is the bolt11 Lightning invoice to be paid and `unit` is the unit the wallet would like to pay with.
+Here, `request` is the bolt11 Lightning invoice to be paid and `unit` is the unit the wallet would like to pay with. `amount` is the amount to be paid if the invoice does not have a defined amount.
 
 The mint `Bob` then responds with a `PostMeltQuoteBolt11Response`:
 


prusnak:
I think we should specify whether amount can be provided if bolt11 invoice contains an amount.

There are 3 options:
- amount must not be provided
- amount can be provided but is ignored
- amount can be provided but must match the amount in bolt11 invoice

I like the last option the most.

05.md
```diff
@@ -28,10 +28,11 @@ The wallet `Alice` includes the following `PostMeltQuoteBolt11Request` data in i
 {
   "request": <str>,
   "unit": <str_enum["sat"]>
```
prusnak, line 30: missing comma at the end of line

thesimplekid:
> I think we should specify whether amount can be provided if bolt11 invoice contains an amount.

Good point, I've updated wording to reflect this.

prusnak (approved):

elnosh (approved):

callebtc:
Setting for the info endpoint (some backends can't pay amountless invoices)

thesimplekid:
> Setting for the info endpoint (some backends can't pay amountless invoices)

Added in 6628ff1758c8bb2ec7989c7e3c56936e75e11fac

05.md
```diff
@@ -27,11 +27,12 @@ The wallet `Alice` includes the following `PostMeltQuoteBolt11Request` data in i
...
 {
   "request": <str>,
-  "unit": <str_enum["sat"]>
+  "unit": <str_enum["sat"]>,
+  "amount": <int|null>
 }
 ```
 
-Here, `request` is the bolt11 Lightning invoice to be paid and `unit` is the unit the wallet would like to pay with.
+Here, `request` is the bolt11 Lightning invoice to be paid and `unit` is the unit the wallet would like to pay with. `amount` is an integer or null, representing the amount to be paid. If the invoice already defines an amount, the amount in the PostMeltQuoteBolt11Request **MUST** match the amount in the invoice exactly. The request may only be in amountless invoice if support for amountless invoices is signaled in the `MeltMethodSettings`.
```
callebtc, line 35: Something is wrong with that last sentence. Also, it sounds like `amount` should always be passed. 

Should be more precise: `amount` can be passed to pay amountless bolt11 invoices. For bolt11 invoices with amount, the `amount` field will be ignored.

prusnak (approved):
re-ACK after the new changes

05.md
```diff
@@ -27,11 +27,16 @@ The wallet `Alice` includes the following `PostMeltQuoteBolt11Request` data in i
 ```json
 {
   "request": <str>,
-  "unit": <str_enum["sat"]>
+  "unit": <str_enum["sat"]>,
+  "options": {
+    "amountless": {
+      "amount_msat": <int>
+    }
+  } // Optional
```
prusnak, line 35: perhaps this belongs one row or four rows above?

thesimplekid, line 35:
```suggestion
  "options": { // Optional
    "amountless": {
      "amount_msat": <int>
    }
  }
  ```
  Think this is better?
  
  
```

prusnak: I think it is because previously I thought the whole block is optional.

error_codes.md
```diff
@@ -8,6 +8,8 @@
 | 11002 | Transaction is not balanced (inputs != outputs) | [NUT-02][02], [NUT-03][03], [NUT-05][05] |
 | 11005 | Unit in request is not supported                | [NUT-04][04], [NUT-05][05]               |
 | 11006 | Amount outside of limit range                   | [NUT-04][04], [NUT-05][05]               |
+| 11007 | Amountless invoice is not supported             | [NUT-05][05]                             |
+| 11008 | Amount in request does not equal invoice        | [NUT-05][05]                             |
```
elnosh, line 12: I was implementing these changes and just noticed, these error codes will collide with the ones in this PR: https://github.com/cashubtc/nuts/pull/223. So I guess one of the two should change.

elnosh:
gonuts: https://github.com/elnosh/gonuts/pull/137 
just pending resolution on the conflicting error codes with https://github.com/cashubtc/nuts/pull/223

PR #174, opened 2024-10-10
error code for failed lightning payments

elnosh:
add error code for failed lightning payments in melts. If wallets see this error they can be sure the proofs are not stuck in a pending state.


thesimplekid:
I noticed nutshell also has a general lightning error 20000 not sure what its used for? Should that be added as well, or is the only error we want to return a payment failed?

thesimplekid (approved):

callebtc:
> I noticed nutshell also has a general lightning error 20000 not sure what its used for?

nutshell uses 20000 as a generic "Lightning error" – we should probably define it as "generic payment error" but nutshell should use a specific error code.

Issue #62, opened 2023-11-11
NUT-06: Mint should announce multiple endpoints that it is reachable from


callebtc:
Mint should announce https / Tor addresses in the info response at which it is reachable from.

jooray:
Adding additional protocols in addition to http would be benificial, if I can register a handler to use it in nutshell

Egge21M:
Can we use https://github.com/nostr-protocol/nips/pull/1110 as a way for a mint to announce changes to their npub?

1. Client connects to mint
2. Clients retrieves nostr pubkey from /info
3. Client periodically checks the pubkey on nostr for updates

callebtc:
> Can we use [nostr-protocol/nips#1110](https://github.com/nostr-protocol/nips/pull/1110) as a way for a mint to announce changes to their npub?
> 
> 1. Client connects to mint
> 2. Clients retrieves nostr pubkey from /info
> 3. Client periodically checks the pubkey on nostr for updates

That sounds like a good idea but I think it would be an overkill right now. If we go further down the nostr route, the mint should probably announce a lot more than just the mint URLs (whole info endpoint for example, and keys etc). Since the chance of your DNS getting rugged is really small, I wouldn't consider this a pressing issue right now though.

callebtc:
Self-bump

prusnak:
Adding multiple URLs to the config and to the /info response is trivial.

But how should a mint behave if there are multiple mint URLs defined?

Should it bake all URLs into the cashu tokens? 🤔 The `m` field - mint URL - currently does not support a list. Maybe it should?

callebtc:
> But how should a cashu token behave if there are multiple URLs defined for a mint?
> 
> Should we bake all URLs into the cashu tokens? 🤔 The `m` field - mint URL - currently does not support a list. Maybe it should?

Good question, I think we can get away with a single URL for now. Wallets that know multiple endpoints (or fetch the info) can choose a transport they like when they receive a token. The URL in the token should always be assumed be reachable.

prusnak:
PR in https://github.com/cashubtc/nuts/pull/175

ok300:
> Wallets that know multiple endpoints (or fetch the info) can choose a transport they like when they receive a token. The URL in the token should always be assumed be reachable.

Doesn't this make Offline Receive trickier? The receiver might only fetch the info much later, when the baked mint URL may not be reachable.

[...]

PR #175, opened 2024-10-11
NUT-06: add urls field

prusnak:
fixes https://github.com/cashubtc/nuts/issues/62

- [x] CDK - https://github.com/cashubtc/cdk/pull/398
- [x] Nutshell - https://github.com/cashubtc/nutshell/pull/638

### DIFF
--- a/06.md
+++ b/06.md
@@ -39,6 +39,10 @@ With the mint's response being of the form `GetInfoResponse`:
   ],
   "motd": "Message to display to users.",
   "icon_url": "https://mint.host/icon.jpg",
+  "urls": [
+    "https://mint.host",
+    "http://mint8gv0sq5ul602uxt2fe0t80e3c2bi9fy0cxedp69v1vat6ruj81wv.onion"
+  ],
   "time": 1725304480,
   "nuts": {
     "4": {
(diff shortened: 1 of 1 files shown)


06.md
```diff
@@ -39,6 +39,10 @@ With the mint's response being of the form `GetInfoResponse`:
   ],
   "motd": "Message to display to users.",
   "icon_url": "https://mint.host/icon.jpg",
+  "urls": [
```
prusnak, line 42: should we call the field `urls` or `endpoints` or something else?

thesimplekid:
I think we should not just list urls but follow a struct more similar to the contact feild for examples:

```json 
{
  "endpoints": [
    {
      "method": "http",
      "endpoint": "https://fake.mint.dev"
    },
    {
      "method": "nostr",
      "endpoint": "nprofile1qs..."
    },
    {
      "method": "tor",
      "endpoint": "abcdefghijklmn.onion"
    }
  ]
}

```

prusnak:
> I think we should not just list urls but follow a struct more similar to the contact feild for examples:

🤷 URI schemes were invented exactly for this reason.I am just not a big fan of using dicts and I think it would have been much better if contact field also used URI schemes instead of the JSON dict abomination we have now. :-) That said I don't mind either way and I prefer consistency over correctness. If there is consensus to go with the JSON dict instead of the URI scheme, I will update the PR.

ok300:
`TokenV4` has a field `m` for the mint URL: https://cashubtc.github.io/nuts/00/#token-format_1

Consider the following scenario:

Let's say mint Bob in accessible on 2 URLs. If Alice has a token for Bob's mint (with "Bob URL 1" embedded) and sends this token to Carol, then Carol will only have "Bob URL 1" to get to the mint. If the mint becomes unreachable on URL 1 but remains reachable on it's URL 2, Carol won't know the 2nd URL and won't be able to redeem it.

Is that true, or am I missing something?

The fix would be for `TokenV4` to include all of Bob's known URLs, but that would mean making the token bigger (esp. with tor URLs).

prusnak:
> Consider the following scenario:

That is exactly what I described here:

https://github.com/cashubtc/nuts/issues/62#issuecomment-2404500881

thesimplekid:
Merges in cdk https://github.com/cashubtc/cdk/pull/398

prusnak:
Merged in nutshell https://github.com/cashubtc/nutshell/pull/638

thesimplekid (approved):

PR #178, opened 2024-10-18
NUT-18: fix typo

prusnak:
json says `"r"` for single use payments while the description says `"s"`

I think it should be `"s"`


callebtc (approved):
Thank you!

PR #182, opened 2024-10-30
Define Keyset ID V2

davidcaseria:


02.md
```diff
@@ -75,8 +77,9 @@ An example implementation in Python:
 ```python
 def derive_keyset_id(keys: Dict[int, PublicKey]) -> str:
-    pubkeys_concat = b"".join([p.serialize() for p in sorted_keys.values()])
-    return "00" + hashlib.sha256(pubkeys_concat).hexdigest()[:14]
+    keyset_id_bytes = b"".join([p.serialize() for p in sorted_keys.values()])
+    keyset_id_bytes += b"unit:sat"
```
prusnak, line 81: I am wondering ... since we already depend for CBOR (for Token v4), maybe we can use CBOR here too instead of this custom serialization?

02.md
```diff
@@ -14,10 +14,12 @@ A mint can have multiple keysets at the same time. For example, it could have on
...
+A keyset `id` is an identifier for a specific keyset. It can be derived by anyone who knows the set of public keys of a mint. Wallets **MAY** compute the keyset `id` for a given keyset by themselves to confirm that the mint is supplying the correct keyset ID (see below).
 
 The keyset `id` is in each `Proof` so it can be used by wallets to identify which mint and keyset it was generated from. The keyset field `id` is also present in the `BlindedMessages` sent to the mint and `BlindSignatures` returned from the mint (see [NUT-00][00]).
 
+To save space, a `Token`, as definied in [NUT-00][00], **SHOULD** contain an abbreviated version of the keyset `id` in the `Proof`. The length of the abbreviated `id` **MUST** be at least two bytes (i.e., the version byte and first byte of the hash). However, when the token is serialized, the chosen `id` length **MUST** not be ambiguous within the mint's existing keyset `id`s. A wallet **MUST** resolve the abbreviated keyset `id` to the full `id`. If the abbreviated `id` is ambiguous (i.e., multiple keyset `id`s are resolvable), the wallet **MUST** error. The full keyset `id` is only needed when the wallet interacts with the mint.
```
prusnak, line 21: What happens to the tokens when a mint generates a new keyset that collides with the already emitted tokens?

Let's say the old keyset id is `01cb2343....`, the tokens use `01cb` in the `i` field and then a new keyset is created which has id `01cb948...`. The chance of this happening is 1:256 (around 0.4%) - if you generate more keysets, the chance is significantly higher.

02.md
davidcaseria: What would the structure of the CBOR be?

callebtc:
Thank you!

Please: 
- add a verbose description in the PR what this is about
- we should keep at least 7 bytes, as others have mentioned
- this PR seems to *replace* the old keyset ID instead of adding a new version. Everything would break. We should instead add a new additional version.

02.md
```diff
@@ -14,10 +14,14 @@ A mint can have multiple keysets at the same time. For example, it could have on
...
+A keyset `id` is an identifier for a specific keyset. It can be derived by anyone who knows the set of public keys of a mint. Wallets **MAY** compute the keyset `id` for a given keyset by themselves to confirm that the mint is supplying the correct keyset ID (see below).
 
 The keyset `id` is in each `Proof` so it can be used by wallets to identify which mint and keyset it was generated from. The keyset field `id` is also present in the `BlindedMessages` sent to the mint and `BlindSignatures` returned from the mint (see [NUT-00][00]).
 
+To save space, a `Token`, as definied in [NUT-00][00], **SHOULD** contain an abbreviated version of the keyset `id` in the `Proof`. The length of the abbreviated `id` **MUST** be at least eight bytes (i.e., the version byte and first byte of the hash). However, when the token is serialized, the chosen `id` length **MUST** not be ambiguous within the mint's existing keyset `id`s. A wallet **MUST** resolve the abbreviated keyset `id` to the full `id`. If the abbreviated `id` is ambiguous (i.e., multiple keyset `id`s are resolvable), the wallet **MUST** error. The full keyset `id` is only needed when the wallet interacts with the mint.
```
ok300, line 21: > ... the abbreviated `id` **MUST** be at least eight bytes (i.e., the version byte and first byte of the hash)

8 != 1 + 1

02.md
```diff
@@ -56,12 +60,32 @@ Notice that since transactions can spend inputs from different keysets, the sum
...
+5 - prefix it with a keyset ID version byte
+```
+
+An example implementation in Python:
+
+```python
+def derive_keyset_id(keys: Dict[int, PublicKey]) -> str:
+    sorted_keys = dict(sorted(keys.items()))
+    keyset_id_bytes = b"".join([p.serialize() for p in sorted_keys.values()])
+    keyset_id_bytes += b"unit:sat"
```
ok300, line 83: Once optimization could be to change `b"unit:sat"` -> `b"u:sat"` or even just `b"sat"`. Might also be worth specifying the unit should be lowercased first.

02.md
davidcaseria: Oops. Thanks!

02.md
davidcaseria: I don't think we have specified the units as case-insensitive yet. We should probably also do that if we include this change, right?

02.md
```diff
@@ -148,24 +172,24 @@ To receive the public keys of a specific keyset, a wallet can call the `GET /v1/
 
 Request of `Alice`:
 
-We request the keys for the keyset `009a1f293253e41e`.
+We request the keys for the keyset `01c9c20fb8b348b389e296227c6cc7a63f77354b7388c720dbba6218f720f9b785`.
```
lollerfirst, line 175: What happens if a wallet that does not yet support V2 keysets makes a `GET v1/keys/{8-byte-keyset-id}` request?

davidcaseria: A request with a keyset ID prefix `00` should still be supported. A wallet that doesn't support keyset IDs with a prefix of `01` shouldn't get to where they are making this request since the keyset ID parsing will have already failed if the wallet is checking the version.

If a wallet doesn't check the version and sends a request for an 8-byte keyset ID from a token, we can consider having the mint resolve it for the wallet from the URL. But that feels like an optional requirement for the mint.

Does that address your question?

02.md
ok300: Since the keyset ID commits to the unit, then it would make sense IMO.

If a client derives the keyset ID using a different capitalization, or if the mint changes the capitalization later on, the keyset ID would be invalid.

02.md
```diff
@@ -14,10 +14,14 @@ A mint can have multiple keysets at the same time. For example, it could have on
...
 
 The keyset `id` is in each `Proof` so it can be used by wallets to identify which mint and keyset it was generated from. The keyset field `id` is also present in the `BlindedMessages` sent to the mint and `BlindSignatures` returned from the mint (see [NUT-00][00]).
 
+To save space, a `Token`, as defined in [NUT-00][00], **SHOULD** contain an abbreviated version of the keyset `id` in the `Proof`. The length of the abbreviated `id` **MUST** be at least eight bytes (i.e., the version byte and the first seven bytes of the hash). However, while unlikely, when the token is serialized, the chosen `id` length **MUST** not be ambiguous within the mint's existing keyset `id`s. A wallet **MUST** resolve the abbreviated keyset `id` to the full `id`. If the abbreviated `id` is ambiguous (i.e., multiple keyset `id`s are resolvable), the wallet **MUST** error. The full keyset `id` is only needed when the wallet interacts with the mint.
```
callebtc, line 21: This seems to vague and leaves room for interpretation. It could simply say that the token must include the first 8 bytes / 16 hex characters of the keyset ID it originates from.

callebtc: We could make it more explicit and call it `s_id` which is `id_hex[:16]` or `id_bytes[:8]`.

02.md
davidcaseria: @callebtc what do you think about units being case-insensitive and hashing the lowercase unit value?

lollerfirst:
Should we maybe add a paragraph to this that describes the way the Mint handles proofs with Short IDv2?
The Mint calculates each keyset's short IDv2 in addition to IDv2 and uses the short IDv2s for lookups in cases where it receives proofs with short IDv2.

prusnak:
Quite honest, I am not sure such small change justifies creating a new keyset version and the headache attached to it.

Maybe we should keep this improvement in mind and apply it together with some future proposal that changes something significant?

davidcaseria:
@prusnak this was suggested to enable wallets to store proofs without referencing mint URLs. @thesimplekid @ok300 do you think this is still necessary?

thesimplekid:
> this was suggested to enable wallets to store proofs without referencing mint URLs.

The longer keyset ids to reduce the chance of a collision and the keyset expiry are things we've talked about awhile and I would like to see, I think it makes sense to move forward with this. I cant think of anything else we wanted to include that we should hold this for, but if there is I am open to it.

02.md
thesimplekid: We should specify we hash the lower case.

prusnak:
> the keyset expiry are things we've talked about awhile and I would like to see, I think it makes sense to move forward with this

I agree. This is exactly the kind of change I was mentioning when I said to "justify" creating a new keyset version.

ok300:
A few NITs, otherwise it's an ACK from me:

The specified behavior is a bit ambiguous in 2 places.

> The keyset `id` is in each `Proof` so it can be used by wallets to identify which mint and keyset it was generated from.

This doesn't say if it's the long or short ID. It should probably say "short ID" or rename the variable to `s_id` for clarity.

> To save space, a `Token`, as defined in [NUT-00][00], **SHOULD** contain an abbreviated version of the keyset `id` in the `Proof`, (the `s_id`).

This makes it sound preferable, but optional to use the shorter `s_id` in the proof, whereas the PR text indicates it's mandatory, since the Token size is kept unchanged.

lollerfirst:
@davidcaseria I think it would be better if the long keyset ID was 32-bytes:
* Keyset version byte
* 31-bytes of the 32-bytes of the SHA-256 digest

lollerfirst:
CDK draft: https://github.com/cashubtc/cdk/pull/702

prusnak:
> I think it would be better if the long keyset ID was 32-bytes:
> 
> * Keyset version byte
> * the first 31-bytes out of the 32-bytes SHA-256 digest

why?

lollerfirst:
@prusnak For simmetry w.r.t. V1 so that the length is still a power of 2. I think it's less confusing if we keep the same process but only change what is needed to be changed.

prusnak:
> I think it's less confusing if we keep the same process and only change what is needed to be changed.

It is non-standard and does not bring any single benefit imo. Only headache.

ok300:
IMO there is an ambiguity in the new spec:

> To save space, a `Token`, as defined in [NUT-00][00], **SHOULD** contain an abbreviated version of the keyset `id` in the `Proof`, (the `s_id`).

This implies the `Proof`s in a `Token` could use either the short ID or the long ID.

Is there any reason where the long ID makes sense?

If not, I'd suggest to change that `SHOULD` to a `MUST`.

---

The only reason I can think of is to avoid short ID collisions, but that's already covered by another section:

> If the abbreviated `s_id` is ambiguous (i.e., multiple keyset `id`s are resolvable), the wallet **MUST** error. The full keyset `id` is only needed when the wallet interacts with the mint.

02.md
```diff
@@ -14,10 +14,14 @@ A mint can have multiple keysets at the same time. For example, it could have on
...
+A keyset `id` is an identifier for a specific keyset. It can be derived by anyone who knows the set of public keys of a mint. Wallets **MAY** compute the keyset `id` for a given keyset by themselves to confirm that the mint is supplying the correct keyset ID (see below).
 
 The keyset `id` is in each `Proof` so it can be used by wallets to identify which mint and keyset it was generated from. The keyset field `id` is also present in the `BlindedMessages` sent to the mint and `BlindSignatures` returned from the mint (see [NUT-00][00]).
 
+To save space, a `Token`, as defined in [NUT-00][00], **SHOULD** contain an abbreviated version of the keyset `id` in the `Proof`, (the `s_id`). The length of the `s_id` **MUST** be eight bytes (i.e., the version byte and the first seven bytes of the hash: `id_bytes[:8]` or `id_hex[:16]`). A wallet **MUST** resolve the abbreviated keyset `id` to the full `id`. If the abbreviated `s_id` is ambiguous (i.e., multiple keyset `id`s are resolvable), the wallet **MUST** error. The full keyset `id` is only needed when the wallet interacts with the mint.
```
ok300, line 21:
```suggestion
### Short ID

The short keyset ID (`s_id`) is defined as the abbreviated representation of the full Keyset ID. The length of the `s_id` is eight bytes (i.e., the version byte and the first seven bytes of the hash: `id_bytes[:8]` or `id_hex[:16]`).

The `s_id` is only relevant for wallets, when encoding and decoding `Tokens`. To save space, a `Token`, as defined in [NUT-00][00], **SHOULD** use the short keyset ID `s_id` in the `Proof`. The recommended exception to that is when the `Proofs` reference a keyset where the `s_id` is ambiguous on that mint; in that case, the sending wallet **SHOULD** instead use the full keyset ID. Receiving wallets **MUST** therefore be able to decode both `Tokens` with a short keyset ID, as well as those with a full keyset ID.

When receiving a `Token` with a `s_id`, a wallet **MUST** resolve the abbreviated `s_id` to the full `id`. If the `s_id` is ambiguous (i.e., multiple keyset `id`s are resolvable on the same mint), the wallet **MUST** error. Otherwise, if the `Token` is successfully parsed, the wallet **SHOULD** save the full-length keyset `id` with proofs in its database.

The mint is unaware of the `s_id`. All API calls exposed by the mint use the full keyset ID.
```

As discussed, the ambiguity mentioned above ( https://github.com/cashubtc/nuts/pull/182#issuecomment-2773223419 ) is intentional and the spec should allow for both short and long Keyset IDs in the `Token`.

@davidcaseria I re-worded the paragraph that originally confused me. It now spells out in more detail who should use the `s_id` when.

ok300 (approved):

02.md
```diff
@@ -14,10 +14,22 @@ A mint can have multiple keysets at the same time. For example, it could have on
...
+
+The `s_id` is only relevant for wallets, when encoding and decoding `Tokens`. To save space, a `Token`, as defined in [NUT-00][00], **SHOULD** use the short keyset ID `s_id` in the `Proof`. The recommended exception to that is when the `Proofs` reference a keyset where the `s_id` is ambiguous on that mint; in that case, the sending wallet **SHOULD** instead use the full keyset ID. Receiving wallets **MUST** therefore be able to decode both `Tokens` with a short keyset ID, as well as those with a full keyset ID.
+
+When receiving a `Token` with a `s_id`, a wallet **MUST** resolve the abbreviated `s_id` to the full `id`. If the `s_id` is ambiguous (i.e., multiple keyset `id`s are resolvable on the same mint), the wallet **MUST** error. Otherwise, if the `Token` is successfully parsed, the wallet **SHOULD** save the full-length keyset `id` with proofs in its database.
+
+The mint is unaware of the `s_id`. All API calls exposed by the mint use the full keyset ID.
+
+A Wallet **SHOULD** save the full-length keyset `id` with proofs in its database.
```
ok300, line 31:
```suggestion
```

NIT: Duplicate line, it's already included in the last part of the paragraph on L27 above.

lollerfirst (approved):

lollerfirst:
Good to merge??

thesimplekid:
ACK 0f3cef68eaab308d07a441f55ff37540f864faa5

thesimplekid:
closes #229

thesimplekid:
Merged in cdk https://github.com/cashubtc/cdk/pull/702

thesimplekid (approved):

lollerfirst:
@davidcaseria  I think the safest option is to version the secret derivation (like we're doing with the keysets) and then specify that wallets should follow:
- version 1 (legacy, with BIP32) for keysets with ID v1
- version 2 (the new one, with HMAC-SHA512) for keysets with ID v2
This way wallets have immediate way of knowing which derivation to use when generating secrets for a particular keyset, and we eliminate all "counter" related problems.

02.md
```diff
@@ -14,10 +14,20 @@ A mint can have multiple keysets at the same time. For example, it could have on
 
 ### Keyset ID
 
-A keyset `id` is an identifier for a specific keyset. It can be derived by anyone who knows the set of public keys of a mint. Wallets **CAN** compute the keyset `id` for a given keyset by themselves to confirm that the mint is supplying the correct keyset ID (see below).
+A keyset `id` is an identifier for a specific keyset. It can be derived by anyone who knows the set of public keys of a mint. Wallets **MAY** compute the keyset `id` for a given keyset themselves to confirm that the mint is supplying the correct keyset ID (see below).
```
davidcaseria: Change to **SHOULD**

13.md
```diff
@@ -12,11 +12,193 @@ In this document, we describe the process that allows wallets to recover their e
...
+
+1. `message = b"Cashu_KDF_HMAC_SHA512" || keyset_id_bytes || counter_k_bytes || derivation_type_byte`, where:
+   - `"Cashu_KDF_HMAC_SHA512"` is the domain separation or purpose string, encoded to bytes as UTF-8.
+   - `keyset_id_bytes` are the raw bytes of the keyset_id (hex decoded).
+   - `counter_k_bytes` is the counter encoded as an unsigned 64-bit integer in big-endian format.
+   - `derivation_type_byte` is exactly 1 byte specifying the type of derivation required:
+     - `0` for secrets
+     - `1` for blinded messages
+2. `hmac_digest = HMAC_SHA512(seed, message)`, where `HMAC_SHA512` is the [hash-based message authentication code](https://en.wikipedia.org/wiki/HMAC), using SHA-512 as the hashing algorithm.
+3. `secret = hmac_digest[:32]` and `r = hmac_digest[32:]`.
```
Egge21M: Shouldn't these both be hmac_digest[:32] as the `derivation_type_byte` now creates different outputs based on derivation type?

a1denvalu3: Yes you're correct.

a1denvalu3:
```suggestion
3. `secret = hmac_digest[:32]` or `blinding_factor = hmac_digest[:32]`.
```
@davidcaseria

02.md
```diff
@@ -105,6 +149,7 @@ Response `GetKeysetsResponse` of `Bob`:
       "unit": <str>,
       "active": <bool>,
       "input_fee_ppk": <int|null>,
+      "final_expiry": <int|null>
```
asmogo, line 150: why are we calling this `final`?

02.md
```diff
@@ -14,10 +14,20 @@ A mint can have multiple keysets at the same time. For example, it could have on
...
 
+### Short ID
+
+The short keyset ID (`s_id`) is defined as the abbreviated representation of the full Keyset ID. The length of the `s_id` is eight bytes (i.e., the version byte and the first seven bytes of the hash: `id_bytes[:8]` or `id_hex[:16]`).
+
+The `s_id` is only relevant for wallets, when encoding and decoding `Tokens`. To save space, a `Token`, as defined in [NUT-00][00], **SHOULD** use the short keyset ID `s_id` in the `Proof`. The recommended exception to that is when the `Proofs` reference a keyset where the `s_id` is ambiguous on that mint; in that case, the sending wallet **SHOULD** instead use the full keyset ID. Receiving wallets **MUST** therefore be able to decode both `Tokens` with a short keyset ID, as well as those with a full keyset ID.
+
+When receiving a `Token` with a `s_id`, a wallet **MUST** resolve the abbreviated `s_id` to the full `id`. If the `s_id` is ambiguous (i.e., multiple keyset `id`s are resolvable on the same mint), the wallet **MUST** error. If the `Token` is successfully parsed, the wallet **SHOULD** save the full-length keyset `id` with proofs in its database.
```
callebtc:
```suggestion
When receiving a `Token` with a `s_id`, wallet **MUST** resolve the abbreviated `s_id` to the full `id` to parse the `Proofs` contained in it. If the `s_id` is ambiguous (i.e., multiple keyset `id`s are resolvable on the same mint), the wallet **MUST** error. 
```

02.md
```diff
@@ -14,10 +14,20 @@ A mint can have multiple keysets at the same time. For example, it could have on
...
 
 The keyset `id` is in each `Proof` so it can be used by wallets to identify which mint and keyset it was generated from. The keyset field `id` is also present in the `BlindedMessages` sent to the mint and `BlindSignatures` returned from the mint (see [NUT-00][00]).
 
+### Short ID
+
+The short keyset ID (`s_id`) is defined as the abbreviated representation of the full Keyset ID. The length of the `s_id` is eight bytes (i.e., the version byte and the first seven bytes of the hash: `id_bytes[:8]` or `id_hex[:16]`).
+
+The `s_id` is only relevant for wallets, when encoding and decoding `Tokens`. To save space, a `Token`, as defined in [NUT-00][00], **SHOULD** use the short keyset ID `s_id` in the `Proof`. The recommended exception to that is when the `Proofs` reference a keyset where the `s_id` is ambiguous on that mint; in that case, the sending wallet **SHOULD** instead use the full keyset ID. Receiving wallets **MUST** therefore be able to decode both `Tokens` with a short keyset ID, as well as those with a full keyset ID.
```
callebtc:
```suggestion
Wallets use `s_id` to save space when encoding and decoding `Tokens` as defined in [NUT-00][00]. The recommended exception to that is when the `s_id` is ambiguous; in that case, the sending wallet **SHOULD** instead use the full keyset ID. Receiving wallets **MUST** be able to decode both `Tokens` with a short keyset ID, as well as those with a full keyset ID.
```

02.md
```diff
@@ -14,10 +14,20 @@ A mint can have multiple keysets at the same time. For example, it could have on
...
+
+When receiving a `Token` with a `s_id`, a wallet **MUST** resolve the abbreviated `s_id` to the full `id`. If the `s_id` is ambiguous (i.e., multiple keyset `id`s are resolvable on the same mint), the wallet **MUST** error. If the `Token` is successfully parsed, the wallet **SHOULD** save the full-length keyset `id` with proofs in its database.
+
+The mint is unaware of the `s_id`. All API calls exposed by the mint use the full keyset ID.
```
asmogo:
```suggestion
The mint is unaware of the `s_id`. All API endpoints exposed by the mint **MUST** use the full keyset ID.
```

02.md
```diff
@@ -56,12 +66,38 @@ Notice that since transactions can spend inputs from different keysets, the sum
...
+> [!CRITICAL]
+> Wallet implementations should reject any attempt at importing new keysets which IDs
+> collide with any of the previously added keysets.
+
+#### Keyset ID
 
-Keyset IDs have a version byte (two hexadecimal characters). The currently used version byte is `00`.
+Keyset IDs have a version byte (two hexadecimal characters). The currently used version byte is `01`.
 
 The mint and the wallets of its users can derive a keyset ID from the keyset of the mint. The keyset ID is a lower-case hex string. To derive the keyset ID of a keyset, execute the following steps:
```
asmogo:
```suggestion
Everybody can derive a keyset ID from the keysets of a mint. The keyset ID is a lower-case hex string. To derive the keyset ID of a keyset, execute the following steps:
```

02.md
```diff
@@ -56,12 +66,38 @@ Notice that since transactions can spend inputs from different keysets, the sum
...
 
+```
+1 - sort public keys by their amount in ascending order
+2 - concatenate all public keys to one byte array
```
asmogo:
```suggestion
2 - concatenate all public keys to a single byte array
```

02.md
```diff
@@ -79,6 +115,14 @@ def derive_keyset_id(keys: Dict[int, PublicKey]) -> str:
     return "00" + hashlib.sha256(pubkeys_concat).hexdigest()[:14]
 ```
 
+### Keyset Final Expiry
+
+A unix epoch number for a future point in time that represents the final expiry of the keyset. After the keyset's final expiry, the Mint is no longer obliged to fulfill promises signed with the keys from that keyset.
+
+This effectively implies that the Mint can irrevocably remove all of the nullifiers (`Y` values/spent ecash) associated with the expired keyset.
+
+The final expiry can be `null` if the keyset has no final-expiry.
```
asmogo:
```suggestion
The final expiry **MAY** be `null` if the keyset has no final-expiry.
```

prusnak:
Just a technical note:

You are using hmac-sha512 and then using only 256 bits of the result. 

If hmac-sha256 is used, then you can use whole result and sha256 is MUCH more friendly for embedded devices such as hardware wallets or smartcards.

13.md
```diff
@@ -12,11 +12,199 @@ In this document, we describe the process that allows wallets to recover their e
 
 An ecash token, or a `Proof`, consists of a `secret` generated by the wallet, and a signature `C` generated by the wallet and the mint in collaboration. Here, we describe how wallets can deterministically generate the `secrets` and blinding factors `r` necessary to generate the signatures `C`.
 
-The wallet generates a `private_key` derived from a 12-word [BIP39](https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki) `mnemonic` seed phrase that the user stores in a secure place. The wallet uses the `private_key`, to derive deterministic values for the `secret` and the blinding factors `r` for every new ecash token that it generates.
+The wallet generates a `seed` derived from a 12-word [BIP39](https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki) `mnemonic` seed phrase that the user stores in a secure place. The wallet uses the `seed`, to derive deterministic values for the `secret` and the blinding factors `r` for every new ecash token that it generates.
```
asmogo, line 15:
```suggestion
The wallet **MUST** generate a `seed` derived from a 12 - 24 word [BIP39](https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki) `mnemonic` seed phrase that the user stores in a secure place. The wallet **MUST** use the `seed`, to derive deterministic values for the `secret` and the blinding factors `r` for every new ecash token that it generates.
```

13.md
```diff
@@ -12,11 +12,199 @@ In this document, we describe the process that allows wallets to recover their e
...
 
-In order to do this, the wallet keeps track of a `counter_k` for each `keyset_k` it uses. The index `k` indicates that the wallet needs to keep track of a separate counter for each keyset `k` it uses. Typically, the wallet will need to keep track of multiple keysets for every mint it interacts with. `counter_k` is used to generate a [BIP32](https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki) derivation path which can then be used to derive `secret` and `r`.
+In order to do this, the wallet keeps track of a `counter_k` for each `keyset_k` it uses. The index `k` indicates that the wallet needs to keep track of a separate counter for each keyset `k` it uses. Typically, the wallet will need to keep track of multiple keysets for every mint it interacts with.
```
asmogo:
```suggestion
In order to do this, the wallet keeps track of a `counter_k` for each `keyset_k` it uses. The index `k` indicates that the wallet **MUST** keep track of a separate counter for each keyset `k` it uses. The wallet **MUST** keep track of multiple keysets for every mint it interacts with.
```

13.md
```diff
@@ -12,11 +12,199 @@ In this document, we describe the process that allows wallets to recover their e
...
+
+1. `message = b"Cashu_KDF_HMAC_SHA512" || keyset_id_bytes || counter_k_bytes || derivation_type_byte`, where:
+   - `"Cashu_KDF_HMAC_SHA512"` is the domain separation or purpose string, encoded to bytes as UTF-8.
+   - `keyset_id_bytes` are the raw bytes of the keyset_id (hex decoded).
+   - `counter_k_bytes` is the counter encoded as an unsigned 64-bit integer in big-endian format.
+   - `derivation_type_byte` is exactly 1 byte specifying the type of derivation required:
+     - `0` for secrets
+     - `1` for blinded messages
+2. `hmac_digest = HMAC_SHA512(seed, message)`, where `HMAC_SHA512` is the [hash-based message authentication code](https://en.wikipedia.org/wiki/HMAC), using SHA-512 as the hashing algorithm.
+3. `secret = hmac_digest[:32]` or `blinding_factor = hmac_digest[:32]`.
```
asmogo:
```suggestion
2. `hmac_digest_secret = HMAC_SHA512(seed, secret_message)` and `hmac_digest_blinding_factor = HMAC_SHA512(seed, blinding_factor_message)`, where `HMAC_SHA512` is the [hash-based message authentication code](https://en.wikipedia.org/wiki/HMAC), using SHA-512 as the hashing algorithm.
3. `secret = hmac_digest_secret[:32]` or `blinding_factor = hmac_digest_blinding_factor[:32]`.
```

02.md
```diff
@@ -56,12 +66,38 @@ Notice that since transactions can spend inputs from different keysets, the sum
...
+```
+
+An example implementation in Python:
+
+```python
+def derive_keyset_id(keys: Dict[int, PublicKey]) -> str:
+    sorted_keys = dict(sorted(keys.items()))
+    keyset_id_bytes = b"".join([p.serialize() for p in sorted_keys.values()])
+    keyset_id_bytes += b"unit:sat"
+    keyset_id_bytes += b"final_expiry:"+str(1896187313).encode("utf-8")
```
prusnak: can we add delimiters here?

they won't change the size of the result and make it so much better for debugging implementations

example

```suggestion
    sorted_keys = dict(sorted(keys.items()))
    keyset_id_bytes = b",".join([p.serialize() for p in sorted_keys.values()])
    keyset_id_bytes += b"|unit:sat"
    keyset_id_bytes += b"|final_expiry:"+str(1896187313).encode("utf-8")
```

13.md
```diff
@@ -12,11 +12,199 @@ In this document, we describe the process that allows wallets to recover their e
...
+
+The HMAC-SHA512 KDF is built as the following:
+
+1. `message = b"Cashu_KDF_HMAC_SHA512" || keyset_id_bytes || counter_k_bytes || derivation_type_byte`, where:
+   - `"Cashu_KDF_HMAC_SHA512"` is the domain separation or purpose string, encoded to bytes as UTF-8.
+   - `keyset_id_bytes` are the raw bytes of the keyset_id (hex decoded).
+   - `counter_k_bytes` is the counter encoded as an unsigned 64-bit integer in big-endian format.
+   - `derivation_type_byte` is exactly 1 byte specifying the type of derivation required:
+     - `0` for secrets
+     - `1` for blinded messages
```
prusnak: let's make it obvious this is a binary zero or one and not ascii zero or one:

```suggestion
     - `0x00` for secrets
     - `0x01` for blinded messages
```

asmogo (commented):
Suggesting minor changes. 
Updated KDF with hmac_digest_secret and hmac_digest_blinding_factor.

02.md
a1denvalu3: When I first wrote this what I imagined could happen is at some point we'll want other expiry fields (for example `active_expiry` to signal when this keyset is expected to be made inactive). `final_expiry` is stronger and signals a potential deletion from the DB.

13.md
davidcaseria: I think this suggestion was nacked in previous conversations.

13.md
```diff
@@ -12,11 +12,199 @@ In this document, we describe the process that allows wallets to recover their e
...
+This versioning approach eliminates counter-based compatibility issues and ensures wallets can correctly derive secrets for any keyset by inspecting the keyset ID version.
+
+### HMAC-SHA256 Derivation
+
+For keysets with version `01`, the wallet uses `counter_k`, `seed` and `keyset_id` as inputs to a Key Derivation Function (KDF) which output is used to derive `secret` and `r`.
+
+The HMAC-SHA256 KDF is built as the following:
+
+1. `message = b"Cashu_KDF_HMAC_SHA512" || keyset_id_bytes || counter_k_bytes || derivation_type_byte`, where:
+   - `"Cashu_KDF_HMAC_SHA512"` is the domain separation or purpose string, encoded to bytes as UTF-8.
```
prusnak:
```suggestion
1. `message = b"Cashu_KDF_HMAC_SHA256" || keyset_id_bytes || counter_k_bytes || derivation_type_byte`, where:
   - `"Cashu_KDF_HMAC_SHA256"` is the domain separation or purpose string, encoded to bytes as UTF-8.
```

13.md
```diff
@@ -12,11 +12,199 @@ In this document, we describe the process that allows wallets to recover their e
...
+        return derive_secret_and_r_hmac(seed, keyset_id, counter_k)
+    else:
+        raise ValueError(f"Unsupported keyset version: {keyset_version}")
+
+def derive_secret_and_r_hmac(seed: bytes, keyset_id: str, counter_k: int):
+    """
+    HMAC-SHA256 derivation for keyset version 01
+    """
+    # Step 1: Create the message
+    message = b"Cashu_KDF_HMAC_SHA512" + bytes.fromhex(keyset_id) + counter_k.to_bytes(8, 'big')
```
prusnak:
```suggestion
    message = b"Cashu_KDF_HMAC_SHA256" + bytes.fromhex(keyset_id) + counter_k.to_bytes(8, 'big')
```

13.md
```diff
@@ -12,11 +12,199 @@ In this document, we describe the process that allows wallets to recover their e
...
+function deriveSecretAndRHmac(
+  seed: Buffer,
+  keysetId: string,
+  counterK: number,
+) {
+  // Create message: domain separator + keyset_id + counter
+  const counterBuffer = Buffer.alloc(8);
+  counterBuffer.writeBigUInt64BE(BigInt(counterK));
+  const message = Buffer.concat([
+    Buffer.from("Cashu_KDF_HMAC_SHA512"),
```
prusnak:
```suggestion
    Buffer.from("Cashu_KDF_HMAC_SHA256"),
```

prusnak (dismissed):
We need to change SHA512 -> SHA256 in domain separation purpose strings.

lollerfirst:
@davidcaseria https://github.com/davidcaseria/nuts/pull/8

02.md
```diff
@@ -56,11 +66,37 @@ Notice that since transactions can spend inputs from different keysets, the sum
...
+```
+
+An example implementation in Python:
+
+```python
+def derive_keyset_id(keys: Dict[int, PublicKey]) -> str:
+    sorted_keys = dict(sorted(keys.items()))
+    keyset_id_bytes = b",".join([p.serialize() for p in sorted_keys.values()])
+    keyset_id_bytes += b"|unit:sat"
+    keyset_id_bytes += b"|final_expiry:"+str(1896187313).encode("utf-8")
```
callebtc:
```suggestion
    if final_expiry:
        keyset_id_bytes += b"|final_expiry:"+str(1896187313).encode("utf-8")
```

this should be conditional on whether `final_expiry` is set.

callebtc: Commit #b25abaa changed the keyset ID calculation to include a separator but test vectors and implementations didn't. Needs to be resolved.

prusnak:
I prefer option 2 - it is so much cleaner.

If this has been already implemented in CDK or other implementations, I can live with 1. But we should know better and not push discussed drafts into production before they are accepted.

13.md
```diff
@@ -12,11 +12,199 @@ In this document, we describe the process that allows wallets to recover their e
...
+  counterK: number,
+) {
+  // Legacy BIP32 derivation for version 00
+  // Implementation would use BIP32 library (e.g., bip32, bitcoinjs-lib)
+  // Calculate keyset_id_int for BIP32 derivation path
+  const keysetIdInt = BigInt(`0x${keysetId}`) % BigInt(2 ** 31 - 1);
+
+  // Derive using BIP32 paths:
+  // secret_path = `m/129372'/0'/{keysetIdInt}'/{counterK}'/0`
+  // r_path = `m/129372'/0'/{keysetIdInt}'/{counterK}'/1`
```
callebtc: redundant comments?

SatsAndSports:
Could we add an `active_at_least_until` field in active keysets?

As the SIG_ALL message change (https://github.com/cashubtc/nuts/pull/302) commits the outputs to a particular keyset, and holders of (partially-) signed transactions might want to be able to hold the transactions for a longer time, they would like some confidence that an `active` keyset in their output will remain `active`

One example of this is the Cashu Channel, where - like in Lightning - you want users to keep their transactions away from mint until their are ready to close the channel.

thesimplekid:
ACK 9f5a7471523f04e2deab8d3f0a535dc60e3f26cd

13.md
```diff
@@ -104,9 +325,22 @@ If the wallet used the restore endpoint [NUT-09][09] for regenerating the `Proof
...
+- Repeat until three consecutive batches are returned empty
+- Reset `counter` to the value at the last successful restore + 1
+
+For `01` keysets:
+
+- Set `counter = 0`
+- Restore `Proofs` in batches of 100 using the legacy derivation, and increment `counter`
+- Repeat until three consecutive batches are returned empty
+- Reset `counter` to the value at the last successful restore + 1
+- Restore `Proofs` in batches of 100 using the HMAC-SHA256 derivation, and increment `counter`
```
callebtc: Seems unnecessary to try both methods, assuming that we transition to the new KDF only for V2 keysets

asmogo (approved):

thesimplekid (approved):
ACK 97b957304c2bd2c23eb2c03c278c10f19ce11693

02.md
```diff
@@ -56,11 +56,36 @@ Notice that since transactions can spend inputs from different keysets, the sum
...
-The mint and the wallets of its users can derive a keyset ID from the keyset of the mint. The keyset ID is a lower-case hex string. To derive the keyset ID of a keyset, execute the following steps:
+Keyset IDs are derived from public data. To derive the keyset ID of a keyset, execute the following steps:
+
+```
+1 - sort public keys by their amount in ascending order
+2 - concatenate all public keys to a single byte array
+3 - add the lowercase UTF8-encoded unit string to the byte array (e.g. "unit:sat")
+4 - If a final expiration is specified, add the UTF8-encoded string (e.g "final_expiry:1896187313")
+4 - HASH_SHA256 the concatenated byte array
+5 - prefix it with a keyset ID version byte "01"
```
robwoodgate: The keyset ID derivation doesn't protect/guarantee the amounts have not been tampered with, only that the amounts are still ordered the same.

What stops a mint from changing the amounts bound to each key to devalue and "inflate" away their liabilities?

eg: 1, 2, 4, 8, 16, 32, 64

could become: 1, 2, 3, 4, 8, 16, 32
or maybe even: 1, 1, 1, 1, 1, 1, 1

Perhaps we should concat all the amounts in order too... or maybe just adding the sum of key amounts would be enough to reduce the risk of serious damage?

callebtc: This is a really good point. It would change the derivation algorithm and all test vectors etc.

prusnak: Not sure what is the state of implementation and mainly usage of Keyset ID v2, but I think it is worth the change if the usage is non-existent or minimal.

If we will be changing this, I would suggest to include my suggestion which was dismissed earlier: https://github.com/cashubtc/nuts/pull/182#discussion_r2306837257

robwoodgate:
Cross referencing the NUT-00 conventions PR, which currently warns against doing MOD N. https://github.com/cashubtc/nuts/pull/322

Will need to update this if we proceed with the the proposed blinding scheme here.

02.md
a1denvalu3: @callebtc @prusnak @robwoodgate https://github.com/a1denvalu3/nutshell/pull/6/changes

I have a sub-branch of my Nutshell IDv2 implementation with the modified derivation scheme and adjusted test vectors **IF** we want to go ahead with this.

SatsAndSports: And maybe include the feerate `input_fee_ppk` in the id too?

In the mint changes the feerate, that might break some pre-signed transactions, and therefore I'd like to discourage changing of feerates.

If a pre-signed swap spends from a keyset, and the mint changes the `input_fee_ppk` of that keyset before the swap is executed, the swap might fail because it won't have enough inputs to cover the outputs

If the mint wants to change the feerate, they could de-activate the keyset and create a new keyset with the new feerate

02.md
```diff
@@ -105,6 +142,7 @@ Response `GetKeysetsResponse` of `Bob`:
       "unit": <str>,
       "active": <bool>,
       "input_fee_ppk": <int|null>,
+      "final_expiry": <int|null>
```
SatsAndSports, line 150: This implies that, if a keyset has no final expiry, then the response must include `final_expiry: null`. But I guess we want to make that optional? i.e. wallets won't expect the `final_expiry` field to be present

Can we make that explicit somewhere here? i.e. The mint may omit `input_fee_ppk` if the fee is 0, or ~maybe~ may omit the `final_expiry` if there is no expiry

02.md
```diff
@@ -119,22 +157,25 @@ Here, `id` is the keyset ID, `unit` is the unit string (e.g. "sat") of the keyse
 {
   "keysets": [
     {
-      "id": "009a1f293253e41e",
```
SatsAndSports, line 122: How about including one v1 keyset in this example response, if only to remind wallet authors that mints are likely to include both v1 and v2 keysets?

02.md
```diff
@@ -56,11 +56,36 @@ Notice that since transactions can spend inputs from different keysets, the sum
...
+#### Keyset ID V2
 
-Keyset IDs have a version byte (two hexadecimal characters). The currently used version byte is `00`.
+Keyset IDs are 33 byte hex strings with a version byte (two hexadecimal characters). The currently used version byte is `01`.
 
-The mint and the wallets of its users can derive a keyset ID from the keyset of the mint. The keyset ID is a lower-case hex string. To derive the keyset ID of a keyset, execute the following steps:
+Keyset IDs are derived from public data. To derive the keyset ID of a keyset, execute the following steps:
+
+```
+1 - sort public keys by their amount in ascending order
```
SatsAndSports: A bit pedantic, but as the amounts from `/v1/keys` are strings, not numbers, we should be clear about the order:

```suggestion
1 - sort public keys by their amount in ascending order. As the amount in the mint's /v1/keys response is a string, e.g. `"32"`, we sort by increasing numerical - not lexicographical - order
```

02.md
callebtc: > f we will be changing this, I would suggest to include my suggestion which was dismissed earlier: [#182 (comment)](https://github.com/cashubtc/nuts/pull/182#discussion_r2306837257)

yes! we need to add separators if we append the amounts as well, let's combine both changes

02.md
callebtc: that's how we usually indicate that a value is optional through the spec.

is it an equivalent "json" to omit a key if the value is `null`? i.e.`{a: 1, b: null} == {a: 1} `?

02.md
callebtc: I think it's fine, since we're deprecating v1 with this.

callebtc:
I think all points should be addressed now.

robwoodgate:
@callebtc - looking great. Just one comment - should `verify_02.py` be placed under tools instead of in the root folder?

02.md
thesimplekid: > And maybe include the feerate input_fee_ppk in the id too?

> In the mint changes the feerate, that might break some pre-signed transactions, and therefore I'd like to discourage changing of feerates.

This would be a deviation from current spec as mints currently are allowed to change fee of a past keysets. Though I'm not sure if nutshell allows it, cdk doesn't it rotates to a new keyset as you suggested but that was an implementation choice not a spec requirement. 

I think there are arguments for and against allowing mints to change keyset fees that are unrelated to the id.

robwoodgate: > > And maybe include the feerate input_fee_ppk in the id too?
> 
> > In the mint changes the feerate, that might break some pre-signed transactions, and therefore I'd like to discourage changing of feerates.
> I think there are arguments for and against allowing mints to change keyset fees that are unrelated to the id.

FYI. Calle's latest iteration has fee baked in.

lescuer97:
should we add a couple test vectors for the preimage?

02.md
```diff
@@ -56,11 +56,41 @@ Notice that since transactions can spend inputs from different keysets, the sum
...
+7 - prefix it with a keyset ID version byte "01"
+```
+
+An example implementation in Python:
+
+```python
+def derive_keyset_id_v2(keys: Dict[int, PublicKey], unit: str, input_fee_ppk: Optional[int], final_expiry: Optional[int]) -> str:
+    sorted_keys = dict(sorted(keys.items()))
+    keyset_id_bytes = b",".join(
+        [f"{k}:{v.serialize()}".encode("utf-8") for k, v in sorted_keys.items()]
```
prusnak, line 81: Looking at this code it occurred to me, that this works fine only if there are no duplicate entries in key values of `keys`. And this seems to be the case since `keys` is expected to be a dict. (otherwise the `keys` would need to be a list of tuples instead of a dict).

However, does the keyset spec disallow multiple public keys with the same denomination? I don't think we should allow this, but in that case, we should add this limitation to the keyset spec. Or is such limitation already in the keyset spec?

02.md
```diff
@@ -56,11 +56,41 @@ Notice that since transactions can spend inputs from different keysets, the sum
...
+    )
+    keyset_id_bytes += f"|unit:{unit}".encode("utf-8")
+    if input_fee_ppk is not None and input_fee_ppk != 0:
+        keyset_id_bytes += f"|input_fee_ppk:{input_fee_ppk}".encode("utf-8")
```
thesimplekid, line 85: I think it's worth noting that including the fee in the ID now means that mints cannot change the fee of past keysets. This to some extent reduces the mint's ability to dynamically change fees in response to a DoS, for example, as the first operation will still have the lower fee, only the second operation will incur the higher fee of the new keyset. I am not necessarily against this change; I just think we should be aware of it and make sure we are making this tradeoff intentionally.

callebtc (approved):

02.md
```diff
@@ -56,11 +56,41 @@ Notice that since transactions can spend inputs from different keysets, the sum
...
 
-Keyset IDs have a version byte (two hexadecimal characters). The currently used version byte is `00`.
+Keyset IDs are 33 byte hex strings with a version byte (two hexadecimal characters). The currently used version byte is `01`.
 
-The mint and the wallets of its users can derive a keyset ID from the keyset of the mint. The keyset ID is a lower-case hex string. To derive the keyset ID of a keyset, execute the following steps:
+Keyset IDs are derived from public data. To derive the keyset ID of a keyset, execute the following steps:
+
+```
+1 - sort public keys by their amount in ascending numerical order
+2 - concatenate each amount and its corresponding public key hex string (as "amount:publickey_hex") to a single byte array, separating each pair with a comma (",")
```
asmogo: We should explicitly mention that these keys must be lowercase because SHA-256 is case-sensitive, while hex is not. 
```suggestion
2 - concatenate each amount and its corresponding lowercase public key hex string (as "amount:publickey_hex") to a single byte array, separating each pair with a comma (",")
```

PR #184, opened 2024-11-01
Add NUT-XX: Describe a primes-based keyset, proof scheme

ok300:
Draft describing a primes-based keyset and proof scheme to achieve smaller tokens.

For easier viewing: https://github.com/ok300/nuts/blob/ok300-keyset-token-primes/xx.md

### DIFF
--- a/xx.md
+++ b/xx.md
@@ -100,13 +100,11 @@ function find_proof_amounts(keyset_amounts[], token_amount):
         return [token_amount]
     
     proof_amounts = []
+    target = token_amount
     
-    target =
-        if token_amount % 2 == 0
-            token_amount
-        else
-            append 1 to proof_amounts[]
-            token_amount - 1
+    if token_amount % 2 == 1
+        append 1 to proof_amounts[]
+        target = target - 1
         
     for p1 in keyset_amounts
         if p1 > target / 2
--- a/xx.md
+++ b/xx.md
@@ -15,7 +15,7 @@ Goldbach's conjecture states that every even natural number greater than 2 is th
 Based on this, a token can be decomposed in proofs as follows:
 
 | Token amount |            Proof amounts |
-|-------------:|-------------------------:|
+| -----------: | -----------------------: |
 |            1 |                        1 |
 |          `p` |                      `p` |
 |         `2n` |                `p1 + p2` |
@@ -27,7 +27,6 @@ In other words, any amount can be built from at most 3 proofs, but sometimes fro
 
 To allow wallets to use this scheme, a keyset has to include signatures for `1` and primes up to `P`.
 
-
 ### Client Considerations
 
 There are multiple possible decompositions[^2] for each token amount. When building a token, the wallet only has to find one.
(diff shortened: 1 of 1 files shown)


prusnak:
Isn't this terrible for privacy?

ok300:
Good question.

Intuitively, I think not because except for `1` and the prime token amounts, all amounts have multiple ways to be represented as a sum of keyset amounts. This means the mint cannot assume the decomposition, because different clients may find different ones for the same token amount.

I ran some tests to see what's the distribution of the keyset amounts used to represent tokens between 1 and 1000. Sorted by occurrences, it's this:

<details>
<summary>Click to show table</summary>

| keyset amount | occurrences |
|--------------:|------------:|
|            1	 |         333 |
|            3	 |         294 |
|            5	 |         206 |
|            7	 |         161 |
|           11	 |          68 |
|           13	 |          48 |
|           19	 |          41 |
|           31	 |          16 |
|           919 |         	15 |
|           ... (2x) |         	14 |
|           17	 |          13 |
|           ... (7x) |         	11 |
|           ... (11x) |         	10 |
|           ... (5x)	 |           9 |
|           .... (13x)	 |           8 |
|           ... (7x)	 |           7 |
|           ... (58x)	 |           6 |
|           ... (13x)	 |           5 |
|           ... (14x)	 |           4 |
|            ... (28x)	 |           3 |
|           997 |          	2 |

</details>

No keyset amount appears only once. The biggest one (`997`) appears twice, then 28 of them appear 3 times each, 14 of them appear 4 times, and so on.

For `997` appearing twice, it means
- if Alice mints 999 (using 997, 2)
- if Susan mints 997 (using only 997)
- if Carol swaps or melts 997, the mint can't know who she got it from

The more users the mint has, the harder this becomes. As part of Carol's swap or melt, fees are paid so 997 becomes 996 or lower and has to be further decomposed.

It gets exponentially harder for other amounts, which are used more often.

callebtc:
> Isn't this terrible for privacy?

Yes, it seems. The whole idea with keysets for different amounts is to *reduce* the number of keys so we can gain privacy. We're purposefully trade efficiency vs privacy in favor of privacy with the keysets that we use. Here, 1000 keys can't even represent 10000 sats. With powers of two, we can represent the entire bitcoin supply with just 54 keys.

@ok300 appreciate the effort but I don't see why this should be a NUT. A mint can choose whatever keyset they like, nothing stops them from using (way too many!) prime numbers for decomposing the amounts. Then, it becomes a wallet-side issue to decompose these amounts efficiently. I don't see why anything should have to be specified in the spec for that, everyone can *just do it* without breaking compatibility (although, I would strongly advice against it because of the privacy implications).

Another note, on the format. Although I think the tables and the benchmarks are *very* interesting, I think they should rather be written up in an article or gist, but they are not fit for the specs either. The specs have the sole purpose of establishing consensus.

callebtc:
nACK

prusnak:
NACK, the proposal decreases privacy significantly

PR #185, opened 2024-11-01
Add nix flake for prettier checks

thesimplekid:
just for you @callebtc


callebtc:
why do we need this unsolicited nix config flash again?

callebtc:
visits spec repo, sees this:
image

thesimplekid:
> why do we need this unsolicited nix config flash again?

This would be useful to me as I don't have prettier installed globally to run the linting check, and it would be easier for me (and others on nix) to have a dev flake defined in the repo to drop into then creating a custom dev shell each time. However, I do realize this is likely an issue for only me so feel free to close. 

Changing the CI only has the benefit since there is a flake defined it ensures what is run locally is run in CI, though in such a simply CI likely no real benefit.

prusnak:
Closing then.

PR #187, opened 2024-11-07
Add NUT-XX: Token non-adjacent form (NAF)

ok300:
This PR proposes a way to build a more compact token. For a table comparing it to the common token representation (powers of two), see https://gist.github.com/ok300/c94caa4b21807307bb21ab0dade38508 .

For easier reading: https://github.com/ok300/nuts/blob/ok300-token-naf/xx.md

---

I'm not sure if this belongs as a new NUT, or as an addendum to NUT-02 "Keysets and fees", or as something else entirely. Arguments I considered:
- why it should be a NUT: because a mint that supports it will indicate it in its Mint Info response, which lets wallets know it supports more compact tokens without requiring wallets to fetch its keysets
- why it should be an addendum to NUT-02: to make mints and wallets aware of this more compact way of using keysets, since NUT-02 is focused on keysets

Please let me know if there is a better place for this, or if you find any issue with the reasoning in it.

### DIFF
--- a/xx.md
+++ b/xx.md
@@ -1,0 +1,35 @@
+# NUT-XX: Token non-adjacent form (NAF)
+
+`optional`
+
+---
+
+This NUT proposes a more compact scheme for representing token amounts using powers of two.
+
+---
+
+Every natural number has a unique binary representation, which means it can be written as a sum of distinct powers of two. This is a common way of choosing proof amounts when constructing a token.
+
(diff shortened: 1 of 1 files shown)


prusnak:
I don't get this proposal.

What is stopping me from swapping 4 sats for [4, -4, 4] and then swapping [4, 4] for [8] and discarding [-4]?

ok300:
Didn't think of that, good catch. Yes, bad idea.

PR #188, opened 2024-11-10
NUT-20: signature on mint request

thesimplekid:
This NUT defines a protocol extension that enables signature-based authentication for mint quote redemption. When requesting a mint quote, clients can provide a public key. The mint will then require a valid signature from the corresponding secret key before processing the mint.

This is defined as an optional extension to NUT-04 and the intention is to reuse it and make it mandatory for bolt12 and onchain. 

- [x] CDK https://github.com/cashubtc/cdk/pull/446
- [x] Nutshell https://github.com/cashubtc/nutshell/pull/670
- [ ] Nutmix
- [ ] goNuts
- [ ] cashu-ts


19.md
```diff
@@ -0,0 +1,154 @@
...
+    }
+  ]
+}
+```
+
+If the invoice was not paid yet, `Bob` responds with an error. In that case, `Alice` **CAN** repeat the same request until the Lightning invoice is settled, as in NUT04. If `Alice` does not include a witness on the `PostMintBolt11Request` but did include a `pubkey` in the `PostMintBolt11QuoteRequest` the `Bob` **MUST** respond with an error, `Alice` **CAN** repeat the request with a witness in order to mint the ecash.
+
+## Settings
+
+The settings for this NUT indicate the support for requiring a signature before minting. They are part of the info response of the mint ([NUT-06][06]) which in this case reads
```
lollerfirst, line 146:
```suggestion
The settings for this NUT indicate the support for requiring a signature before minting. They are part of the info response of the mint ([NUT-06](06.md)) which in this case reads
```

prusnak: Another way how to fix the issue is to add the following at the end of the file:

```
[06]: 06.md
```

This seems to be preferred way used in other documents.

lollerfirst:
@thesimplekid Should we make it clear that wallets should use an ephemeral (one time) key?

error_codes.md
```diff
@@ -16,6 +16,7 @@
 | 20005 | Quote is pending                                | [NUT-04][04], [NUT-05][05]               |
 | 20006 | Invoice already paid                            | [NUT-05][05]                             |
 | 20007 | Quote is expired                                | [NUT-04][04], [NUT-05][05]               |
+| 20008 | Witness not provided for mint quote             | [NUT-19][19]                             |
```
lollerfirst, line 19: We need an error for the invalid witness as well.

thesimplekid:
> @thesimplekid Should we make it clear that wallets should use an ephemeral (one time) key?

Added note on this

19.md
```diff
@@ -0,0 +1,180 @@
...
+    {
+      "id": "009a1f293253e41e",
+      "amount": 8,
+      "C_": "0277d1de806ed177007e5b94a8139343b6382e472c752a74e99949d511f7194f6c"
+    }
+  ]
+}
+```
+
+If the invoice was not paid yet, `Bob` responds with an error. In that case, `Alice` **CAN** repeat the same request until the Lightning invoice is settled, as in NUT04. If `Alice` does not include a witness on the `PostMintBolt11Request` but did include a `pubkey` in the `PostMintBolt11QuoteRequest` the `Bob` **MUST** respond with an error, `Alice` **CAN** repeat the request with a witness in order to mint the ecash.
```
davidcaseria, line 157: nit: I think it's best to stick to the [RFC 2219](https://datatracker.ietf.org/doc/html/rfc2119) keywords and definitions (i.e., **CAN** -> **MAY** or in this case **SHOULD**?).

We use **CAN** (or even can) many times elsewhere so this is by no means a blocking comment, but it's something I noticed in the spec that we **SHOULD** ( :wink: ) cleanup.

thesimplekid: Good catch I thought CAN was a keyword.

19.md
```diff
@@ -0,0 +1,180 @@
...
+
+```
+msg =  quote_id || B_0 || ... || B_n
+```
+
+Here, `||` denotes string concatenation. The `B_` of each output is **a hex string**.
+
+#### Signature scheme
+
+To mint a quote where a public key was provided, the minter needs to include signatures in the `PostMintBolt11Request`. We use `libsecp256k1`'s serialized 64-byte Schnorr signatures on the SHA256 hash of the message to sign as defined above.
```
lollerfirst, line 82:
```suggestion
To mint a quote where a public key was provided, the minter needs to include signatures in the `PostMintBolt11Request`. We use a [BIP340](https://github.com/bitcoin/bips/blob/master/bip-0340.mediawiki) signature on the SHA-256 hash of the message to sign as defined above.
```
Also why hash it again if everything that goes into the schnorr signature is hashed anyway?

19.md
```diff
@@ -0,0 +1,180 @@
...
+```
+
+#### Message aggregation
+
+To provide a signature for a mint request, the owner of one of the signing public keys must concatenate the `PostMintQuoteBolt11Response.quote` and the `B_` fields of all `BlindedMessages` (outputs, see [NUT-00][00]) to a single message string in the order they appear in the `PostMintRequest`. This string concatenated is then hashed and signed (see [Signature scheme](#signature-scheme)).
+
+If a request has `n` outputs the message to sign becomes:
+
+```
+msg =  quote_id || B_0 || ... || B_n
```
lollerfirst, line 75: me and calle were discussing whether it should be better to just JSON-serialize the entire payload and then use the utf-8 encoding of the `quote_id` and the entire serialized payload.

I'm fine with either approach what do you think?

lollerfirst: For reference how I am doing it right now in my Nutshell branch (before having seen this):

```python
sigbytes = bytes.fromhex(signature)
serialized_outputs = b"".join([o.json().encode("utf-8") for o in outputs])
msgbytes = quote.quote.encode("utf-8") + serialized_outputs
```

19.md
thesimplekid: > Also why hash it again if everything that goes into the schnorr signature is hashed anyway?

I used the wording from NUT-11. Its only hashed once maybe that wording is confusing and should be removed?

lollerfirst: So yea I've also seen this other times in the NUTs. I personally think it's not needed but for the sake of consistency let's leave it like this. (Also we cannot change NUT-11 without breaking a lot of implementations)

thesimplekid: Yeah I think its best this and NUT-11 match.

19.md
```diff
@@ -0,0 +1,180 @@
...
+> **Privacy:** To prevent linking multiple mint quotes together, wallets **SHOULD** generate a unique public key for each mint quote request
+
+The mint `Bob` then responds with a `PostMintQuoteBolt11Response`:
+
+```json
+{
+  "quote": <str>,
+  "request": <str>,
+  "state": <str_enum[STATE]>,
+  "expiry": <int>
```
callebtc, line 38: We should probably return the `pubkey` here as well (so that it's clear that the quote is locked)

19.md
```diff
@@ -0,0 +1,180 @@
...
+```
+
+Response of `Bob`:
+
+```json
+{
+  "quote": "9d745270-1405-46de-b5c5-e2762b4f5e00",
+  "request": "lnbc100n1pj4apw9...",
+  "state": "UNPAID",
+  "expiry": 1701704757
```
callebtc, line 64: We would also return `pubkey` here if we did above.

19.md
callebtc: > calle and I were discussing whether it should be better to just JSON-serialize the entire payload and then use the utf-8 encoding of the `quote_id` and the entire serialized payload.

I think that was a misunderstanding. I was wondering if we should serialize the outputs as hex to bytes or as utf-8. I believe it would be more coherent to use utf-8 here so we can use the same encoding as for the quote ID. That's what I thought we refer to when we said "serialize as json".

i.e.

```python
serialized_outputs = b"".join([o.B_.encode("utf-8") for o in outputs])
msgbytes = quote.quote.encode("utf-8") + serialized_outputs
```

error_codes.md
```diff
@@ -16,6 +16,8 @@
 | 20005 | Quote is pending                                | [NUT-04][04], [NUT-05][05]               |
 | 20006 | Invoice already paid                            | [NUT-05][05]                             |
 | 20007 | Quote is expired                                | [NUT-04][04], [NUT-05][05]               |
+| 20008 | Invalid witness                                 | [NUT-19][19]                             |
```
callebtc, line 19: If we want to make a `required` flag for the mint info, we would need an error code for the `Quote without pubkey` case

callebtc: Maybe "invalid witness" and "no witness provided" can be the same error.

19.md
thesimplekid: Is `o.B_` here the hex encoding?

thesimplekid:
```
    /// Constructs the message to be signed according to NUT-19 specification.
    ///
    /// The message is constructed by concatenating (as UTF-8 encoded bytes):
    /// 1. The quote ID (as UTF-8)
    /// 2. All blinded secrets (B_0 through B_n) converted to hex strings (as UTF-8)
    ///
    /// Format: `quote_id || B_0 || B_1 || ... || B_n`
    /// where each component is encoded as UTF-8 bytes
    pub fn msg_to_sign(&self) -> Vec<u8> {
        // Pre-calculate capacity to avoid reallocations
        let capacity = self.quote.len() + (self.outputs.len() * 66);
        let mut msg = Vec::with_capacity(capacity);
        msg.append(&mut self.quote.clone().into_bytes()); // String.into_bytes() produces UTF-8
        for output in &self.outputs {
            // to_hex() creates a hex string, into_bytes() converts it to UTF-8 bytes
            msg.append(&mut output.blinded_secret.to_hex().into_bytes());
        }
        msg
    }
```

thesimplekid: f866a2511c160870c0e250df176c595001b6d481

19.md
```diff
@@ -0,0 +1,182 @@
...
+
+In addition to signaling optional support, the mint can require a public key for mint quotes and a corresponding signature during minting. If `required` is set to `true` in the NUT-19 settings:
+
+1. The wallet **MUST** provide a public key when requesting a mint quote
+2. The wallet **MUST** provide a valid witness signature when submitting the mint request
+
+> **Note:** Enabling this setting makes the mint incompatible with wallets that do not support NUT-19
+
+[00]: 00.md
+[06]: 06.md
```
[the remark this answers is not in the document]
thesimplekid: I think we can remove the required flag here. Leaving it up to wallets to choose to use it or not for NUT04 and future nuts that want to require it (eg bolt12) can specify that there.

callebtc (approved):
LGTM

thesimplekid:
ACK cccbb1dd068791557b4a1ccc6e5091a901056a09

error_codes.md
```diff
@@ -1,21 +1,23 @@
...
+| 12001 | Keyset is not known                               | [NUT-02][02], [NUT-04][04]               |
+| 12002 | Keyset is inactive, cannot sign messages          | [NUT-02][02], [NUT-03][03], [NUT-04][04] |
+| 20001 | Quote request is not paid                         | [NUT-04][04]                             |
+| 20002 | Tokens have already been issued for quote         | [NUT-04][04]                             |
+| 20003 | Minting is disabled                               | [NUT-04][04]                             |
+| 20005 | Quote is pending                                  | [NUT-04][04], [NUT-05][05]               |
+| 20006 | Invoice already paid                              | [NUT-05][05]                             |
+| 20007 | Quote is expired                                  | [NUT-04][04], [NUT-05][05]               |
+| 20008 | Signature on mint request not provided or invalid | [NUT-19][19]                             |
+| 20009 | Pubkey required on mint quote                     | [NUT-19][19]                             |
```
thesimplekid, line 20:
```suggestion
| 20008 | Signature on mint request not provided or invalid | [NUT-20][20]                             |
```

error_codes.md
```diff
@@ -30,3 +32,4 @@
 [10]: 10.md
 [11]: 11.md
 [12]: 12.md
+[19]: 19.md
```
thesimplekid, line 35:
```suggestion
[20]: 20.md
```

error_codes.md
```diff
@@ -1,21 +1,22 @@
...
+| 11006 | Amount outside of limit range                     | [NUT-04][04], [NUT-05][05]               |
+| 12001 | Keyset is not known                               | [NUT-02][02], [NUT-04][04]               |
+| 12002 | Keyset is inactive, cannot sign messages          | [NUT-02][02], [NUT-03][03], [NUT-04][04] |
+| 20001 | Quote request is not paid                         | [NUT-04][04]                             |
+| 20002 | Tokens have already been issued for quote         | [NUT-04][04]                             |
+| 20003 | Minting is disabled                               | [NUT-04][04]                             |
+| 20005 | Quote is pending                                  | [NUT-04][04], [NUT-05][05]               |
+| 20006 | Invoice already paid                              | [NUT-05][05]                             |
+| 20007 | Quote is expired                                  | [NUT-04][04], [NUT-05][05]               |
+| 20008 | Signature on mint request not provided or invalid | [NUT-20][20]                             |
```
thesimplekid, line 19:
```suggestion
| 20008 | Signature on mint request not provided or invalid | [NUT-20][20]                             |
| 20009 | Pubkey required on mint quote                              | [NUT-20][20]                             |
```

error_codes.md
```diff
@@ -16,6 +16,8 @@
 | 20005 | Quote is pending                                | [NUT-04][04], [NUT-05][05]               |
 | 20006 | Invoice already paid                            | [NUT-05][05]                             |
 | 20007 | Quote is expired                                | [NUT-04][04], [NUT-05][05]               |
+| 20008 | Signature for mint request invalid              | [NUT-20][20]                             |
+| 20009 | Pubkey required for mint quote                  | [NUT-20][20]                             |
```
lollerfirst, line 20: Is this error still relevant?

thesimplekid: Will be used for BOLT12 and onchain

04.md
```diff
@@ -24,7 +26,7 @@ The wallet of `Alice` includes the following `PostMintQuoteBolt11Request` data i
 {
   "amount": <int>,
   "unit": <str_enum["sat"]>,
-  "description": <str|null>
+  "description": <str> // Optional
```
callebtc, line 29: For future reference: https://json-schema.org/understanding-json-schema/reference/object#required

tests/20-test.md
lollerfirst, line 1: These test vectors only allow to check for correct signature verification. Missing the signature creation part with private key.

lollerfirst: Although not strictly necessary, because if the construction of the hashed message is correct, one might expect the signature creation and validation to be correct as well.

PR #194, opened 2024-11-21
On chain mint and melt

thesimplekid:
Depends on https://github.com/cashubtc/nuts/pull/188

- [ ] CDK 
- [ ] Nutshell
- [ ] Nutmix
- [ ] goNuts
- [ ] cashu-ts

Supersedes https://github.com/cashubtc/nuts/pull/107


22.md
```diff
@@ -0,0 +1,231 @@
...
+POST https://mint.host:3338/v1/mint/quote/btconchain
+```
+
+The wallet of `Alice` includes the following `PostMintQuoteBtcOnchainRequest` data in its request:
+
+```json
+{
+  "amount": <int>,
+  "unit": <str_enum["sat"]>,
+  "pubkey": <str>
```
lollerfirst, line 25: NUT-19 says this should be optional, depending on whether or not the `MintInfo` requires it.

22.md
```diff
@@ -0,0 +1,231 @@
...
+POST https://mint.host:3338/v1/mint/btconchain
+```
+
+The wallet `Alice` includes the following `PostMintBtcOnchainRequest` data in its request
+
+```json
+{
+  "quote": <str>,
+  "outputs": <Array[BlindedMessage]>,
+  "witness": <str>
```
lollerfirst, line 119: Here as well. It's not clear if this NUT requires a Mint that requires NUT-19 signatures or only supports them.

lollerfirst:
@thesimplekid  Why do we need new NUTs for this? Can't we just add some appendices to NUT-3, NUT-4 and  NUT-5 on what additional fields a particular `Post{Mint, Melt}OnchainRequest`   should have?

thesimplekid:
> @thesimplekid Why do we need new NUTs for this? Can't we just add some appendices to NUT-3, NUT-4 and NUT-5 on what additional fields a particular `Post{Mint|Melt}OnchainRequest` should have?
> 
> We can then modify NUT-6 to include information about a particular minting and melting method called `btconchain` like this:

I think its a reasonable question, and one way to do it. But I don't really think we should be changing or adding to past nuts. I think it gets confusing, its better to just have new nuts. That being said maybe what id defined in the new nuts could be reduced and and reference the existing nuts and some of it is repetitive. Though I did it this way because this way the nut stands more on its own and you don't have to check another nut as much. But think there are arguments for and against that

prusnak:
NUTs 3, 4 and 5 are mandatory. 

Adding optional fields to them feels wrong.

Also on-chain minting has some specifics (and we already see that via quote locking) and I think this deserves another NUTs.

I would like to also see Liquid supported and I feel this should be again another set of NUTs, although the distinction here is probably not that strong and Liquid support can become a part of the BTC on-chain NUTs.

thesimplekid:
Closing this as its outdated. Will open new onchain spec soon^tm

PR #198, opened 2024-11-27
NUT-21: Clear authentication – NUT-22: Blind authentication

callebtc:
These NUTs describe a blind authentication protocol that uses OAuth2.0 OpenID connect standards to limit access to their mint. Mint operators can restrict access to their mint to a known set of users with registered accounts, while still providing them privacy during the use of the mint with the help of blind signatures (blind auth tokens).

Essentially, we're using ecash tokens for blind authentication. We're using ecash to secure ecash.

There are two NUTs in this PR:

- **NUT-21: Clear authentication** describes how connect a mint to an OpenID Connect service and wallets to obtain a clear auth token (CAT) that identifies the user
- **NUT-22: Blind authentication** describes how to turn a CAT into multiple blind auth tokens (BATs) which can then be used to access the protected endpoints of the mint.

The diagram below illustrates the protocol flow.

```

┌──────────────────────────────────────────────────────────────────────────┐
│ ┌────────┐                     ┌────────┐                  ┌───────────┐ │
│ │  User  │                     │  Mint  │                  │   OpenID  │ │
├─┴────────┘─────────────────────└────────┴──────────────────└───────────┴─┤
│                                                                          │
│                           1. Clear authentication                        │
│                           =======================                        │
│                                                                          │
│                                       Mint registers OpenID service      │
│                                     ◄───────────────────────────────     │
│                                                                          │
│           Wallet GET /v1/info                                            │
│    ─────────────────────────────────►                                    │
│                                                                          │
│                                OpenID login                              │
│    ────────────────────────────────────────────────────────────────►     │
│                                                                          │
│                              Respond with CAT                            │
│    ◄────────────────────────────────────────────────────────────────     │
│                                                                          │
│                           2. Blind authentication                        │
│                           =======================                        │
│                                                                          │
│           Request BAT using CAT                                          │
│    ─────────────────────────────────►                                    │
│                                                                          │
│                Return BAT                                                │
│    ◄─────────────────────────────────                                    │
│                                                                          │
│           Use BAT to access mint                                         │
│    ─────────────────────────────────►                                    │
│                                                                          │
└──────────────────────────────────────────────────────────────────────────┘
```

## Notes
- This requires the use of an external OpenID Connect service. I have used [Keycloak](https://www.keycloak.org/) so far and it was very easy to set up. Another popular choice seems to be [ory/hydra](https://github.com/ory/hydra).
- NUT-XX does not describe how the OAuth2.0 and OpenID Connect protocols work. There are many libraries that can handle this for you but I found it also easy to implement it myself.
- NUT-XX+1 begins after you have obtained an access token from the OpenID connect server. 
- NUT-XX+1 defines a bunch of endpoints that behave very similarly to the normal `/v1/` ecash endpoints we already use. I found it quite convenient to re-use code from the wallet and the mint side to mint, validate, and keep track of spent blind auth tokens.

## Todo:
- [x] Mints should not burn a BAT on an errored request so that external users with a BAT contained in a TokenV4 can re-try melting (instead of having only one shot)
- [x] `client_id` should be part of the info endpoint so it can be dynamic (instead of requiring the OICD use `cashu-client`)
- [ ] Add recommended parameters for `max_bat_mint`, rate-limiting BAT minting (in time), and rotating `auth` keysets


xx+1.md
```diff
@@ -0,0 +1,223 @@
...
+
+Blind authentication tokens (BATs) are used to access the protected endpoints of the mint and make sure that only users that previously presented a valid CAT can access the mint's features such as minting, melting, or swapping ecash. Wallets provide a BAT in the request header when making a request to one of the mint's protected endpoints. The mint parses the header for a BAT, verifies the signature (like with normal ecash as described in [NUT-00][00]), checks if the token has previously been spent, and if not, adds it to its spent BAT token database.
+
+## Blind authentication tokens are ecash
+
+Blind authentication tokens (BATs) are essentially the same as normal ecash tokens and are minted in the same way. They are signed with a special keyset of the mint that has the unit `auth` and a single amount `1`.
```
ok300, line 17: Does this mean NUT-02 "Keysets and fees" needs an update to say `auth` should not be used as unit outside of NUTXX+1?

callebtc: > Does this mean NUT-02 "Keysets and fees" needs an update to say `auth` should not be used as unit outside of NUTXX+1?

Yes, however that's kind of self-evident, don't think NUT-02 needs to refer to this.

xx+1.md
```diff
@@ -0,0 +1,223 @@
...
+
+## Mint
+
+### Signaling protected endpoints and settings
+
+The mint lists each protected endpoint that requires a blind authentication token (BAT) in the `MintBlindAuthSetting` in its [NUT-06][06] info response:
+
+```json
+"XX+1" : {
+  "max_mint": 50,
```
lescuer97, line 194: maybe change to max_bat / max_bat_mint? So it is a little bit more specific?

doesn't really matter though

starbackr-dev:
I’m glad to see 2 NUT specs (Clear +Bind) and it is what is needed. I’ve made some suggestions below which are open for debate and acceptance. 


1. We need to provide a way for the mint to use clear or blinded authed endpoints depending on the types of endpoint. E.g. Minting and melting may require clear authed endpoint while transfer/swap may require blinded auth. This information can be provided using NUT-96 mint information. 

There can also be mints with only blinded auth. E.g. whitelisted pubkeys. Simply call the blind_auth endpoint with Auth as pubkey and get blinded auth tokens. No need for clear auth or open ID. 

```
"XX" : {
  "openid_discovery": "https://mint.com:8080/realms/nutshell/.well-known/openid-configuration",

  "CAT_endpoints": [
    {
        "method": "POST",
        "path": "/v1/auth/blind/mint"
        “Path”: “/v1/mint/quote/bolt11”

    }
  ]
}

"XX+1" : {
“Blind_auth”: "/v1/auth/blind/mint",

"BAT_endpoints": [
    {
        "method": "POST",
        "path": "/v1/swap"
    }
  ]

}
```
This way the protocol is not restricting to use only blinded auth rather the mint can choose to implement clear or blinded auth and provide the information in the manifest/NUT06

2. Always use the “Authorization: Bearer” format. This is typical to all protected APIs in use today. Let’s follow the same pattern. Since BAT starts with authA, the mint should know if this is CAT or BAT. 

        Authorization: Bearer <CAT>/<BAT>

Questions:

How will the CAT expiry time be attached to BAT expiry time?
"max_mint": 50 - is there a benchmark/best practice for an optimal number to use here?

callebtc:
> 1. We need to provide a way for the mint to use clear or blinded authed endpoints depending on the types of endpoint.
This is already possible.

> There can also be mints with only blinded auth. E.g. whitelisted pubkeys... No need for clear auth or open ID

This is not in scope of this spec.

> 2\. Always use the “Authorization: Bearer” format.

This is a matter of taste and makes no practical difference.

xx+1.md
```diff
@@ -0,0 +1,223 @@
...
+      "method": "POST",
+      "path": "^/v1/mint/.*"
+    }
+  ]
+}
+```
+
+`bat_max_mint` is the number of blind authentication tokens (BATs) that can be minted in a single request using the `POST /v1/auth/blind/mint` endpoint.
+
+`protected_endpoints` contains the endpoints that are protected by blind authentication. `method` denotes the HTTP method of the endpoint, and `path` is a regex pattern that must match the path of the URL. In this example, all `/v1/mint/*` endpoints are protected and require blind authentication.
```
ok300, line 210: Can a mint choose its own `protected_endpoints`, or is this set in the spec?

xx+1.md
```diff
@@ -0,0 +1,223 @@
...
+
```
ok300, line 211:
```suggestion
## Wallet Implementation Notes

Wallets would typically provide BATs for each call targeting one of the `protected_endpoints`. However, this brings some complexity (track `protected_endpoints` for each mint, conditional spending of a BAT per API request, etc) and possibly undefined behavior in case of mints where the actual protected endpoints don't match the advertised `protected_endpoints`.

One way to bypass the above is if wallets treat `protected_endpoints` only as an informative hint (i.e. used when showing mint details to the user, but not used to decide for which API calls to use BATs). Calls to any endpoints that are not `/v1/auth/*` could be tried first without a BAT, then in case of an "auth required" error, they could be repeated with a BAT. This makes the wallet dynamically handle the auth configuration of the mint with minimal code overhead and without having to track `protected_endpoints`.
```

callebtc: Not sure I agree or understand this paragraph. It's trivial to check if you need to provide a bat, just a few lines of code.

thesimplekid (approved):

xx+1.md
ok300: The naive wallet impl:
- must persist `protected_endpoints` per mint
- must periodically update `protected_endpoints` per mint
- must regex match the path for each call before calling
- if, due to a mint misconfiguration or bug, the presented `protected_endpoints` doesn't 100% match the actual protected endpoints
  - If an endpoint is protected, but not in `protected_endpoints`, wallet operations would fail
    - Worse: might only happen for some mints for which the wallet dev cannot test against because he's not in the mint's authorized user group, so issue could persist a long time until its found and fixed.
  - If an endpoint is in `protected_endpoints`, but not actually protected, wallet will unnecessarily spend BATs when calling it
  - If the wallet doesn't regularly check and update the mint's `protected_endpoints`, it risks acting on outdated info

Doable, but IMO inviting a lot of [unnecessary complexity](https://grugbrain.dev).

The proposed wallet impl:
- disregard `protected_endpoints` (or only use it for info purposes, like show it in mint details to the user)
- for every mint call, use a try / catch as follows
  - try to call it without auth
  - on `31001: Endpoint requires blind auth` error: retry same call with a BAT auth token

This achieves the same effect with less work, less code and less risk of weird edge-cases.

What do you think?

PR #199, opened 2024-11-28
Add spec for binary token serialization

Egge21M:
This adds some standardization for upcoming raw / binary token serialization, simply prepending the token with a identifier, as well as a version "byte" and encoding everything using `cbor`.

### DIFF
--- a/00.md
+++ b/00.md
@@ -315,8 +315,8 @@ For this the serialised token is prepended with a prefix and a version byte.
 utf8("craw") || utf8(<token_version>) || <serialised_token>
 ```
 
-- Token V3 Version: `utf8("craw") utf8("A") cbor(token_object)`
-- Token V4 Version: `utf8("craw") utf8("B") cbor(token_object)`
+- Token V3 Version: `utf8("craw") || utf8("A") || cbor(token_object)`
+- Token V4 Version: `utf8("craw") || utf8("B") || cbor(token_object)`
 
 [00]: 00.md
 [01]: 01.md


prusnak:
I am not a big fan of this proposal because it is not very flexible and requires cbor dependency for future versions which might not be based on cbor.

IMO much better encoding is just simple:

`CRAW <uint8_t version> <binary data>`

That way an app can very simply check whether it understands the payload by checking the first 5 bytes.

Egge21M:
> I am not a big fan of this proposal because it requires cbor dependency for future versions which might not be based on cbor.
> 
> IMO much better encoding is just simple:
> 
> `CRAW <uint8_t version> <binary data>`
> 
> That way an app can very simply check whether it understands the payload by checking the first 5 bytes.

Very good point, however this is meant to be for transports that are binary only. So we should probably do something like

`utf8(CRAW) <uint8_t version> <binary data>`

00.md
```diff
@@ -308,14 +308,14 @@ cashuBo2F0gqJhaUgA_9SLj17PgGFwgaNhYQFhc3hAYWNjMTI0MzVlN2I4NDg0YzNjZjE4NTAxNDkyMT
...
 
 ```
-cbor(["craw", <version_unsigned_int>, <token_object>])
+utf8('CRAW') <version_uint8> <serialised_token>
 ```
 
-- Token V3 Version: 1
-- Token V4 Version: 2
+- Token V3 Version: `utf8('CRAW') 0x01 cbor(token_object)`
+- Token V4 Version: `utf8('CRAW') 0x02 cbor(token_object)`
```
prusnak, line 318:
```suggestion
- Token V3 Version: `utf8('CRAW') 0x03 <serialised_token>`
- Token V4 Version: `utf8('CRAW') 0x04 <serialised_token>`
```

prusnak (changes requested):

00.md
Egge21M: Actually I would rather keep 1 & 2, as we use `A` & `B` for encoded token

Egge21M: Yeah, thats a good idea. So we will use the same version indicator as the regular tokens encoded as utf8 byte stream. I will also change the prefix to non-capitcal to be consistent with the regular encoding scheme.

prusnak (approved):
ACK

00.md
```diff
@@ -306,6 +306,18 @@ We serialize this JSON using CBOR which can be seen [here](https://cbor.nemo157.
 cashuBo2F0gqJhaUgA_9SLj17PgGFwgaNhYQFhc3hAYWNjMTI0MzVlN2I4NDg0YzNjZjE4NTAxNDkyMThhZjkwZjcxNmE1MmJmNGE1ZWQzNDdlNDhlY2MxM2Y3NzM4OGFjWCECRFODGd5IXVW-07KaZCvuWHk3WrnnpiDhHki6SCQh88-iYWlIAK0mjE0fWCZhcIKjYWECYXN4QDEzMjNkM2Q0NzA3YTU4YWQyZTIzYWRhNGU5ZjFmNDlmNWE1YjRhYzdiNzA4ZWIwZDYxZjczOGY0ODMwN2U4ZWVhY1ghAjRWqhENhLSsdHrr2Cw7AFrKUL9Ffr1XN6RBT6w659lNo2FhAWFzeEA1NmJjYmNiYjdjYzY0MDZiM2ZhNWQ1N2QyMTc0ZjRlZmY4YjQ0MDJiMTc2OTI2ZDNhNTdkM2MzZGNiYjU5ZDU3YWNYIQJzEpxXGeWZN5qXSmJjY8MzxWyvwObQGr5G1YCCgHicY2FtdWh0dHA6Ly9sb2NhbGhvc3Q6MzMzOGF1Y3NhdA
 ```
 
+#### Binary Token
+
+V3 and V4 Token can be transmitted in a binary format when applicable (for example when transmitting via NFC).
+For this the serialised token is prepended with a prefix and a version byte.
+
+```
+utf8("craw") utf8(<token_version>) <serialised_token>
```
callebtc, line 315:
```suggestion
utf8("craw") | utf8(<token_version>) | <serialised_token>
```

where `|` denotes concatenation.

00.md
```diff
@@ -306,6 +306,18 @@ We serialize this JSON using CBOR which can be seen [here](https://cbor.nemo157.
...
+```
+
+- Token V3 Version: `utf8("craw") utf8("A") cbor(token_object)`
+- Token V4 Version: `utf8("craw") utf8("B") cbor(token_object)`
```
callebtc, line 319:
```suggestion
- Token V3 Version: `utf8("craw") || utf8("A") || cbor(token_object)`
- Token V4 Version: `utf8("craw") || utf8("B") || cbor(token_object)`
```

callebtc (approved):
LGTM

PR #201, opened 2024-12-13
Payment Method Refactor

davidcaseria:


purpleKarrot (changes requested):
I am in favor of the technical language and the fact that specifics to the payment method are no longer part of the Request and Resonse type names.

thesimplekid:
Close for #258

PR #202, opened 2024-12-18
Remove binary spec for v3 token

Egge21M:
Tying the old token formats to the binary token versioning was a mistake. This PR removes any mentions of v3 token in the binary token spec (wallets transmitting binary token should use the v4 format or newer) and starts counting binary token version at 1


00.md
```diff
@@ -308,15 +308,14 @@ cashuBo2F0gqJhaUgA_9SLj17PgGFwgaNhYQFhc3hAYWNjMTI0MzVlN2I4NDg0YzNjZjE4NTAxNDkyMT
...
+Token can be transmitted in a binary format when applicable (for example when transmitting via NFC).
 For this the serialised token is prepended with a prefix and a version byte.
 
 ```
 utf8("craw") || utf8(<token_version>) || <serialised_token>
 ```
 
-- Token V3 Version: `utf8("craw") || utf8("A") || cbor(token_object)`
-- Token V4 Version: `utf8("craw") || utf8("B") || cbor(token_object)`
+- Binary Encoding V1: `utf8("craw") || utf8("B") || cbor(token_v4_object)`
```
prusnak, line 318: I would rather say V4 or none for now.

```suggestion
- Binary Encoding: `utf8("craw") || utf8("B") || cbor(token_v4_object)`
```

callebtc: Agree, I would call it V4.

Egge21M: I guess V4 makes sense because the Object structure is the v4 one. Good point

prusnak (approved):

PR #205, opened 2024-12-24
NUT06:  Add Mint TOS to mint info

thesimplekid:


prusnak:
ToS is usually a veeery long document. I think it would be better to have tos_url and link to the document instead (as a bonus the document can be in PDF format that way).

06.md
```diff
@@ -44,6 +44,7 @@ With the mint's response being of the form `GetInfoResponse`:
     "http://mint8gv0sq5ul602uxt2fe0t80e3c2bi9fy0cxedp69v1vat6ruj81wv.onion"
   ],
   "time": 1725304480,
+  "tos": "Terms of Service of the mint.",
```
prusnak, line 47:
```suggestion
  "tos_url": "https://mint.host/tos.pdf",
```

06.md
```diff
@@ -96,6 +97,7 @@ With the mint's response being of the form `GetInfoResponse`:
 - (optional) `icon_url` is the URL pointing to an image to be used as an icon for the mint. Recommended to be squared in shape.
 - (optional) `urls` is the list of endpoint URLs where the mint is reachable from.
 - (optional) `time` is the current time set on the server. The value is passed as a Unix timestamp integer.
+- (optional) `tos` is the Terms of Service of the mint.
```
prusnak, line 100:
```suggestion
- (optional) `tos_url` is the URL pointing to the Terms of Service of the mint.
```

prusnak (commented):
ToS is usually a veeery long document. I think it would be better to replace the tos field with tos_url and provide link to the document instead so the info response is not so big.

As a bonus the document can be easily formatted - let's say in HTML or PDF format.

thesimplekid:
> ToS is usually a veeery long document. I think it would be better to replace the tos field with tos_url and provide link to the document instead so the info response is not so big.
> 
> As a bonus the document can be easily formatted - let's say in HTML or PDF format.

You bring up a good point that including it directly in the info response would increase its size and as this is something that a wallet should fetch often we should avoid this. However, linking to a document without defining a structure or file type would make it very difficult for a client to render and show to an end user. I think the TOS should be just a text string to avoid making it difficult for a wallet to show when the mint is added.

prusnak:
> difficult for a wallet to show when the mint is added.

maybe we should strongly recommend that ToS are in a format easily rendered by a browser such as plain-text, Markdown, PDF or HTML?

prusnak:
> as plain-text, Markdown, PDF or HTML?

Thinking more about this and HTML is probably too dangerous because it can contain some code to deanonymize user :-(

callebtc:
I would vote for URL. Wallets can render it as a web page.

ok300:
I agree the TOS URL is better than including the TOS in `MintInfo`.

I have two proposals:

First, the spec should limit the linked filetype to plaintext.

- this avoids confusion for wallets about how to render it
- this avoids possible attack or deanonimization vectors (HTML, JS, PDF, etc)
- this doesn't add extra dependency burdens on the wallet around rendering (markdown library, etc)
- this is a tried and tested format (most GH projects have a `LICENSE` file that is plaintext, IETF RFCs are in planitext)

Second, why not instead add a `/tos` endpoint to the mint? This could return a plaintext file and replace the need for `tos` or `tos_url` fields in `MintInfo`.

- this gives mint implementations more control (can offer a default TOS)
- this makes it simpler for mint operators to add or change the TOS (likely only editing a file on the mint installation, vs. finding a place to host the TOS file, or if self-hosted: correctly configuring the reverse proxy to point to it)
- this removes further attack or deanonimization vectors (e.g. when the TOS is hosted on a 3rd party and the wallet simply accesses it; 3rd party can add headers, cookies, JS, or log requests)

prusnak:
> Second, why not instead add a `/tos` endpoint to the mint?

I guess this really depends who do we want to optimize for:

a) individuals running the mint - for these it is easier to just drop tos.txt on the mint server and be done with it

b) companies running the mint - where it might be beneficial to have a way how to update the tos without having to bother the admin to change the tos on the mint server (and legal team just asks the webadmin to refresh the tos)

I am fine with optimizing for a) but I wanted to show there is also a b)

callebtc:
IMO, nobody should have to download a PDF file to read the TOS. The TOS should be provided as a web link. It could be a HTML file, it could be a file on github, or anything that can be rendered in a browser. Even if it's an individual of group (a), if they put all the effort into writing a ToS, I'm sure they can find a way to host it somewhere.

06.md
callebtc: PDF is scary, can we just call it a normal URL?

`https://mint.host/tos`

ok300:
> I guess this really depends who do we want to optimize for

My proposals were aimed to optimize for the privacy and risk minimization of mint users, at minimal cost to the mint.

HTML, using a link that points anywhere on the web -- optimizes for the reverse: mint ease of use, at the cost of mint user privacy leaks, deanonimization risks.

prusnak (approved):

lollerfirst:
ACK

lollerfirst (approved):

callebtc (approved):

PR #219, opened 2025-02-01
Keyset Final Expiry

lollerfirst:
Allow a Mint to get rid of very old keysets and nullifiers (spent ecash) associated with them:
keysets now specify a "final expiry" after which the Mint is no longer obligated to serve ecash signed with keys belonging to that keyset.


lollerfirst:
Closing this. Opening a PR -> KeysetsV2. It should be incorporated in keysets v2.

PR #222, opened 2025-02-06
Add mandatory DLEQs to nut-22

Egge21M:
This addition makes sure that `AuthProofs` can not be pinned by enforcing NUT-12 DLEQ proofs on signature reponses

### DIFF
--- a/22.md
+++ b/22.md
@@ -162,6 +162,8 @@ The wallet un-blinds the response to obtain the signatures `C` as described in [
 }
 ```
 
+To prevent pinning, wallets MUST validate the DLEQ proofs `dleq` as defined in [NUT-12](./12.md). Should `AuthProofs` be sent to another user of the mint, it MUST include the `dleq` proof so that the receiving user can validate it.
+
 ## Using blind authentication tokens
 
 The wallet checks the `MintBlindAuthSetting` of the mint to determine which endpoints require blind authentication. Similar to `NUT-21`, the wallet performs a match on the `protected_endpoints` in the `MintBlindAuthSetting` before attempting a request to one of the mint's endpoints. If the match is positive, the wallet needs to add a blind authentication token (BAT) to the request header.
@@ -176,6 +178,10 @@ authA[base64_authproof_json]
 
 This string is a BAT.
 
+> [!CAUTION]
+>
+> To protect the privacy of the wallet, the BAT MUST NOT contain the `dleq` proof when it is sent to the mint in the request header.
+
 ### Request header
 
 We add this serialized BAT to the request header:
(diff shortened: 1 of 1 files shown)


lollerfirst (approved):

22.md
```diff
@@ -183,8 +183,16 @@ and make the request as we usually would.
 
 `AuthProofs` are single-use. The wallet MUST delete the `AuthProof` after a successful request, and SHOULD delete it even if request results in an error. If the wallet runs out of `AuthProofs`, it can [mint new ones](#minting-blind-authentication-tokens) using its clear authentication token (CAT).
 
+### DLEQs
+
+To prevent a mint pinning blind authentication token wallets SHOULD check that the returned `AuthProofs` contain a valid DLEQ as defined in [NUT-12](./12.md).
```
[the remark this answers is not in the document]
Egge21M: Not sure about the semantics TBH. My thought process was that while it would be better for wallets to check this, it would not break the functionality of the NUT or compatibility with mints / wallets if they don't.

22.md
```diff
@@ -183,8 +183,16 @@ and make the request as we usually would.
...
+
 ## Mint
 
+### DLEQs
+
+The mint MUST return a DLEQ proof with all signatures it returns as defined in [NUT-12](./12.md)
```
callebtc, line 194:
```suggestion
The mint MUST return DLEQ proofs for every signature it returns as defined in [NUT-12](./12.md)
```

22.md
callebtc, line 188:
```suggestion
To prevent pinning, wallets MUST validate the DLEQ proofs contained in each `AuthProof` as defined in [NUT-12](./12.md).
```

thesimplekid:
> We should also add the dleq field to the definition of AuthProof if we make it mandatory.

Added in [03f395c](https://github.com/cashubtc/nuts/pull/222/commits/03f395c170b25ad44dc817d34c529ebadb2292c1)

22.md
```diff
@@ -151,9 +151,14 @@ The wallet un-blinds the response to obtain the signatures `C` as described in [
...
-  "secret": str,
-  "C": hex_str
+  "id": <hex_str>,
+  "secret": <str>,
+  "C": <hex_str>,
+  "dleq": {
+    "e": <str>,
+    "s": <str>,
+    "r": <str>
+  }
```
callebtc, line 161: Oops, sorry I think I had a brain hiccup. The wallet should *never* send `r` to the mint because it links the token to its creation. The `r` value is only included in a proof when it's sent from user to user. 

I think we should probably ditch this whole `dleq` field after all and just make sure we validate it when we mint the proof.

callebtc (changes requested):
DLEQ should be checked but doesn't have to be stored and definitely shouldn't be sent to the mint.

22.md
callebtc: Reworded, thank you

callebtc (approved):

PR #236, opened 2025-03-17
change error code numbers

lescuer97:
Error codes seem to be duplicated for blind auth. I just changed the codes to not collide. 

## Tracking Mint implementations

- [ ] Nutshell
- [ ] CDK
- [ ] Nutmix
- [ ] GoNuts
- [ ] MNT

### DIFF
--- a/22.md
+++ b/22.md
@@ -213,10 +213,10 @@ The mint lists each protected endpoint that requires a blind authentication toke
 
 See [Error Codes][errors]:
 
-- `31001`: Endpoint requires blind auth
-- `31002`: Blind authentication failed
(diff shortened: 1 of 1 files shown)


ok300:
What is the overlap?

[NUT-21](https://github.com/cashubtc/nuts/blob/main/21.md) error codes (CAT):
- `30001`
- `30002`

[NUT-22](https://github.com/cashubtc/nuts/blob/main/22.md) error codes (BAT):
- `31001`
- `31002`
- `31003`
- `31004`

lescuer97:
The overlap is that I'm dyslexic and all this time I had not notice the error. Thank you for pointing it out. 

I'll close the pr.

PR #239, opened 2025-03-24
NUT-24: HTTP Payment required

hzrd149:
This PR adds NUT-24 that describes how the HTTP 402 status code + an `X-Cashu` header can be used to require payment for http resources
[readable version](https://github.com/hzrd149/cashu-nuts/blob/402/23.md)

The goal for this NUT is to keep it tightly coupled to HTTP and keep all communication in-band to reduce complexity.


callebtc:
This should reuse the payment request from NUT-18 and specify which fields need to be set https://github.com/cashubtc/nuts/pull/243

23.md
```diff
@@ -0,0 +1,26 @@
...
+
+### Client payment
+
+After receiving a 402 response with a `X-Cashu` header, the client may retry the request with a `cashuB` token in the `X-Cashu` header as payment.
+
+The token MUST be from one of the mints listed in the `mints` array, and MUST be of the same `unit` and greater than or equal to the `amount`
+
+## Payment response
+
+If the server receives tokens from a mint that is not in the `mints` array, an incorrect `unit`, or an insufficient amount of tokens it should respond with a HTTP `400` status code.
```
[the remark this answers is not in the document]
callebtc: The server might want to melt them, I think it's better to leave this to the implementor.

callebtc:
Looks ready. Please leave your reviews. We usually wait for an implementation before we merge.

positiveblue:
It's nice to see more people building on the `HTTP 402` 🎉 

eCash has something really distinct and it's that "the token is the money" which means that people can use it as "proof of payment" (you send it to me and I claim it instantly). This is different from other payment methods where you can do pull based payments (ex: credit cards) or push based but you need to check if the payment was completed outbound (ex: Bitcoin)


We have been working for a while with a [generic flow that supports all kind of payment methods](https://www.l402.org). The trade offs for ecash is 

- Cons: 
  - One extra HTTP trip
- Pros: 
  - One integration, many payment methods
  - Support for multiple offers (ex: you need to become a paid user but can be professional, pro or enterprise)
  - Allows solutions that use third party gateways (you pay somewhere else and they send me a webhook that the offer was paid, easy to swap with current solutions like stripe)
  
  
  
I understand why you want to create a NUT for this. Same of why [x402](x402.org) exists for "strictly on chain payments" but even in that case I would try to bake in the "multiple offer support" in.

Egge21M:
> If you plan to keep `a int` I suggest that you clarify to specify that it be `fractional unit` of the currency for a 402 payment required.

There is no such thing in Cashu. Amounts are always integers of the specified unit

trbouma:
> > If you plan to keep `a int` I suggest that you clarify to specify that it be `fractional unit` of the currency for a 402 payment required.
> 
> There is no such thing in Cashu. Amounts are always integers of the specified unit

So, if "usd" can only specify whole dollars?

Egge21M:
> So, if "usd" can only specify whole dollars?

A single unit of "usd" represents a dollar cent. There are no floats in the Cashu protocol.

trbouma:
> > So, if "usd" can only specify whole dollars?
> 
> A single unit of "usd" represents a dollar cent. There are no floats in the Cashu protocol.

That's ok - if it's clarified in the spec. Currency is usually expressed as a main (base) unit and a fractional unit. I read the spec as the main unit, but it's actually the fractional unit. It wasn't apparent to me. Just wanting to reduce the potential confusion.

FWIW

From wikipedia

Each currency typically has a main currency unit (the [dollar](https://en.wikipedia.org/wiki/Dollar), for example, or the [euro](https://en.wikipedia.org/wiki/Euro)) and a fractional unit, often defined as 1⁄100 of the main unit: 100 [cents](https://en.wikipedia.org/wiki/Cent_(currency)) = 1 [dollar](https://en.wikipedia.org/wiki/Dollar), 100 [centimes](https://en.wikipedia.org/wiki/Centime) = 1 [franc](https://en.wikipedia.org/wiki/Franc), 100 pence = 1 [pound](https://en.wikipedia.org/wiki/Pound_sterling), although units of 1⁄10 or 1⁄1000 occasionally also occur. Some currencies do not have any smaller units at all, such as the [Icelandic króna](https://en.wikipedia.org/wiki/Icelandic_kr%C3%B3na) and the [Japanese yen](https://en.wikipedia.org/wiki/Japanese_yen).

Egge21M:
I think this is out of scope for this NUT / PR. Cashu does not dictate anything about what "unit" can be. It can be a currency or something completely different. Also two mints can have key sets for different things that still share a name. The unit and "what it means" is only the mints responsibility and the spec can not specify anything about that.

thesimplekid:
implemented in athenut https://github.com/thesimplekid/athenut-mint/commit/723f8cd0c8012bedcd6fae7674c5b248b1142dec

23.md
```diff
@@ -0,0 +1,47 @@
...
+    "u": str,
+    "m": Array[str],
+    "p2pk": P2PKOption <optional>
+}
+```
+
+- `a`: The amount required in the specified unit
+- `u`: The currency unit (e.g., "sat", "usd", "api")
+- `m`: Array of mint URLs that the server accepts tokens from
+- `p2pk`: Optional P2PK requirements
```
Egge21M, line 29: Is P2PKOption defined anywhere?

positiveblue:
Bump

```
I would try to bake the "multiple offer support" in
```

gudnuf:
Seems like this new nut is unnecessary with the change in #243 that makes the transport field optional. If the transport field in a payment request is undefined then the client can know to respond with payment in band. 

This PR does define the x-cashu behavior which is useful, but I wonder if it deserves a new nut?

hzrd149:
> Seems like this new nut is unnecessary with the change in https://github.com/cashubtc/nuts/pull/243 that makes the transport field optional. If the transport field in a payment request is undefined then the client can know to respond with payment in band.

It helps to have the HTTP 402 process clearly spelled out in a NUT so that all servers and clients implementing it don't have to worry about 4+ different implementations ( which is usually what happens without clear instructions )

> This PR does define the x-cashu behavior which is useful, but I wonder if it deserves a new nut?

I don't know if it needs it own NUT, but it might make NUT-18 more crowded if it was merged into that

Egge21M:
I think it is worth keeping this in a separate NUT, as it defines the header structure and protocol (both are clearly not part of NUT-18 scope).

hzrd149:
Created an experimental blossom server using NUT-23 for payments https://github.com/hzrd149/morning-glory
Although its not running anywhere at the moment

23.md
lescuer97: I believe it comes from this pr: https://github.com/cashubtc/nuts/pull/243/files

thesimplekid: We should change this to be a nut10 object like in https://github.com/cashubtc/nuts/pull/248

thesimplekid (changes requested):

hzrd149:
Would it make sense to mention using the `Authorization` header with a bearer token as an alternative? or switching the `X-Cashu` header to the `Authorization` for client -> server requests?

Routstr is using the `Authorization` header https://docs.routstr.com/api-reference/introduction#authentication

The biggest downside I can see to using the `Authorization` is that it would prevent clients or servers from using common authentication methods like JWT along side cashu payments. maybe they won't be used together at all but I think its worth considering

9qeklajc:
> Would it make sense to mention using the `Authorization` header with a bearer token as an alternative? or switching the `X-Cashu` header to the `Authorization` for client -> server requests?
> 
> Routstr is using the `Authorization` header https://docs.routstr.com/api-reference/introduction#authentication
> 
> The biggest downside I can see to using the `Authorization` is that it would prevent clients or servers from using common authentication methods like JWT along side cashu payments. maybe they won't be used together at all but I think its worth considering

When I began this project [here](https://github.com/9qeklajc/ecash-402-client?tab=readme-ov-file#example-api-request-with-ecash-payment), I considered separating the Authorization header from the cashu token and came up with [this solution](https://github.com/9qeklajc/ecash-402-client?tab=readme-ov-file#example-api-request-with-ecash-payment). I introduced two distinct headers: X-PAYMENT-SATS for the payment and X-CHANGE-SATS for the change.

I'm considering whether we could use the X-Cashu header for both the payment and the change. Depending on the context, it could be interpreted as follows (if we decide on this approach, we should clarify it in the specification):

    Request -> Payment
    Response -> Change

I will use [this reference](https://github.com/cashubtc/nuts/pull/239#issuecomment-2762693799) as a guide to implement for the client [here](https://github.com/9qeklajc/ecash-402-client?tab=readme-ov-file#example-api-request-with-ecash-payment)

Egge21M:
> The biggest downside I can see to using the `Authorization` is that it would prevent clients or servers from using common authentication methods like JWT along side cashu payments. maybe they won't be used together at all but I think its worth considering

I use both headers simultaneously in npub.cash and I don't think `Authorization` is the right header for a payment. I don't see any benefit of switching from the `X-Cashu` header.

23.md
```diff
@@ -0,0 +1,47 @@
...
+### Server response
+
+HTTP servers may respond with the HTTP status code 402 with a `X-Cashu` header containing an [encoded][18-encoding] NUT-18 [payment request][18] with the following fields:
+
+```
+{
+    "a": int,
+    "u": str,
+    "m": Array[str],
+    "p2pk": P2PKOption <optional>
```
thesimplekid:
```suggestion
    "nut10": NUT10Option <optional>
```

23.md
thesimplekid:
```suggestion
- `nut10`: The required [NUT-10][10] locking condition
```

23.md
```diff
@@ -0,0 +1,47 @@
...
+
+### Client payment
+
+After receiving a 402 response with a `X-Cashu` header, the client may retry the request with a `cashuB` token in the `X-Cashu` header as payment.
+
+The token MUST be from one of the mints listed in the `mints` array, and MUST be of the same `unit` and greater than or equal to the `amount`.
+
+> [!IMPORTANT]
+>
+> **P2PK:** If the service expects a P2PK-locked token, it MUST validate the P2PK script, require [DLEQ proofs][12], and check for double-spending.
```
thesimplekid:
```suggestion
> [!IMPORTANT]
> The transport can be empty! If the transport is empty, we implicitly assume that the payment will be in-band. An example is X-Cashu where the payment is expected in the HTTP header of a request. We can only hope that the protocol you're using has a well-defined transport.
```

thesimplekid (changes requested):

callebtc (approved):
LGTM

PR #240, opened 2025-03-24
Add NFC Payment Requests

lollerfirst:
adds a `"ndef"` transport method to NUT-18 and describes it.

### DIFF
--- a/18.md
+++ b/18.md
@@ -25,7 +25,8 @@ A Payment Request is defined as follows
   "s": bool <optional>,
   "m": Array[str] <optional>,
   "d": str <optional>,
-  "t": Array[Transport]
+  "t": Array[Transport],
+  "p": str <optional>
 }
 ```
 
@@ -38,6 +39,7 @@ Here, the fields are
 - `m`: A set of mints from which the payment is requested
 - `d`: A human readable description that the sending wallet will display after scanning the request
 - `t`: The method of `Transport` chosen to transmit the payment (can be multiple, sorted by preference)
+- `p`: The hex representation of a 33-byte compressed public key for locking the ecash ([NUT-11](11))
 
 ## Transport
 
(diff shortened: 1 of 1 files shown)


18.md
```diff
@@ -76,7 +76,22 @@ The `n` tag specifies the NIPs the receiver supports. At least one tag value MUS
 - type: `post`
 - target: `<endpoint url>`
 
-The execute the payment, the sender makes a `POST` request to the specified endpoint URL with the `PaymentRequestPayload` as the body.
+To execute the payment, the sender makes a `POST` request to the specified endpoint URL with the `PaymentRequestPayload` as the body.
+
+#### NFC (Tap-2-Pay)
+
+- type: `ndef`
+- target: `<empty>`
```
gandlafbtc, line 84: maybe the target could be optionally a pubkey, so that the sender can lock to it? Or maybe it would be better as a tag. Just thinking about the usecase where the receiving device is offline

gandlafbtc (approved):

18.md
lollerfirst: Yea I think it would be better as a tag. `target` is used to specify the target location of where to send the ecash to, but in this case there is no need to. I think it's confusing to use it for anything else, so just leave it empty.

lollerfirst: Again it doesn't make much sense because NDEF is a protocol, not an endpoint or something to reach.

gudnuf:
#243 adds "The transport can be empty! If the transport is empty, we implicitly assume that the payment will be in-band." to NUT-18. I wonder if having an empty transport makes more sense than adding a new transport.

For example, a payment request that I post on my nostr profile with an `ndef` transport is meaningless. It seems to me that the only time this would be used is if the payer reads the request via nfc in which case if the transport field was empty the payer would know to send the token back in the same way that the request was received.

lollerfirst:
Yes. It sounds right.

PR #209, opened 2025-01-06
Include send to pubkey condition in payment request

davidcaseria:
Include an optional way to request a payment locked to a specified public key.

Implementations:
 - [x] [CDK](https://github.com/cashubtc/cdk/pull/526)

PR #204, opened 2024-12-23
Added x-cashu method to NUT-18

Egge21M:
This adds a new method to NUT-18, that allows payment requests to be used with the X-Cashu protocol. This is an attempt to standardize the flow around x-cashu further, to enable interoperability between protected endpoints and wallets.

PR #243, opened 2025-03-26
NUT-18: Add optional P2PK options and transport field

callebtc:
Introduce optional `p2pk` options for specifying P2PK locking parameters in payment requests. Make the `transport` field optional to allow for in-band payment assumptions.

Supersedes and closes #204 
Supersedes and closes #209

Todo:
- [x] use `base64_urlsafe` in encoding example

### DIFF
--- a/18.md
+++ b/18.md
@@ -123,7 +123,7 @@ Here, `id` is the payment id (corresponding to `i` in request), `memo` is an opt
 
 The payment request is serialized using CBOR, encoded in `base64_urlsafe`, together with a prefix `creq` and a version `A`:
 
-`"creq" + "A" + base64(CBOR(PaymentRequest))`
+`"creq" + "A" + base64_urlsafe(CBOR(PaymentRequest))`
 
 ### Example
 
--- a/18.md
+++ b/18.md
@@ -43,9 +43,10 @@ Here, the fields are
 
 ## Payment parameters
 
-The payment request can include *optional* locking conditions the payee requires from the payer. For example, the payee might require a P2PK-locked token so that they can receive payments offline. 
+The payment request can include _optional_ locking conditions the payee requires from the payer. For example, the payee might require a P2PK-locked token so that they can receive payments offline.
 
 ### P2PK
+
 The `p2pk` field contains the `P2PKOption` object that specifies the payees requested [NUT-11][11] P2PK locking options. Its elements are derived from the supported [Tags][11-tags] for P2PK locks. The `P2PKOption` read:
 
 ```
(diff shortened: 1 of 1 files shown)


18.md
```diff
@@ -38,11 +39,35 @@ Here, the fields are
...
+
+The payment request can include _optional_ locking conditions the payee requires from the payer. For example, the payee might require a P2PK-locked token so that they can receive payments offline.
+
+### P2PK
+
+The `p2pk` field contains the `P2PKOption` object that specifies the payees requested [NUT-11][11] P2PK locking options. Its elements are derived from the supported [Tags][11-tags] for P2PK locks. The `P2PKOption` read:
+
+```
+{
+  "pubkeys": Array[str],
```
thesimplekid, line 54: Question on the pubkeys we don't have a `num_sigs` here so if multiple pubkeys are defined any of them can but used or should all of them be used and it should be 1 of n?

callebtc: Forgot num_sigs! Thank you. Should've included all tags.

thesimplekid:
cdk: https://github.com/cashubtc/cdk/pull/699

gudnuf:
Could this be changed to more generally define NUT-10 secrets? Instead of extending NUT-18 every time a new spending condition is introduced, could payment requests have a `nut10` option where `NUT10Option` is something like:

```json
{
   "kind": <str>,
   "data": <str>,
   "tags": [[ "key", "value1", "value2", ...],  ... ], // (optional)}
}
```
For p2pk this is more verbose, but it would allow requests to ask for proofs locked to any condition that follows NUT-10.

gudnuf:
> > Could this be changed to more generally define NUT-10 secrets? Instead of extending NUT-18 every time a new spending condition is introduced, could payment requests have a `nut10` option where `NUT10Option` is something like:
> > ```json
> > {
> >    "kind": <str>,
> >    "data": <str>,
> >    "tags": [[ "key", "value1", "value2", ...],  ... ], // (optional)}
> > }
> > ```
> 
> Not a bad idea, worth considering IMO.

I opened #248 with this proposed change and implemented in cashu-ts: https://github.com/cashubtc/cashu-ts/pull/285

thesimplekid:
> I think with the changes proposed by @gudnuf this LGTM. We can then move to discuss the exact structure of it in NUT-11

I agree doing it this way will prevent having to update the nut in the future for other conditions ie htlc

thesimplekid:
We can close this for https://github.com/cashubtc/nuts/pull/248 right?

Egge21M:
> We can close this for https://github.com/cashubtc/nuts/pull/248 right?

#248 is a branch of this. We can change target or merge it into this PR

gudnuf:
> > We can close this for #248 right?
> 
> #248 is a branch of this. We can change target or merge it into this PR

I just changed the branch for 248 to target main, so I'll close this... don't shoot ✋😮🤚

PR #245, opened 2025-04-09
NUT-04: Add more details about `expiry`

ok300:
For easier reading: https://github.com/ok300/nuts/blob/ok300-nut04-expiry/04.md

- [x] CDK
- [x] Nutshell
- [ ] nutmix

### DIFF
--- a/04.md
+++ b/04.md
@@ -45,13 +45,15 @@ The mint `Bob` then responds with a `PostMintQuoteBolt11Response`:
 }
 ```
 
-Where `quote` is the quote ID and `request` is the payment request to fulfill. `expiry` is the Unix timestamp until which the mint quote is valid. `amount` and `unit` correspond to the same values provided in the request.
+Where `quote` is the quote ID and `request` is the payment request to fulfill. `amount` and `unit` correspond to the same values provided in the request.
 
 `state` is an enum string field with possible values `"UNPAID"`, `"PAID"`, `"ISSUED"`:
 
 - `"UNPAID"` means that the quote's request has not been paid yet.
 - `"PAID"` means that the request has been paid.
 - `"ISSUED"` means that the quote has already been issued.
+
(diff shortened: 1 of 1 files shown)


04.md
```diff
@@ -45,14 +45,16 @@ The mint `Bob` then responds with a `PostMintQuoteBolt11Response`:
...
+Where `quote` is the quote ID and `request` is the payment request to fulfill. `amount` and `unit` correspond to the same values provided in the request.
 
 `state` is an enum string field with possible values `"UNPAID"`, `"PAID"`, `"ISSUED"`:
 
 - `"UNPAID"` means that the quote's request has not been paid yet.
 - `"PAID"` means that the request has been paid.
 - `"ISSUED"` means that the quote has already been issued.
 
+`expiry` is the Unix timestamp until which the mint quote is valid. Note that `expiry` is currently not enforced, instead it is more of a helpful hint from the mint, which allows wallets to tell when the invoice (`request`) expires without having to parse it. It generally holds that, if a quote's `request` can be paid, the quote should transition to `PAID` regardless of `expiry`. Similarly, if a mint quote is `PAID`, the wallet should be able to claim the proofs (i.e. by minting the tokens, which transitions the quote to `ISSUED`) regardless of `expiry`. The field could however be enforced in the future, for example in setups where `request` doesn't have a built-in expiry, for example an onchain address.
+
```
callebtc, line 57: It's a bit too much text for a field without use. I propose:
```suggestion
`expiry` is the Unix timestamp until which the mint quote is valid. Note that the only purpose of this field is to be informative to the user and can not be enforced by the mint.
```

04.md
```diff
@@ -45,14 +45,16 @@ The mint `Bob` then responds with a `PostMintQuoteBolt11Response`:
...
-Where `quote` is the quote ID and `request` is the payment request to fulfill. `expiry` is the Unix timestamp until which the mint quote is valid. `amount` and `unit` correspond to the same values provided in the request.
+Where `quote` is the quote ID and `request` is the payment request to fulfill. `amount` and `unit` correspond to the same values provided in the request.
 
 `state` is an enum string field with possible values `"UNPAID"`, `"PAID"`, `"ISSUED"`:
 
 - `"UNPAID"` means that the quote's request has not been paid yet.
 - `"PAID"` means that the request has been paid.
 - `"ISSUED"` means that the quote has already been issued.
 
+`expiry` is the Unix timestamp until which the `request` can be paid (i.e. the bolt11 invoice expiry). Note that the only purpose of this field is to be informative to the user and can not be enforced by the mint.
```
thesimplekid, line 56:
```suggestion
`expiry` is the Unix timestamp until which the `request` can be paid (i.e. the bolt11 invoice expiry). Note that the only purpose of this field is to be informative to the user and **MUST NOT** be enforced by the mint.
```

gudnuf:
Does this change only apply to minting via the bolt11 method because this is describing the `PostMintQuoteBolt11Response`? Or will all other methods like onchain also be required to NOT enforce the expiry?

I'm asking because it does seem like there are cases where the mint should enforce the expiry in the quote.

thesimplekid (approved):

PR #248, opened 2025-04-15
modify #243 to replace p2pk with nut10

gudnuf:
An alternative update to NUT-18 that would allow for payment requests to define any spending condition supported by NUT-10

### DIFF
--- a/18.md
+++ b/18.md
@@ -26,7 +26,7 @@ A Payment Request is defined as follows
   "m": Array[str] <optional>,
   "d": str <optional>,
   "t": Array[Transport] <optional>,
-  "p2pk": P2PKOption <optional>,
+  "nut10": NUT10Option <optional>,
 }
 ```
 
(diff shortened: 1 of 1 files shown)


18.md
```diff
@@ -39,25 +39,24 @@ Here, the fields are
...
-  "sigflag": str <optional>,
-  "locktime": int <optional>,
-  "refund": Array[str] <optional>
+  "kind": str,
+  "data": str,
+  "tags": [[ "key", "value1", "value2", ...],  ... ], // (optional)
 }
 ```
 
+The payer **MUST** create a `nonce` for each proof locked to the specified condition.
```
Egge21M, line 58: nit: probably out of scope. This is part of NUT-10

lollerfirst: @gudnuf

thesimplekid (approved):

18.md
```diff
@@ -38,11 +39,32 @@ Here, the fields are
...
+
+> [!IMPORTANT]
+> The payee must validate the incoming tokens themselves in order to decide whether they can accept the payment. This includes checking the DLEQ proof and whether the token includes a long-enough timelock to satisfy the payee.
 
 ## Transport
 
 `Transport` specifies methods for sending the ecash to the receiver. A transport consists of a type and a target.
 
+> [!IMPORTANT]
+> The transport can be empty! If the transport is empty, we implicitly assume that the payment will be in-band. An example is X-Cashu where the payment is expected in the HTTP header of a request. We can only hope that the protocol you're using has a well-defined transport.
```
KnowWhoami, line 66: > We can only hope that the protocol you're using has a well-defined transport.

Why is the `Transport` field optional? 
While this may provides flexibility to downstream users to implement thier own Transportation mechanism , it could raise security issues etc  in their protocols ,if  the custom implementation is poorly designed or lacks standarization.

Would it better to explicitly define the `Transport` field or at least ,  specify a recommended / standarized way as default transport method to mitigate such risks??

gudnuf: Prior to this PR, `Transport` was required. Having no defined transport makes sense for use cases like NFC or L402 where we are expected to respond with a token via the same transport that the request was received. For example, with L402 we get a request in the X-cashu header and then our response includes a token in the X-cashu header.

> Would it better to explicitly define the Transport field or at least , specify a recommended / standarized way as default transport method to mitigate such risks??

We don't know what people will build so these recommended transports wouldn't make sense in all cases

thesimplekid:
Merged in cdk

gudnuf:
merged in cashu-ts

PR #250, opened 2025-04-23
NUT-XX  - Mint Remote signer standard

lescuer97:
This is a standard for Mints to be able to communicate with a remote signer for cryptographic operations as well as keyset information.   

Their is also a standard way for signers to derive their keysets from their master key. 

The goal of this nut is to better isolate the private keys of the mint and standardize the communications between the signer and mint. 

Implementations:
- [ ] Nutshell
- [ ] CDK https://github.com/cashubtc/cdk/pull/1257
- [ ] Nutmix https://github.com/lescuer97/nutmix/pull/216


lescuer97:
I still need to add test vectors for the key derivation

nut-xx.md
```diff
@@ -0,0 +1,37 @@
...
+
+Mints will connect to the signer and ask for validation of proofs as well signing blinded messages. This NUT does not specify the transport method needed to communicate with the mint.
+
+The purpose of this signers is only for cryptographic operations it should not keep any state about the proofs and
+blinded messages it sees.
+
+## Protocol
+
+The Mint and the Signer MUST authenticate with each other using mTLS. The services MUST NOT accept any request if they
+are not authenticated with each other.
```
prusnak, line 17: mTLS requirement excludes all of the hardware wallets

I think we should design the document in a way where a remote signer can authenticate via other means

lescuer97: This is a Mint only nut. Could you explain what you mean?

lescuer97: This is an interesting point. I create a section for authentication and specify that mTLS auth is over the network. 

When we get more ways we could just add that. 

Sounds better?

prusnak: Yes, better 👍

nut-xx.md
```diff
@@ -13,11 +13,17 @@ blinded messages it sees.
 
 ## Protocol
 
-The Mint and the Signer MUST authenticate with each other using mTLS. The services MUST NOT accept any request if they
-are not authenticated with each other.
-
 The signer MUST speak the same language using the GRPC protobuf defined in `remote-signer.proto` inside the repository.
 
+### Authetification
```
prusnak, line 18:
```suggestion
### Authentication
```

nut-xx.md
```diff
@@ -0,0 +1,43 @@
+# NUT-XX: Remote signer communications
+
+`optinal`
```
lollerfirst, line 3:
```suggestion
`optional`
```

remote-signer.proto
```diff
@@ -0,0 +1,174 @@
...
+  // rotates the keysets 
+  rpc RotateKeyset (RotationRequest) returns (KeyRotationResponse);
+}
+
+message BlindSignResponse {
+    oneof result {
+        BlindSignatures sigs  = 1;
+        Error error = 2;
+    }
+}
```
davidcaseria, line 19: These messages seem unnecessarily complicated and will result in difficult-to-use types in some languages. Can something like this work?

```suggestion
message BlindSignResponse {
    Error error = 1;
    repeated BlindSignature blind_signatures = 2;
}
```

If the error field is not null, then clients know an error occurred. Otherwise, they can process the rest of the fields of the response as a success.

lescuer97: I think this is actually a good modification. What do you think? @crodas

nut-xx.md
```diff
@@ -0,0 +1,43 @@
...
+`optional`
+
+---
+
+This NUT define a standard way for Mints to communicate with a remote signer. The mint can protect its private key by isolating it to a service.
+
+Mints will connect to the signer and ask for validation of proofs as well signing blinded messages. This NUT does not specify the transport method needed to communicate with the mint.
+
+The purpose of this signers is only for cryptographic operations it should not keep any state about the proofs and
+blinded messages it sees.
```
thesimplekid, line 12: I think we should remove or modify this line as a signer may want to keep some information about proofs if the signer wants to implement signing polices.

remote-signer.proto
```diff
@@ -0,0 +1,162 @@
...
+  rpc BlindSign(BlindedMessages) returns (BlindSignResponse);
+  rpc VerifyProofs(Proofs) returns (BooleanResponse);
+  // returns all the keysets for the mint
+  rpc Keysets(EmptyRequest) returns (KeysResponse);
+  // rotates the keysets
+  rpc RotateKeyset(RotationRequest) returns (KeyRotationResponse);
+}
+
+message BlindSignResponse {
+  Error error = 1;
```
thesimplekid, line 15: This maybe me just not reading the proto correctly, Does the error need to be optional?

prusnak: All fields in proto3 spec are optional. 

Only old proto spec distinguishes between required and optional fields. 

Quoting from documentation: Required fields are considered harmful by so many they were removed from proto3 completely.

thesimplekid: Ah okay thanks

remote-signer.proto
```diff
@@ -0,0 +1,162 @@
...
+  BlindSignatures sigs = 2;
+}
+
+message BlindedMessages {
```
davidcaseria, line 19: Because signing blinded messages and verifying proofs are done in separate functions, should we include metadata in these requests for what operation is being performed and an ID to correlate requests?

lescuer97: I think that we might want the metadata field to be a somewhat open ended.
What do you think of this:

```
message Proofs {
  repeated Proof proof = 1;
  map<string, string> metadata = 2;
}
```

lescuer97: What I would like to avoid is having to update the NUT everytime a new metadata field is needed. How do you think we should transmit this information?

davidcaseria:
```suggestion
message BlindedMessages {
  repeated BlindedMessage blinded_messages = 1;
  Operation operation = 2;
  string correlation_id = 3;
}

enum Operation {
  MINT = 1;
  MELT = 2;
  SWAP = 3;
}
```

lescuer97: > I like it. I'll implement the latest changes to the proto file to my PR

what do you think is better? having the key value map or the explicit metadata?

lescuer97: yeah,taking into account what you guys have said. I think I agree. I'll add this fields to necessary functions

remote-signer.proto
```diff
@@ -0,0 +1,162 @@
...
+  uint32 max_order = 3;
+}
+
+enum CurrencyUnitType {
+  SAT = 0;
+  MSAT = 1;
+  USD = 2;
+  EUR = 3;
+  AUTH = 4;
+}
```
asmogo, line 74: I suggest renaming these enums to match the protobuf style guide.
https://protobuf.dev/programming-guides/style/#enums
Also, using the UNSPECIFIED / UNKNOWN value in position 0 is recommended.

remote-signer.proto
lescuer97: added this 2 fields to both the signing and proof verification functions

remote-signer.proto
lescuer97: thank you for the observation. I have changed the Enums

remote-signer.proto
```diff
@@ -0,0 +1,162 @@
...
+message Proofs {
+  repeated Proof proof = 1;
+}
+
+message Proof {
+  uint64 amount = 1;
+  string keyset_id = 2;
+  bytes secret = 3;
+  bytes C = 4;
+  optional Witness witness = 5;
```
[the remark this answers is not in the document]
thesimplekid: Verifying the witness does not require the mint keys so not needed by the signer. Just need to make sure we are still verifying it in the mint.

remote-signer.proto
```diff
@@ -0,0 +1,162 @@
...
+}
+
+message Keys {
+  map<uint64, bytes> keys = 1;
+}
+
+message RotationRequest {
+  CurrencyUnit unit = 1;
+  uint64 input_fee_ppk = 2;
+  uint32 max_order = 3;
```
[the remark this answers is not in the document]
lescuer97: I don't think this is better the signer should probably have the control of the ordering.

remote-signer.proto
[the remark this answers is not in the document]
lescuer97: just removed the witness

remote-signer.proto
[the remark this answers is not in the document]
thesimplekid: This allows the mint to select what denominations the keyset should be. Using only the max_order forces the signer to assume power of two which is not defined in spec and the signer should be able to support other denomination choices.

remote-signer.proto
```diff
@@ -0,0 +1,173 @@
...
+message BlindedMessages {
+  repeated BlindedMessage blinded_messages = 1;
+  Operation operation = 2;
+  string correlation_id = 3;
+}
+
+// Represents a blinded message
+message BlindedMessage {
+  uint64 amount = 1;
+  string keyset_id = 2;
```
davidcaseria, line 35: Should we change this to bytes to use the bytes representation of the keyset ID instead of the hex string version?

lescuer97:
> I think this NUT should propose the gRPC notification event when keys are updated, maybe [using streaming](https://grpc.io/docs/languages/go/basics/#server-side-streaming-rpc)?

I don't think  this is worth it because the rotation command will normally come from the mint

remote-signer.proto
lescuer97: that seems fine to me

remote-signer.proto
```diff
@@ -0,0 +1,174 @@
...
+  repeated KeySet keysets = 2;
+}
+
+message KeySet {
+  bytes id = 1;
+  CurrencyUnit unit = 2;
+  bool active = 3;
+  uint64 input_fee_ppk = 4;
+  Keys keys = 5;
+  uint64 version  = 6;
```
crodas: The [final_expiry](https://github.com/cashubtc/cdk/pull/1032/files#diff-c10602a1f5c7a90d9fcecbece2af8030819dd6ed65e17499cadab17bd6ac97d7R65) is needed

remote-signer.proto
```diff
@@ -0,0 +1,174 @@
...
+  optional BlindSignatureDLEQ dleq = 4;
+}
+
+message BlindSignatureDLEQ {
+  bytes e = 1;
+  bytes s = 2;
+}
+
+// Witness type
+message Witness {
```
crodas: https://github.com/cashubtc/cdk/pull/1032#discussion_r2324560676

remote-signer.proto
lescuer97: Added it to the structure and the keyset rotation call

remote-signer.proto
crodas: Here is the latest version

https://github.com/cashubtc/cdk/pull/1055/files#diff-c10602a1f5c7a90d9fcecbece2af8030819dd6ed65e17499cadab17bd6ac97d7

lescuer97:
@crodas  I added the final_expiry to be optional

vnprc:
There are a number of gaps and vulnerabilities in the deterministic derivation path function that are addressed in this proposed NUT spec: https://github.com/cashubtc/nuts/pull/292

Namely, leading and trailing whitespace, capitalization differences, and functionally equivalent unicode character sequences can cause visually identical currency units to hash to different derivation paths. Also, the lack of a reserved range for currently hard coded currency unit derivation paths will lead to derivation path conflicts. These properties could be exploited by attackers using specially crafted currency unit strings.

If y'all would like to review PR 292 I think it would be complementary to this spec proposal and they could potentially go at the same time.

nut-xx.md
```diff
@@ -0,0 +1,72 @@
...
+
+Implementations **MUST** accept currency unit labels case-insensitively and ignore leading or trailing ASCII whitespace. When serializing a unit into JSON (e.g., as part of a keyset description in NUT-01 responses), implementations **SHOULD** emit the uppercase representation of the unit so that mints and wallets display consistent labels.
+
+Before deriving an index, the input label **MUST** be transformed as follows:
+
+1. Remove leading and trailing ASCII whitespace characters (space, tab, carriage return, line feed).
+2. Apply Unicode Normalization Form C (NFC).
+3. Convert the normalized string to uppercase using Unicode-aware semantics.
+
+| Input unit   | Canonical form | Index        |
```
aaronmcdaid: When you say 'index' here, do you mean 'unit reference'? This is essentially the first mention of 'index' in this doc, and I don't know what it is.

(I see many others have commented already. Apologies in advance if I now ask a few questions that have already been answered 😀)

lescuer97: hi! it just talks about the actual index number that is going to be used later on the derivation path

lescuer97: hi! it just talks about the actual index number that is going to be used later on the derivation path.

Should probably use better naming

protos/remote-signer.proto
a1denvalu3, line 1: shouldn't we create a "protos" dir?

nut-xx.md
```diff
@@ -0,0 +1,78 @@
...
+
+For compatibility reasons all signers SHOULD implement the following BIP32 derivation path.
+
+- m = master key
+- 129372' (UTF-8 for 🥜)
+- unit_reference = Big endian encoded integer of the first 4 bytes of the sha256 hash of the canonical unit string reference modulo by 2^31. We modulo because we want to stay inside the `2^31 - 1` range.
+  ex: sha256sum('auth')[:4] = bdf49c3c = 3186924604
+  3186924604 % 2^31 = 1039440956.
+- version: uint32
+- index_of_amount = index of and the amounts of the keyset as if the where laid in an array. ex: [1, 2, 4, 8, 16, ...]
```
a1denvalu3: this is a bit confusing

nut-xx.md
a1denvalu3: I think nonce index would be less confusing

a1denvalu3 (approved):

protos/remote-signer.proto
thesimplekid: Yes, like we do for tests I think.

nut-xx.md
lescuer97: what do you suggest on naming?

nut-xx.md
```diff
@@ -0,0 +1,78 @@
...
+The signer MUST NOT accept any request if authentication has not occurred.
+
+#### Over the Network
+
+The Mint and the Signer MUST authenticate with each other using mTLS.
+
+## Seedphrase usage
+
+[BIP-39](https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki) mnemonic seedphrase must be used for generation
+of the keysets.
```
robwoodgate:
```suggestion
[BIP-39](https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki) mnemonic seedphrase must be used for generation of the keysets.
```

nut-xx.md
```diff
@@ -0,0 +1,78 @@
...
+| `SAT`        | `SAT`          | `1967237907` |
+| `nuts`       | `NUTS`         | `1502388627` |
+| `USD`        | `USD`          | `577560378`  |
+| `usD`        | `USD`          | `577560378`  |
+| `café`       | `CAFÉ`         | `642348965`  |
+| `cafe\u0301` | `CAFÉ`         | `642348965`  |
+| `eurc`       | `EURC`         | `1321886550` |
+
+NOTE: Mints **MUST** make sure that the unit_reference integer has not been repeated before for a different normalized
+unit string.
```
robwoodgate:
```suggestion
NOTE: Mints **MUST** make sure that the unit_reference integer has not been repeated before for a different normalized unit string.
```

nut-xx.md
```diff
@@ -0,0 +1,78 @@
...
+  3186924604 % 2^31 = 1039440956.
+- version: uint32
+- index_of_amount = index of and the amounts of the keyset as if the where laid in an array. ex: [1, 2, 4, 8, 16, ...]
+
+`m / 129372' / unit_reference' / version' / index_of_amount'`
+
+## Configuration
+
+The Signer can also be configured to have some rate limiting features. Limiting the amount of minting that can happen by
+time or single action.
```
robwoodgate:
```suggestion
The Signer can also be configured to have some rate limiting features. Limiting the amount of minting that can happen by time or single action.
```

nut-xx.md
lescuer97: I changed the name and the explanation. I think it's clearer now.

ye0man:
Closing due to age and lack of demand / interest from multiple implementations. @thesimplekid noted cdk may not be compliant with this spec any longer. @lescuer97 please feel free to escalate if you have any thoughts

prusnak:
I agree that this is probably better if handled by Cashu implementations individually. CDK already has concept of signatory and for example running the signatory inside of a secure enclave (current or future ones) might require different protocol anyway (specific to a particular secure enclave for example).

PR #255, opened 2025-05-02
Secret and well-known secret lengths

a1denvalu3:
This PR specifies that a secret should be a 64 hex string. A well-known secret may exceed this constraint, but their limits are to be decided on a case-by-case basis.


thesimplekid:
Closes #44

PR #261, opened 2025-05-14
[NUT-XX] Compact Nut Filters

a1denvalu3:
Golomb-Coded Set Filters provide a compressed data structure for testing set membership.

Wallets can fetch `GET v1/filter/spent` and download the compressed set, then check to see if the ecash notes they have belong in the set (and thus are spent).

This drastically improves privacy in some cases:
* During recovery from seed phrase.
* After a cold-start, wallets might be checking the state of pending (sent) ecash in bulk.

- [x] CDK https://github.com/cashubtc/cdk/pull/754
- [x] Nutshell https://github.com/cashubtc/nutshell/pull/752
- [x] cashu-ts https://github.com/cashubtc/cashu-ts/pull/313
- [ ] nutmix


prusnak:
IIRC the GCS filters have:
- no false negatives
- false positives 

This is fine for BIP158 blockfilters (the worst that can happen is that you download a block which is not relevant for you).

If I understand this correctly, this can be catastrophic for what you are trying to achieve here, right? False positive will mark the note as spent even though it is unspent.

lollerfirst:
> Or do you suggest to use the filter to identify "surely unspent" (unmatched by GCS) notes and then query only the ones that have been "possibly spent" (all matches by GCS)?

Yes that's the idea. Unmatched notes are surely unspent and matched notes are possibly spent with probability $p = 1 - 1/m$. This is only meant for client side use, so the Mint still uses the DB to decide whether to accept the notes.

PR #262, opened 2025-05-20
fix: PostRestoreRequest only contains B_s instead of BlindedMessages

tdelabro:


09.md
```diff
@@ -20,7 +20,7 @@ With the data being of the form `PostRestoreRequest`:
 
 ```json
 {
-  "outputs": <Array[BlindedMessages]>
+  "B_s": <Array[<hex_str>]>
```
lollerfirst, line 23:
```suggestion
  "blind_messages": <Array[<hex_str>]>
```

09.md
```diff
@@ -35,7 +35,8 @@ The mint `Bob` then responds with a `PostRestoreResponse`.
 }
 ```
 
-The returned arrays `outputs` and `signatures` are of the same length and for every entry `outputs[i]`, there is a corresponding entry `signatures[i]`.
+The returned arrays `outputs` and `signatures` are of the same length and for every entry `B_s[i]`, there are corresponding `outputs[i]` and `signatures[i]` entries,
+whith `outputs[i].B_ == B_s[i]`.
```
lollerfirst, line 39:
```suggestion
whith `outputs[i].B_ == blind_messages[i]`. 
```

prusnak:
Yea. We can keep this mind when designing the next version of the protocol but otherwise let's not touch what is not broken.

NACK

thesimplekid:
NACK

PR #263, opened 2025-05-21
Add Supported Units and ISO 4217 Minor Unit Clarification to NUT-01

robwoodgate:
NUT-00 references "Supported Units" in [NUT-01](https://github.com/cashubtc/nuts/blob/main/01.md), but this wasn't defined, leaving ambiguity in the supported units and the amount basis for fiat currencies.

This PR fixes that.

- Adds "Supported Currency Units" section to NUT-01, covering ~~mandatory~~ sat and optional other units, including msat, auth, ISO 4217 codes (usd, eur, etc.), and stablecoins (usdt, eurc, gyen etc.).

- Requires ISO 4217 currencies and stablecoins to use the currency’s minor unit for keyset amounts (e.g., cents for usd, whole units for jpy).

- Updates "Example responses" with a usd example showing amounts in cents (e.g., "1" = 0.01 USD).

Impact:
Clears up ambiguity for fiat and stablecoin amounts, and makes NUT-01 complete and consistent with NUT-00.

EDIT: Following review comments, have made sat optional - mint can support any currency it can mint/melt either directly or indirectly. Have also referenced BTC as a possible unit for clarity and avoidance of doubt as to the unit basis


01.md
```diff
@@ -14,6 +14,23 @@ The mint responds only with its `active` keysets. Keyset are `active` if the min
 
 Note that a mint can support multiple keysets at the same time but will only respond with the active keysets on the endpoint `GET /v1/keys`. A wallet can ask for the keys of a specific (active or inactive) keyset via the endpoint `GET /v1/keys/{keyset_id}` (see [NUT-02][02]).
 
+## Supported Currency Units
+
+A mint **MUST** support the Bitcoin `sat` unit, and may **OPTIONALLY** support any other currency units, including:
```
thesimplekid, line 19: I don't see why  mint must support bitcoin. If they want to be fiat only that should be within spec.

robwoodgate: I conceptually agree - however, I added the `must` primarily for backwards compatibility, as mandatory NUTS 04 & 05 are limited to `unit="sat"` and optional NUT-20 specifies unit enum as sat only: `<str_enum["sat"]>`

thesimplekid: I think this is more imprecisely using sat as an example/default then a real requirement of the spec and it wouldn't be breaking to allow any.

lollerfirst: I agree with @thesimplekid

robwoodgate: Great, have tweaked... Sat is no longer mandatory, and mint can support any units it can mint and melt.

Have also added BTC as a unit for the avoidance of doubt in the future, keeping consistent with major/minor units.

thesimplekid (approved):

PR #149, opened 2024-07-13
NUT-11: `SIG_ALL` sign aggregated message all inputs and ouptuts

callebtc:
This PR specifies the signature flag `SIG_ALL` for swap and for melt transactions involving P2PK-locked or HTLC-locked inputs. It also adds a new tag for P2PK (also used in HTLC) called `n_sigs_refund` that specifies the signature threshold for the `refund` pubkeys (enforced after a timelock).

`SIG_ALL` means that the witness of the _first input_ must be a signature on an concatenated message of the form `proof.secret_n.. . || output.B'_n...` in the case of a swap covering all of the inputs' `secret` fields and all of the outputs' blinded message fields `B_`. In the case of a melt, we also need to include the `quote_id` in the signed message.

##### Aggregation for `swap`

If a swap transaction has `n` inputs and `m` outputs, the message to sign becomes:

```
msg = secret_0 || ... || secret_n || B_0 || ... || B_m
```

##### Aggregation for `melt`

For a melt transaction, the message to sign is composed of all the inputs, the quote ID being paid, and the [NUT-08][08] blank `outputs`.

If a melt transaction has `n` inputs, `m` blank outputs, and a quote ID `quote_id`, the message to sign becomes:

```
msg = secret_0 || ... || secret_n || B_0 || ... || B_m || quote_id
```

Here, `||` denotes string concatenation. The `B_` of each output is **a hex string**.

Changes:
- fix SIG_ALL
- add `n_sigs_refund`
- remove `BlindedMessage.witness` output witness

Todo:

- [x] specify if only valid for `/swap` or also for `/melt` (for both signature flags)
- [x] `SIG_ALL` test vectors
- [ ] explicit error if no outputs are provided for `SIG_ALL`.

Tracking:
- [x] cdk-mintd (PR: https://github.com/cashubtc/cdk/pull/862)
- [x] nutshell (PR: https://github.com/cashubtc/nutshell/pull/735)

PR #272, opened 2025-06-23
Clarifications to multisig

robwoodgate:
Consolidates the changes I recommended as part of my review of PR #149

### DIFF
--- a/11.md
+++ b/11.md
@@ -150,8 +150,12 @@ Here, `||` denotes string concatenation. The `B_` of each output is **a hex stri
 
 ### Multisig
 
+Cashu offers two levels of multi-signature protection: `Locktime MultiSig` and `Refund MultiSig`, which are activated depending on the status of the proof's `locktime` tag.
+
+#### Locktime MultiSig
+
 > [!NOTE]
-> Multisig conditions only apply if the `locktime` tag is not present, or is a timestamp in the future.
+> Locktime Multisig conditions only apply if the `locktime` tag is not present, or is a timestamp in the future.
 
 If the `pubkeys` tag is present, the `Proof` is spendable only if a valid signature is given by at least ONE of the public keys contained in the `Secret.data` field or the `pubkeys` tag.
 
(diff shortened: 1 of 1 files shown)


11.md
```diff
@@ -150,7 +150,12 @@ Here, `||` denotes string concatenation. The `B_` of each output is **a hex stri
 
 ### Multisig
 
-If the tag `n_sigs` is a positive integer, the mint will require valid signatures from `n_sigs` public keys contained in the `Secret.data` field and the `pubkeys` tag.
+> [!NOTE]
+> Multisig conditions only apply if the `locktime` tag is not present, or is a timestamp in the future.
```
callebtc: Since we also have refund multisig, this note isn't accurate anymore, right?

robwoodgate: The way it is drafted, `MultiSig` refers to the case where `data` + `pubkeys` can be used, and `n_sigs` matters.

So in this case, the Note is absolutely accurate.

The next major heading down (`Locktime`) then refines the situation and handles refund keys and `n_sigs_refund`.

Maybe now we have two "multisig" conditions, we should find a better label to distinguish them... 

Eg: `Locktime Multisig` and `Refund Multisig`... or `Primary Multisig` and `Refund Multisig` etc

robwoodgate: @callebtc  - Would this work better do you think?

```markdown
### Multisig

Cashu offers two levels of Multi-signature protection: `Locktime MultiSig` and `Refund MultiSig`, which are activated depending on the status of the proof's `locktime` tag.

#### Locktime MultiSig

> [!NOTE]
> Locktime Multisig conditions only apply if the `locktime` tag is not present, or is a timestamp in the future.

If the `pubkeys` tag is present, the `Proof` is spendable only if a valid signature is given by at least ONE of the public keys contained in the `Secret.data` field or the `pubkeys` tag.

If the `n_sigs` tag is a positive integer, the mint will require at least `n_sigs` of those public keys to provide a valid signature.

If the number of public keys with valid signatures is greater or equal to the number specified in `n_sigs`, the transaction is valid. The signatures are provided in an array of strings in the `P2PKWitness` object.

> [!CAUTION]
>
> Because Schnorr signatures are non-deterministic, we expect a minimum number of unique public keys with valid signatures instead of expecting a minimum number of signatures.

Expressed as an "n-of-m" scheme, `n = n_sigs` is the number of required signatures and `m = 1 (data field) + len(pubkeys tag)` is the number of public keys that could sign.

#### Locktime

If the tag `locktime` is the unix time and the mint's local clock is greater than `locktime`, the `Proof` becomes spendable by anyone, except if the following condition is also true.

> [!NOTE]
> A `Proof` is considered spendable by anyone if it only requires a `secret` and a valid signature `C` to be spent (which is the default case).

#### Refund MultiSig

If the `locktime` tag is in the past and the `refund` tag is present, the `Proof` is spendable only if a valid signature is given by at least ONE of the `refund` pubkeys.

If the `n_sigs_refund` tag is present, the mint will require at least`n_sigs_refund` of the `refund` pubkeys to provide a valid signature.

#### Complex Example
.... rest of file
```

robwoodgate: I've pushed an amendment to distinguish between Locktime and Refund multisig. LMK if you want any other tweaks or changes.

PR #273, opened 2025-07-09
Add NUT-XX for batched minting

Egge21M:
This commit introduces a new specification, NUT-XX, for batched mint operations. It allows wallets to mint multiple proofs in a single transaction, improving efficiency.

### DIFF
--- a/xx.md
+++ b/xx.md
@@ -18,7 +18,7 @@ POST https://mint.host:3338/v1/mint/{method}/check
 Content-Type: application/json
 
 {
-  "quote": [ "quote_id_1", "quote_id_2", … ]
+  "quotes": [ "quote_id_1", "quote_id_2", … ]
 }
 ```
 
@@ -60,7 +60,7 @@ POST https://mint.host:3338/v1/mint/{method}
 Content-Type: application/json
 
 {
-  "quote":   [ "quote_id_1", "quote_id_2", … ],
+  "quotes":   [ "quote_id_1", "quote_id_2", … ],
   "outputs": [ BlindedMessage_1, BlindedMessage_2, … ],
   "signature": [signature_1, signature_2, ... ]
 }
@@ -93,7 +93,7 @@ Example:
 
 ```json
 {
-  "quote":   [ "locked_quote_id_1", "quote_id_2", "locked_quote_id_3" ],
+  "quotes":   [ "locked_quote_id_1", "quote_id_2", "locked_quote_id_3" ],
   "outputs": [ BlindedMessage_1, BlindedMessage_2, ... ],
   "signature": [signature_1, null, signature_3 ]
 }


Egge21M:
I started this as a draft to gather feedback by mint implementors (as the wallet side is trivial). The Cashu side of things should be very easy to do, however getting payment state of multiple payment requests might introduce challenges.

xx.md
```diff
@@ -0,0 +1,81 @@
...
+Content-Type: application/json
+
+{
+  "quote": [ "quote_id_1", "quote_id_2", … ]
+}
+```
+
+- **quote**: an array of quote IDs previously obtained via the [NUT-04 creation process][04-creation].
+
+The mint returns a JSON array of quote objects, each containing:
```
davidcaseria, line 27: What is the behavior if one of the quote IDs is not found?

Egge21M: I think the best would be to simply omit it from the response. That way the client could handle that case and still use the available data without resending the request

xx.md
```diff
@@ -0,0 +1,81 @@
...
+## 2. Executing the Batched Mint
+
+Once all quoted payments are confirmed, the wallet mints the tokens by calling:
+
+```http
+POST https://mint.host:3338/v1/mint/{method}
+Content-Type: application/json
+
+{
+  "quote":   [ "quote_id_1", "quote_id_2", … ],
```
lescuer97: what happens if we have repeated quote_ids? should we fail it?

xx.md
```diff
@@ -0,0 +1,81 @@
...
+  "outputs": [ BlindedMessage_1, BlindedMessage_2, … ]
```
lescuer97: how would you thing NUT-20 signed mint should be handled? 

may be just a new field called `signatures` and if they are mix we just do a null value?

xx.md
Egge21M: Yes, I think making that an invalid request is the best thing to do here. I guess as long as the output sum still matches, a mint could go through with the request, but this feels like adding to much complication. Clients should not include quotes twice.

xx.md
```diff
@@ -0,0 +1,104 @@
...
+
+Before minting, the wallet must verify that each mint quote has been paid.
+It does this by sending:
+
+```http
+POST https://mint.host:3338/v1/mint/{method}/check
+Content-Type: application/json
+
+{
+  "quote": [ "quote_id_1", "quote_id_2", … ]
```
vnprc: Perhaps out of scope for this NUT but I have identified a need to query quotes by NUT-20 locking key for the mining share payment method NUT, aka ehash. I intend to propose this change to NUT-04.

vnprc: It's looking like my proposal might turn into an API that queries quote ID by NUT-20 locking key. Which means I could make one API call to get the quote IDs and then call this API as-is to mint the tokens.

So I guess all I'm saying is LGTM!

xx.md
```diff
@@ -0,0 +1,104 @@
...
+- Signatures for NUT-20 quotes can be added to the `signature` key of the request.
+- Signatures need to be in the same index as the matching quote_id in the `quote` key.
+- If a request contains both signed and unsigned quotes, all unsigned quotes need to map to `null` in the `signature` array.
+- As soon as there is a single signed quote in the request: `quote.length === signature.length`
+
+Example:
+
+```json
+{
+  "quote":   [ "locked_quote_id_1", "quote_id_2", "locked_quote_id_3" ],
```
callebtc:
```suggestion
  "quotes":   [ "locked_quote_id_1", "quote_id_2", "locked_quote_id_3" ],
```

xx.md
```diff
@@ -0,0 +1,104 @@
...
+## 2. Executing the Batched Mint
+
+Once all quoted payments are confirmed, the wallet mints the tokens by calling:
+
+```http
+POST https://mint.host:3338/v1/mint/{method}
+Content-Type: application/json
+
+{
+  "quote":   [ "quote_id_1", "quote_id_2", … ],
```
callebtc:
```suggestion
  "quotes":   [ "quote_id_1", "quote_id_2", … ],
```

xx.md
callebtc:
```suggestion
  "quotes": [ "quote_id_1", "quote_id_2", … ]
```
maybe better nomenclature?

xx.md
```diff
@@ -0,0 +1,104 @@
+# NUT-XX: Batched Mint
+
+`optional`
+
+`depends on: NUT-04`
+
+This spec describes how a wallet can mint multiple proofs in one batch operation.
```
vnprc:
```suggestion
This spec describes how a wallet can mint multiple quotes in one batch operation.
```

vnprc: I don't know if this is a terminology difference between implementations. My understanding is that a proof is an unblinded tuple of secret/signature. Multiple proofs in a quote are guaranteed unless your mint amount is a power of 2. I think it would be less ambiguous to say multiple quotes in the NUT description.

PR #277, opened 2025-08-04
add NUT-20 quote id lookup API spec

vnprc:
optional API endpoint for looking up quotes by NUT-20 pubkey

### DIFF
--- a/20.md
+++ b/20.md
@@ -172,17 +172,119 @@ See [Error Codes][errors]:
 - `20008`: Mint quote with `pubkey` but no valid `signature` provided for mint request.
 - `20009`: Mint quote requires `pubkey` but none given or invalid `pubkey`.
 
+## Quote Lookup by Public Key
+
+Wallets **MAY** retrieve mint quote information by providing the public key(s) used during quote creation. This allows quote recovery when the quote ID is unknown.
+
+### Lookup Request
+
+To lookup quotes by public key, the wallet makes a `POST /v1/mint/quote/lookup` request with the following `PostMintQuoteLookupRequest`:
+
+```json
+{
+  "pubkeys": [<str>]
+}
+```
+
+Where `pubkeys` is an array of hex-encoded public keys to lookup quotes for.
+
+```http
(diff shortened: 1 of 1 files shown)


20.md
thesimplekid, line 1: I think we should make this a new optional nut. Changing nuts should be reserved to when we have to as its much more confusing to track changed to a nut, then support for a new nut.

20.md
```diff
@@ -172,18 +172,120 @@ See [Error Codes][errors]:
...
+      "quote": <str>,
+      "request": <str>,
+      "state": <str_enum[STATE]>,
+      "expiry": <int>,
+      "amount": <int>,
+      "unit": <str_enum[UNIT]>
+    }
+  ]
+}
+```
```
thesimplekid: This structure does not account for future response types. For example bolt12 does not have a `state`. I think this response should either be an enum of MintQuoteResponses or just a list of quote ids. I lean towards the later even though it requires another network call to the the actual quote info.

vnprc: Yeah I struggled with the duplication of data structures here. My first iteration trying to solve this problem was a share hash -> quote ID lookup like you suggest. I changed to pubkey because I need to implement NUT-20 anyway and it solves some use cases you have identified, so it's just a more general solution.

Why do you like the quote ID solution better?

thesimplekid: I had a typo in my original response I meant does **not** account for future types. Looks like you already addressed this in the updated commit.

XX.md
```diff
@@ -0,0 +1,108 @@
...
+
+The mint responds with a `PostMintQuoteLookupResponse` containing only the quote IDs:
+
+```json
+{
+  "quotes": [
+    {
+      "pubkey": <str>,
+      "quote": <str>
+    }
```
thesimplekid: We would also need the payment method here so the client then knows what endpoint to use

callebtc:
what the pubkey isn't unique?

callebtc:
concept nACK 

I think users or services should remember the quote ID

PR #259, opened 2025-05-13
Bolt12 lighting offers payment method

thesimplekid:
Depends on https://github.com/cashubtc/nuts/pull/258

- [x] CDK https://github.com/cashubtc/cdk/pull/874
- [ ] Nutshell
- [x] cashu-ts https://github.com/cashubtc/cashu-ts/pull/342
- [ ] nutmix

PR #280, opened 2025-08-19
abstract NUT-17 subscription kinds

### LINKED ISSUE #279: Abstract NUT-17 subscription kinds to their respective NUTs
As new NUTs are added we may want to extend the supported websocket subscription kinds. Right now, [NUT-17](https://github.com/cashubtc/nuts/blob/main/17.md) defines "bolt11_mint_quote", "bolt11_melt_quote", and "proof_state".

Based on [this discussion](https://github.com/cashubtc/nuts/pull/259#issuecomment-3197864047) I would propose that we do not define the specific subscription kinds in NUT-17. We should instead abstract NUT-17 and define the bolt11 kinds in NUT-23 and the "proof_state" kind in NUT-07.  Then we can define bolt12 specific kinds for #259

gudnuf:
closes #279 

I made NUT-17 general so that it can easily be extended with new subscription kinds. For now I just defined the `proof_state` kind in NUT-07 and `bolt11_mint_quote`/`bolt11_melt_quote` in NUT-23. We will want the ability to add new kinds for things like new payment methods (ie. bolt12)


callebtc:
we can add bolt12 since #259 is now merged

17.md
```diff
@@ -2,17 +2,22 @@
...
 
 ## Subscriptions
 
-The websocket enables real-time subscriptions that wallets can use to receive notifications for a state change of a `MintQuoteResponse` ([NUT-04][04]), `MeltQuoteResponse` ([NUT-05][05]), `CheckStateResponse` ([NUT-07][07]).
+The websocket enables real-time subscriptions that wallets can use to receive notifications for a state change as defined in the corresponding NUT:
 
-A summary of the subscription flow is the following:
+- [NUT-07][07] for proof state subscriptions
+- [NUT-23][23] for bolt11 mint and melt quote subscriptions
+
```
thesimplekid, line 18: bolt12 nut25

17.md
```diff
@@ -251,7 +246,12 @@ Mints signal websocket support via [NUT-06][06] using the following setting:
 }
 ```
 
+<<<<<<< HEAD
 Here, `commands` is an array of the commands that the mint supports. A mint that supports all commands would return `["bolt11_mint_quote", "bolt11_melt_quote", "bolt12_mint_quote", "bolt12_melt_quote", "proof_state"]`. Supported commands are given for each method-unit pair.
+=======
+Here, `commands` is an array of the commands that the mint supports. Supported commands are given for each method-unit pair.
+
+> > > > > > > fb15af3 (abstract NUT-17 subscription kinds)
```
prusnak: This is a git conflict which should be resolved

thesimplekid: thanks

Issue #37, opened 2023-07-08
NUT-05: Melting tokens without blocking


callebtc:
Currently, calling `POST /melt` is blocking until either the Lightning invoice succeeds or fails. This is not necessary anymore because we now have a new `pending` state in the `POST /check` response (see here: https://github.com/cashubtc/nuts/pull/21).

`POST /melt` should have an option to immediately return. I'm unsure what the best way to spec this would be. Input is very much appreciated.

Options:
* `POST /melt?blocking=false` with a URL param
* `PostMeltRequest` with optional `blocking: false` field in json
* New endpoint a la `POST /meltnonblocking`

callebtc:
> Have you thought about using the HTTP Header `Prefer: respond-async` from [RFC7240](https://datatracker.ietf.org/doc/html/rfc7240#section-4.1)?

I didn't know about that, will add it as a fourth option :)

PR #283, opened 2025-08-25
onchain bitcoin payment method

gudnuf:
- [ ] CDK https://github.com/cashubtc/cdk/pull/995
- [ ] Nutshell
- [ ] cashu-ts https://github.com/cashubtc/cashu-ts/pull/349
- [ ] nutmix

### DIFF
--- a/26.md
+++ b/26.md
@@ -19,7 +19,7 @@ For the `onchain` method, the wallet includes the following specific `PostMintQu
 }
 ```
 
-> **Note:** While a pubkey is optional as per [NUT-20][20] for [NUT-04][04] it is required in this NUT and the mint **MUST NOT** issue a mint quote if one is not included.
+> **Note:** While a pubkey is optional as per [NUT-20][20] for [NUT-04][04], it is required in this NUT and the mint **MUST NOT** issue a mint quote if one is not included.
 
 > **Privacy:** To prevent linking multiple mint quotes together, wallets **SHOULD** generate a unique public key for each mint quote request.
 
--- a/26.md
+++ b/26.md
@@ -207,7 +207,7 @@ The mint responds with a `PostMeltQuoteOnchainResponse`:
   "fee_reserve": <int>,
   "state": <str_enum[STATE]>,
   "expiry": <int>,
-  "transaction_id": <str|null>
+  "txid:vout": <str|null>
 }
 ```
 
@@ -219,7 +219,7 @@ Where `fee_reserve` is the additional fee reserve required to broadcast the tran
 - `"PENDING"` means that the transaction has been broadcast but not yet confirmed.
 - `"PAID"` means that the transaction has been confirmed.
 
-`transaction_id` is the Bitcoin transaction ID, present once the transaction has been broadcast.
+`txid:vout` is the outpoint of the payment, present once the transaction has been broadcast.
 
 ### Melting Tokens
 
@@ -244,7 +244,7 @@ If the `outputs` field is included and there is excess from the `fee_reserve`, t
   "fee_reserve": <int>,
   "state": <str_enum[STATE]>,
   "expiry": <int>,
-  "transaction_id": <str>,
+  "txid:vout": <str>,
   "change": <Array[BlindSignature]> // Present if outputs were included and there's change
 }
 ```
@@ -296,7 +296,7 @@ curl -X POST https://mint.host:3338/v1/melt/onchain -d \
   "fee_reserve": 2000,
   "state": "PENDING",
   "expiry": 1701704757,
-  "transaction_id": "3b7f3b85c5f1a3c4d2b8e9f6a7c5d8e9f1a2b3c4d5e6f7a8b9c1d2e3f4a5b6c7"
+  "txid:vout": "3b7f3b85c5f1a3c4d2b8e9f6a7c5d8e9f1a2b3c4d5e6f7a8b9c1d2e3f4a5b6c7:2"
 }
 ```
 
@@ -311,7 +311,7 @@ curl -X POST https://mint.host:3338/v1/melt/onchain -d \
   "fee_reserve": 2000,
   "state": "PAID",
   "expiry": 1701704757,
-  "transaction_id": "3b7f3b85c5f1a3c4d2b8e9f6a7c5d8e9f1a2b3c4d5e6f7a8b9c1d2e3f4a5b6c7"
+  "txid:vout": "3b7f3b85c5f1a3c4d2b8e9f6a7c5d8e9f1a2b3c4d5e6f7a8b9c1d2e3f4a5b6c7:2"
 }
 ```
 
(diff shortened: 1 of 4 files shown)


26.md
```diff
@@ -0,0 +1,322 @@
...
+**Successful melt response** (transaction broadcast):
+
+```json
+{
+  "quote": "TRmjduhIsPxd...",
+  "request": "bc1qxy2kgdygjrsqtzq2n0yrf2493p83kkfjhx0wlh",
+  "amount": 100000,
+  "unit": "sat",
+  "fee_reserve": 2000,
+  "state": "PENDING",
```
gudnuf, line 302: should a melt be considered PAID as soon as the mint broadcasts the transaction, or should PENDING mean the transaction was broadcast but not yet mined?

Also, what happens if fees spike and the transaction does not confirm? Is it the mint or wallet's responsibility? 

It seems to me like the mint should pick a `fee_reserve` that it is comfortable with and then it is the mint's responsibility to make sure the transaction gets mined.

UPDATE: 

after discussions with tsk I made some changes to this nut and now PENDING means that the transaction is being processed by the mint. Processing can either be waiting for a batched transaction to be broadcast or while the tx is in the mempool. COMPLETED means the transaction has been mined into a block. 

On fees, the mint can return multiple melt quotes with absolute fees based on user's desired confirmation time. If fees spike the mint can choose to fail the melt quote and release the pending proofs, or the mint can dip into some fee reserve wallet to cover it's bad estimate

26.md
```diff
@@ -0,0 +1,322 @@
...
+For the `onchain` method, the wallet includes the following specific `PostMintQuoteOnchainRequest` data:
+
+```json
+{
+  "unit": <str_enum[UNIT]>,
+  "pubkey": <str>
+}
+```
+
+> **Note:** While a pubkey is optional as per [NUT-20][20] for [NUT-04][04] it is required in this NUT and the mint **MUST NOT** issue a mint quote if one is not included.
```
asmogo:
```suggestion
> **Note:** While a pubkey is optional as per [NUT-20][20] for [NUT-04][04], it is required in this NUT and the mint **MUST NOT** issue a mint quote if one is not included.
```

26.md
```diff
@@ -0,0 +1,322 @@
...
+}
+```
+
+Where:
+
+- `quote` is the quote ID
+- `request` is the Bitcoin address to send funds to
+- `expiry` is the Unix timestamp until which the mint quote is valid
+- `pubkey` is the public key from the request
+- `amount_paid` is the total confirmed amount paid to the mint via the onchain address
```
asmogo:
```suggestion
- `amount_confirmed` is the total confirmed amount paid to the mint via the onchain address
```

26.md
```diff
@@ -0,0 +1,322 @@
...
+```json
+{
+  "quote": <str>,
+  "request": <str>,
+  "unit": <str_enum[UNIT]>,
+  "expiry": <int|null>,
+  "pubkey": <str>,
+  "amount_paid": <int>,
+  "amount_issued": <int>,
+  "amount_unconfirmed": <int>
```
asmogo: The unconfirmed amount should be optional, since it could be omitted by certain PaymentBackend implementations. 

```suggestion
  "amount_unconfirmed": <int|null>
```

26.md
```diff
@@ -0,0 +1,322 @@
...
+The mint responds with a `PostMintQuoteOnchainResponse`:
+
+```json
+{
+  "quote": <str>,
+  "request": <str>,
+  "unit": <str_enum[UNIT]>,
+  "expiry": <int|null>,
+  "pubkey": <str>,
+  "amount_paid": <int>,
```
asmogo, line 35: Suggesting to rename `amount_paid` to `amount_confirmed`

```suggestion
  "amount_confirmed": <int>,
```

26.md
gudnuf: The nice thing about keeping this as `amount_paid` is its the same for bolt12 so the implementation is easier because for bolt12 and onchain quote you can compare `amount_paid` with `amount_issued`.

thesimplekid: > The nice thing about keeping this as amount_paid is its the same for bolt12 so the implementation is easier because for bolt12 and onchain quote you can compare amount_paid with amount_issued.

I would prefer to keep it for this reason as well.

26.md
```diff
@@ -0,0 +1,322 @@
...
+Check quote state:
+
+```bash
+curl -X GET http://localhost:3338/v1/melt/quote/onchain/TRmjduhIsPxd...
+```
+
+**Melt request**:
+
+```bash
+curl -X POST https://mint.host:3338/v1/melt/onchain -d \
```
thesimplekid, line 288: I think we need to revist this for onchain and we should just do it for all of them while were at it. https://github.com/cashubtc/nuts/issues/37. 

But for onchain specifically it make no sense to wait for the melt to confirm and request should return okay once it is received.

gudnuf: If we are going to say that a melt quote does not complete until the transaction is confirmed, then yeah. If a completed melt quote means the transaction was broadcast, then it seems fine to wait.

Was that the consensus that the quote stays pending until the transaction is confirmed?

thesimplekid: > If we are going to say that a melt quote does not complete until the transaction is confirmed, then yeah. If a completed > melt quote means the transaction was broadcast, then it seems fine to wait.

What if a mint wants to batch then it doesn't work for broadcast even.

26.md
```diff
@@ -0,0 +1,333 @@
...
+```json
+{
+  "quote": <str>,
+  "request": <str>,
+  "amount": <int>,
+  "unit": <str_enum[UNIT]>,
+  "fee_reserve": <int>,
+  "state": <str_enum[STATE]>,
+  "expiry": <int>,
+  "transaction_id": <str|null>
```
thesimplekid: This should be outpoint not just txid. `txid:vout`

26.md
```diff
@@ -244,7 +244,7 @@ If the `outputs` field is included and there is excess from the `fee_reserve`, t
   "fee_reserve": <int>,
   "state": <str_enum[STATE]>,
   "expiry": <int>,
-  "transaction_id": <str>,
+  "txid:vout": <str>,
```
thesimplekid:
```suggestion
  "outpoint: <str>,
```

gudnuf: oh duh, thanks I updated it

26.md
gudnuf: > I think we need to revist this for onchain and we should just do it for all of them while were at it. 

Is there a change you suggest I make here? In NUT-05 it says:

"For methods that involve external payments (like Lightning), this call may block until the payment either succeeds or fails. This can take a long time. Make sure to use no (or a very long) timeout when making this call!"

The way this nut is now seems fine as it shows that a "successful melt response" returns the quote as PENDING and NUT-05 is pretty open-ended in that it says, "this call may block".

thesimplekid:
close for https://github.com/cashubtc/nuts/pull/365

PR #286, opened 2025-09-11
feat: ohttp mint

thesimplekid:
image

### DIFF
--- a/26.md
+++ b/26.md
@@ -17,6 +17,8 @@ This NUT does not change any existing Cashu API semantics. It only defines disco
 - Gateway: The component that holds the OHTTP key config, decapsulates requests, forwards them to the mint backend, and encapsulates responses.
 - Relay: A forwarder that sees only encrypted OHTTP messages and forwards them to the gateway.
 
+For optimal privacy, the relay and mint SHOULD be operated by different entities that do not collude. The privacy guarantees of OHTTP depend on the relay and mint not working together to correlate user traffic patterns.
+
 ## Discovery via NUT‑06
 
 Mints advertise OHTTP support in the `nuts` section of the [NUT‑06][06] `GetInfoResponse`.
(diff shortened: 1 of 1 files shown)


26.md
```diff
@@ -0,0 +1,75 @@
...
+
+This NUT specifies how a Cashu mint can offer its HTTP API over Oblivious HTTP (OHTTP) to improve network‑level privacy for wallets. OHTTP lets a client send encrypted requests to an oblivious gateway (optionally via a relay). The gateway decrypts, forwards the plaintext HTTP request to the mint, and encrypts the response, preventing the mint from learning the client's network address and preventing the relay from seeing request contents.
+
+This NUT does not change any existing Cashu API semantics. It only defines discovery and endpoints needed to use OHTTP with a mint.
+
+## Terminology
+
+- OHTTP: Oblivious HTTP as specified by the IETF OHAI working group.
+- Gateway: The component that holds the OHTTP key config, decapsulates requests, forwards them to the mint backend, and encapsulates responses.
+- Relay: A forwarder that sees only encrypted OHTTP messages and forwards them to the gateway.
```
nothingmuch, line 17: probably worth mentioning that this should be operated independently of the mint, due to the non-collusion assumption required for privacy

26.md
```diff
@@ -0,0 +1,75 @@
...
+- `enabled` (bool): Set to `true` when the mint supports OHTTP.
+- `gateway_url` (str|null, optional): Base URL of the OHTTP gateway to use for key discovery (and, by convention, for the OHTTP gateway endpoint). If omitted or `null`, clients SHOULD assume the gateway is available on the mint's own origin.
+
+Wallets SHOULD treat `gateway_url` as a hint for where to fetch OHTTP keys. A mint MAY operate the gateway on the same origin as the mint API, in which case `gateway_url` can be omitted.
+
+## Gateway Endpoints
+
+A gateway MUST expose the following endpoints:
+
+- `GET /ohttp-keys` → returns the OHTTP key configuration
```
nothingmuch: RFC 9540 specifies that GET on `/.well-known/ohttp-gateway` returns the OHTTP config, the `/ohttp-keys` path was used in the BIP 77 directory before adding RFC 9540 support and is currently there for backwards compatibility only

26.md
```diff
@@ -0,0 +1,75 @@
...
+- `POST /.well-known/ohttp-gateway` → accepts encapsulated OHTTP requests
+  - Request content type: `message/ohttp-req`
+  - Response content type: `message/ohttp-res`
+  - Body: opaque binary as defined by OHTTP
+
+A deployment MAY route additional paths to the same handler (e.g., a catch‑all that accepts `message/ohttp-req`); clients MUST NOT rely on nonstandard paths.
+
+## Client Behavior
+
+- Key fetch: Wallets fetch the OHTTP key configuration from `gateway_url + "/ohttp-keys"`. If `gateway_url` is not provided, wallets SHOULD fetch from the mint origin at `"/ohttp-keys"`.
```
nothingmuch: RFC 9458 cites https://datatracker.ietf.org/doc/html/draft-ietf-privacypass-key-consistency-01 as general reccomendations for key consistency, probably worth linking here.

A mechanism analogous to certificate transparency logs probably makes sense as a future extension, both for OHTTP keys and for the token public keys, I'm happy to share some ideas, and i'm also curious what strategies existing cashu clients use for token key consistency

26.md
```diff
@@ -0,0 +1,75 @@
...
+- Encapsulation: Wallets construct a BHTTP request that mirrors a normal Cashu API call (method, scheme, authority, path, headers, and body), encapsulate it with the fetched key config, and send it to the OHTTP gateway endpoint at `"/.well-known/ohttp-gateway"` through a relay.
```
nothingmuch: BHTTP not yet defined, citation link would be https://www.rfc-editor.org/info/rfc9292

nothingmuch: in BIP 77 we require padding the encapsulated BHTTP ciphertext to 8192 bytes to resist traffic analysis. Perhaps it makes sense to make a similar recommendation for cashu. If the two user bases share the same set of OHTTP relays and the OHTTP gateway does not reveal which protocol the client is using, it would be beneficial if this value was shared by both protocols.

26.md
```diff
@@ -0,0 +1,75 @@
...
+- Headers: Wallets include the same application‑level headers they would send directly to the mint (e.g., auth headers if applicable). The gateway forwards headers to the mint backend.
+- Paths: The BHTTP request path is the regular Cashu API path (e.g., `/v1/mint/quote/bolt11`, `/v1/melt/bolt11`, `/v1/swap`, etc.). The gateway forwards the request to the mint backend preserving method, path, headers, and body.
+
+Wallets MUST send OHTTP messages through a relay to the gateway.
+
+## Errors
+
+Gateways SHOULD respond with appropriate HTTP status codes for transport‑level errors. Since request/response bodies are OHTTP‑encapsulated, application‑level Cashu errors remain unchanged and are handled by the wallet after decapsulation.
```
nothingmuch: I believe this should be:

```suggestion
Relays SHOULD respond with appropriate HTTP status codes for transport‑level errors. Since request/response bodies are OHTTP‑encapsulated, application‑level Cashu errors remain unchanged and are handled by the wallet after decapsulation.
```

unless this refers to transport errors between the gateway and the mint, but in the context of the next sentence that doesn't seem like the intention

26.md
thesimplekid: We use the keyset id. The wallet verified the keyset id is verified by the wallet before being used https://github.com/cashubtc/nuts/blob/main/02.md#deriving-the-keyset-id

26.md
a1denvalu3: This is a nice idea.

PR #287, opened 2025-09-14
clarify that `secret` refers to `x`

aaronmcdaid:
Explicitly state that `secret` refers to the `x` described earlier in this doc. And also, that `secret` is a hexadecimal string


thesimplekid:
The secret is not always hex for example when p2pk and htlc secrets are used.

PR #291, opened 2025-10-02
NUT-XX:Pay-to-Blinded-Key (P2BK): 

lollerfirst:
formalize public key blinding atop NUT-11;
add spec and README entry;
ensure proper Markdown formatting

### DIFF
--- a/26.md
+++ b/26.md
@@ -1,0 +1,183 @@
+# NUT-26: Pay-to-Blinded-Key (P2BK)
+
+`optional`
+
+`depends on: NUT-11`
+
+---
+
+## Summary
+
+This NUT defines Pay-to-Blinded-Key (P2BK), a wallet-to-wallet privacy extension of [NUT-11][11] (P2PK). P2BK blinds the public keys contained in a P2PK Secret so the mint cannot learn the true recipient key material when issuing or redeeming ecash. The scheme requires no changes to mint behavior: wallets MUST normalize P2BK Secrets to standard P2PK Secrets before sending requests to a mint. All Schnorr signing and verification at the mint remains exactly as specified in NUT-11.
+
+## Motivation
+
+- In NUT-11, the receiver public key(s) appear in cleartext in the Secret. On redemption, the mint learns the long-term public key(s) that control the ecash.
+- P2BK hides those keys from the mint by blinding each receiver public key P with a random scalar r: P' = P + r·G. The receiver later derives the corresponding private key k = p + r (mod n) to sign per NUT-11. The mint only ever sees normal P2PK Secrets and signatures.
+
+## Definitions and notation
+
+- Curve: secp256k1 with base point G and order n.
+- Unblinded pubkey (compressed hex): P.
+- Blinding scalar: r with 1 ≤ r ≤ n − 1.
+- Blinded pubkey: P' = P + r·G (compressed hex).
+- Derived private key: k = (p + r) mod n.
+
+## Wire format
+
+Secret kind: "P2BK"
+
+P2BK introduces a new well-known Secret kind. It has the same structure as a NUT-10/11 P2PK Secret except for the kind string and an additional tag that carries the blinding scalars:
+
+```json
+[
+  "P2BK",
+  {
+    "nonce": "<hex32>",
+    "data": "<P0'>",
+    "tags": [
+      ["sigflag", "SIG_INPUTS"],
+      ["pubkeys", "<P1'>", "<P2'>"],
+      ["n_sigs", "<int>"],
+      ["locktime", "<unix_seconds>"],
+      ["refund", "<R0'>", "<R1'>"],
+      ["n_sigs_refund", "<int>"],
+      ["r", "<r0>", "<r1>", "<r2>"]
+    ]
+  }
+]
+```
+
+Notes:
+
+- data holds the primary blinded pubkey P0'.
+- pubkeys holds zero or more additional blinded pubkeys P1', P2', … for locktime multisig.
+- refund holds zero or more blinded refund pubkeys R0', R1', … for refund multisig.
+- r is an array of the blinding scalars as fixed-length 32-byte hex strings (lowercase) without prefix. The order of r MUST match the concatenation order of all blinded keys present in the Secret:
+  1) the primary lock key data, then 2) any keys in pubkeys, then 3) any keys in refund.
+  For example, if there are m lock keys in total (1 in data + pubkeys.length) and t refund keys, the r tag contains m + t entries: [r0, r1, …, r(m−1), r(m), …, r(m+t−1)].
+
+## Normalization for mints (wallet MUST)
+
+Mints do not understand "P2BK". Wallets MUST transform a P2BK Secret into a standard P2PK Secret before using it in any request to a mint (swap/melt/etc.). Normalization is purely a string transformation:
+
+1. Parse the JSON Secret [kind, meta]. If kind != "P2BK", return unchanged.
+2. Remove the r tag from meta.tags if present.
+3. Replace kind with "P2PK".
+4. Re-serialize to JSON string. This string is the message to be signed for NUT-11 witnesses and is what wallets MUST send to mints.
+
+No other fields are changed. In particular, data, pubkeys, refund remain the blinded pubkeys P', R'.
+
+## Signing and verification (unchanged at the mint)
+
+- Signature scheme, sigflag semantics, multisig, locktime/refund behavior, and message aggregation (SIG_INPUTS, SIG_ALL) are as in NUT-11. All signatures are Schnorr signatures over the SHA256 hash of the normalized P2PK Secret string (and any aggregated outputs per NUT-11).
+- Mints MUST NOT receive r values and MUST NOT receive Secrets of kind "P2BK". Wallets MUST perform normalization before sending any request to a mint.
+
+## Wallet behavior
+
+### Sender (building P2BK)
+
+- For each recipient lock/refund pubkey Pi, choose fresh ri ∈ [1, n − 1].
+- Compute Pi' = Pi + ri·G.
+- Construct a P2PK Secret as in NUT-11 but replace all pubkeys with their blinded forms. Blind the message as usual to obtain outputs and promises from the mint.
+- After unblinding signatures to proofs, rewrite the proof Secret to kind "P2BK" and append a tag ["r", r0, r1, …] carrying all ris in the order defined above. Deliver these P2BK proofs to the receiver.
+
+### Receiver (spending P2BK)
+
+- Parse the P2BK Secret and extract r values.
+- Derive the signing private keys ki = (pi + ri) mod n for each of its keys pi that are expected to sign under NUT-11 rules (respecting locktime and multisig tags).
+- Normalize the Secret to P2PK (strip r; set kind = "P2PK").
+- Produce Schnorr signature(s) over the normalized P2PK Secret string according to the applicable sigflag and multisig rules.
+- Submit the proofs to the mint. The mint verifies exactly as in NUT-11.
+
+## Determinism and canonicalization
+
+- The r tag MUST contain lowercase hex and be 64 hex chars per entry (32 bytes), matching the conventional encoding used for scalars elsewhere in Cashu.
+- Wallets SHOULD ignore duplicate r entries and MUST deduplicate derived signing keys if the same ri appears multiple times. Ordering rules still apply to allow position-based mapping.
+- Wallets MUST NOT include an empty r tag; if present, mints will never see it due to normalization, but receivers SHOULD treat an empty r tag as equivalent to absence (no blinding information).
+
+## Constraints and errors
+
+- ri MUST satisfy 1 ≤ ri ≤ n − 1. Senders MUST use uniformly random ri and MUST NOT reuse ri across different blinded keys.
+- The derived private key ki = (pi + ri) mod n MUST be non-zero. If ki = 0 (probability ~1/n), the receiver cannot sign; the remedy is to request a re-send with different randomness.
+- Wallets MUST treat any P2BK Secret missing the r tag as a non-blinded P2PK Secret (i.e., treat like NUT-11 where P' are just arbitrary pubkeys).
+
+## Security considerations
+
+- Mint Privacy: Mints never see r nor unblinded P. They see only blinded pubkeys P' inside a standard P2PK Secret, and standard Schnorr signatures. This prevents mints from trivially linking P' to a receiver's long-term P if P is reused elsewhere.
+- Freshness: Reusing `r`s across payments or across keys enables linkability across P' values. Senders MUST generate fresh `r`s for every blinded key instance.
+- Receiver Key Hygiene: Receivers SHOULD avoid reusing the same base key p across many P2BK payments if linkability via off-chain correlation is a concern, although the mint cannot derive P.
+- Metadata Leakage: The presence of P' alone does not reveal P without knowledge of r or p. The r tag is only exchanged wallet-to-wallet and MUST NOT be forwarded to mints or third parties.
+
+## Compatibility
+
+- Backwards Compatibility: Mints require no changes. Wallets that do not understand P2BK will fail to parse the Secret kind and thus cannot accept such proofs. Senders SHOULD only send P2BK proofs to receivers known to support NUT-26.
+- Interop with NUT-11: All NUT-11 tag semantics (sigflag, n_sigs, locktime, refund, n_sigs_refund) apply unchanged when used with blinded keys.
+
+## Mint info
+
+No new mint capability flag is required because mint behavior is unchanged. [NUT-06][06] remains the source of P2PK support (NUT-11). Wallet capability advertisement is out-of-scope for this NUT.
+
+## Worked example
+
+Receiver base pubkey P (compressed):
+
+```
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -18,10 +18,6 @@ Wallets and mints `MUST` implement all mandatory specs and `CAN` implement optio
 | [05][05] | Melting tokens          |
 | [06][06] | Mint info               |
 
-### Optional
```
prusnak, line 21: Is this removal intentional? It breaks table rendering

robwoodgate (changes requested):
Tags entry must be included to ensure normalization is consistent

26.md
```diff
@@ -0,0 +1,183 @@
...
+- r is an array of the blinding scalars as fixed-length 32-byte hex strings (lowercase) without prefix. The order of r MUST match the concatenation order of all blinded keys present in the Secret:
+  1) the primary lock key data, then 2) any keys in pubkeys, then 3) any keys in refund.
+  For example, if there are m lock keys in total (1 in data + pubkeys.length) and t refund keys, the r tag contains m + t entries: [r0, r1, …, r(m−1), r(m), …, r(m+t−1)].
+
+## Normalization for mints (wallet MUST)
+
+Mints do not understand "P2BK". Wallets MUST transform a P2BK Secret into a standard P2PK Secret before using it in any request to a mint (swap/melt/etc.). Normalization is purely a string transformation:
+
+1. Parse the JSON Secret [kind, meta]. If kind != "P2BK", return unchanged.
+2. Remove the r tag from meta.tags if present.
```
robwoodgate:
```suggestion
2. Remove the r tag from meta.tags if present. If no tags remain, tags MUST be kept as an empty array (i.e: `tags: []`_
```

26.md
```diff
@@ -0,0 +1,183 @@
...
+- Signature scheme, sigflag semantics, multisig, locktime/refund behavior, and message aggregation (SIG_INPUTS, SIG_ALL) are as in NUT-11. All signatures are Schnorr signatures over the SHA256 hash of the normalized P2PK Secret string (and any aggregated outputs per NUT-11).
+- Mints MUST NOT receive r values and MUST NOT receive Secrets of kind "P2BK". Wallets MUST perform normalization before sending any request to a mint.
+
+## Wallet behavior
+
+### Sender (building P2BK)
+
+- For each recipient lock/refund pubkey Pi, choose fresh ri ∈ [1, n − 1].
+- Compute Pi' = Pi + ri·G.
+- Construct a P2PK Secret as in NUT-11 but replace all pubkeys with their blinded forms. Blind the message as usual to obtain outputs and promises from the mint.
```
robwoodgate:
```suggestion
- Construct a P2PK Secret as in NUT-11 but replace all pubkeys with their blinded forms. The `tags` entry MUST be present, even if empty (i.e: `tags: []`). Blind the message as usual to obtain outputs and promises from the mint.
```

26.md
```diff
@@ -0,0 +1,183 @@
...
+- After unblinding signatures to proofs, rewrite the proof Secret to kind "P2BK" and append a tag ["r", r0, r1, …] carrying all ris in the order defined above. Deliver these P2BK proofs to the receiver.
+
+### Receiver (spending P2BK)
+
+- Parse the P2BK Secret and extract r values.
+- Derive the signing private keys ki = (pi + ri) mod n for each of its keys pi that are expected to sign under NUT-11 rules (respecting locktime and multisig tags).
+- Normalize the Secret to P2PK (strip r; set kind = "P2PK").
```
robwoodgate:
```suggestion
- Normalize the Secret to P2PK (strip r; set kind = "P2PK"; ensure `tags` entry is retained if empty).
```

26.md
```diff
@@ -0,0 +1,183 @@
...
+
+- Mint Privacy: Mints never see r nor unblinded P. They see only blinded pubkeys P' inside a standard P2PK Secret, and standard Schnorr signatures. This prevents mints from trivially linking P' to a receiver's long-term P if P is reused elsewhere.
+- Freshness: Reusing `r`s across payments or across keys enables linkability across P' values. Senders MUST generate fresh `r`s for every blinded key instance.
+- Receiver Key Hygiene: Receivers SHOULD avoid reusing the same base key p across many P2BK payments if linkability via off-chain correlation is a concern, although the mint cannot derive P.
+- Metadata Leakage: The presence of P' alone does not reveal P without knowledge of r or p. The r tag is only exchanged wallet-to-wallet and MUST NOT be forwarded to mints or third parties.
+
+## Compatibility
+
+- Backwards Compatibility: Mints require no changes. Wallets that do not understand P2BK will fail to parse the Secret kind and thus cannot accept such proofs. Senders SHOULD only send P2BK proofs to receivers known to support NUT-26.
+- Interop with NUT-11: All NUT-11 tag semantics (sigflag, n_sigs, locktime, refund, n_sigs_refund) apply unchanged when used with blinded keys.
```
robwoodgate:
```suggestion
- Interop with NUT-11: All NUT-11 tag semantics (sigflag, n_sigs, locktime, refund, n_sigs_refund) apply unchanged when used with blinded keys. NUT-11 secrets intended for P2BK blinding MUST include the `tags` entry, even if empty (ie: `tags: []`)
```

robwoodgate (changes requested):

robwoodgate:
The reference implementation for cashu-ts is currently [in the v3-hackday branch of my own repo](https://github.com/robwoodgate/cashu-ts/tree/v3-hackday).

It is build on top of v3-alpha-2, so I'll PR it to v3-alpha-dev branch once we catch it up to my current v3 tip.

README.md
robwoodgate: Well spotted 👍

lollerfirst:
> So when deriving the blinded secret key for a blinded pubkey P', we need to check both candidates, (p + >r) mod n and (-p + r) mod n, and select the one whose derived public key matches P' in the proof.

The Mint also verifies with BIP340 which I think it just always chooses the even Y:
```
Implicit Y coordinates In order to support efficient verification and batch verification, the Y coordinate of P and of R cannot be ambiguous (every valid X coordinate has two possible Y coordinates). We have a choice between several options for symmetry breaking:

    Implicitly choosing the Y coordinate that is in the lower half.
    Implicitly choosing the Y coordinate that is even[[6](https://bips.xyz/340#cite-note-6)].
    Implicitly choosing the Y coordinate that is a quadratic residue (i.e. has a square root modulo p).

The second option offers the greatest compatibility with existing key generation systems, where the standard 33-byte compressed public key format consists of a byte indicating the oddness of the Y coordinate, plus the full X coordinate. To avoid gratuitous incompatibilities, we pick that option for P, and thus our X-only public keys become equivalent to a compressed public key that is the X-only key prefixed by the byte 0x02. For consistency, the same is done for R[[7](https://bips.xyz/340#cite-note-7)].
```

So the actual public key is always even.

prusnak:
> Why didn't this problem emerge in the tests?

just speculating: maybe the tests do not include this case? if that's true, we should add a case that triggers the need for this behavior

robwoodgate:
> @robwoodgate
> 
> > So when deriving the blinded secret key for a blinded pubkey P', we need to check both candidates, (p + >r) mod n and (-p + r) mod n, and select the one whose derived public key matches P' in the proof.
> 
> Why didn't this problem emerge in the tests?

It did. I added tests for Schnorr derived keys when I found out P2BK wasn't consistently working with Nostr secret keys. Then I fixed it :)

robwoodgate:
> @robwoodgate
> 
> > So when deriving the blinded secret key for a blinded pubkey P', we need to check both candidates, (p + >r) mod n and (-p + r) mod n, and select the one whose derived public key matches P' in the proof.
> 
> Why didn't this problem emerge in the tests?

And just to be clear, there was never an issue signing regular Schnorr pubkey locked  proofs. It was just for P2BK where we have to derive an sk for a blinded Schnorr derived pubkey.

26.md
```diff
@@ -0,0 +1,190 @@
...
+
+- In NUT-11, the receiver public key(s) appear in cleartext in the Secret. On redemption, the mint learns the long-term public key(s) that control the ecash.
+- P2BK hides those keys from the mint by blinding each receiver public key P with a random scalar r: P' = P + r·G. The receiver later derives a corresponding private key k from p and r (see below) to sign per NUT-11. The mint only ever sees normal P2PK Secrets and signatures.
+
+## Definitions and notation
+
+- Curve: secp256k1 with base point G and order n.
+- Unblinded pubkey: P.
+  - For SEC1-style keys, P is a 33-byte compressed SEC1 pubkey (hex).
+  - For BIP340 Schnorr/x-only contexts (e.g., Nostr), P is an x-only 32-byte pubkey (hex) with even-Y parity implied by BIP340.
```
robwoodgate: We should probably stay consistent with Cashu ecosystem and say '02' prefixed.

```suggestion
  - For BIP340 Schnorr/x-only contexts (e.g., Nostr), P is an x-only 32-byte pubkey (hex) with even-Y parity (per BIP340), prefixed with '02' for Cashu exosystem interoperability.
```

26.md
```diff
@@ -0,0 +1,190 @@
...
+- Blinding scalar: r with 1 ≤ r ≤ n − 1.
+- Blinded pubkey: P' = P + r·G. Encoding matches P: compressed SEC1 (33-byte hex) for SEC1 keys; x-only (32-byte hex, even-Y) for BIP340 contexts.
```
robwoodgate: Pubkeys appearing in secrets should always be compressed SEC1 (33-byte hex) format.

```suggestion
- Blinded pubkey: P' = P + r·G. Encoding is compressed SEC1 (33-byte hex).
```

robwoodgate (approved):
Thanks for merging my nit suggestions - sorry, I spelt ecosystem wrong in L23
Aside from my gaff - ACK @ e38faa0f8690bd8c7b6f35b2ab72605dfc8963ae

robwoodgate:
I've added Cashu-TS with P2BK support (`v3-hackday-f3b9a9e`) to [Cashu NutLock](https://www.nostrly.com/cashu-nutlock/), so we can play with it in the field. Cashu Redeem and Witness are also P2BK aware now.

TS files for these tools [available here](https://github.com/robwoodgate/nostrly/tree/main/src/js).

26.md
```diff
@@ -0,0 +1,158 @@
...
+
+This example demonstrates that the `pr` entries preserve the blinding scalars for both the primary lock key and the refund key in order.
+
+## Wallet workflow
+
+### Sender (building P2BK proofs)
+
+1. For each receiver key $P_i$ (lock and refund) draw fresh randomness $r_i$
+2. Compute $P'_i = P_i + r_i \cdot G$.
+3. Build the P2PK secret as in NUT-11, substituting all pubkeys with the blinded versions $P'_i$. Keep the `tags` array present even if empty.
```
robwoodgate: Tags if empty was only vital when we were mutating the secret. It can be optional now.

```suggestion
3. Build the P2PK secret as in NUT-11, substituting all pubkeys with the blinded versions $P'_i$.
```

26.md
```diff
@@ -0,0 +1,158 @@
...
+1. Parse the canonical P2PK secret embedded in the proof.
+2. Read the accompanying `p2pk_r` scalars. Wallets SHOULD treat missing metadata as a regular P2PK proof.
+3. Derive signing keys according to key type (SEC1 or BIP340/x-only, see above).
+4. Produce Schnorr signatures over the unchanged P2PK secret per NUT-11 rules (`sigflag`, multisig, refund, etc.).
+5. Submit the proofs to the mint. Verification is identical to NUT-11 and requires no mint changes.
+
+## Determinism and canonicalisation
+
+- Scalars MUST be formatted as lowercase 64-character hex strings (32 bytes).
+- Wallets MUST preserve ordering and SHOULD de-duplicate derived signing keys if the same scalar appears multiple times.
```
robwoodgate: i don't really understand this requirement. I think it should probably be something like:

```suggestion
- Wallets MUST preserve ordering and SHOULD de-duplicate derived private keys that can sign the proof.
```

26.md
```diff
@@ -0,0 +1,158 @@
...
+
+- In NUT-11, the receiver pubkey(s) appear in clear text. When the receiver redeems the proof the mint learns the long-term key material.
+- P2BK hides those keys from the mint by blinding each receiver public key $P$ with a scalar $r$: $P' = P + r \cdot G$. The receiver later reconstructs the signing key from $p$ and $r$ (see below) while presenting the mint with an ordinary P2PK secret.
+
+## Definitions and notation
+
+- Curve: secp256k1 with base point $G$ and order $n$.
+- Unblinded pubkey: $P$.
+  - SEC1 contexts: 33-byte compressed SEC1 pubkey (hex).
+  - BIP340 Schnorr/x-only contexts (e.g. Nostr): 32-byte x-only pubkey (hex) with even-Y parity (per BIP340). For Cashu interoperability wallets SHOULD prefix the x-only key with `02` when serialising to hex strings.
```
robwoodgate: SHOULD -> MUST?

```suggestion
  - BIP340 Schnorr/x-only contexts (e.g. Nostr): 32-byte x-only pubkey (hex) with even-Y parity (per BIP340). For Cashu interoperability wallets MUST prefix the x-only key with `02` when serialising to hex strings.
```

26.md
```diff
@@ -0,0 +1,158 @@
...
+## Determinism and canonicalisation
+
+- Scalars MUST be formatted as lowercase 64-character hex strings (32 bytes).
+- Wallets MUST preserve ordering and SHOULD de-duplicate derived signing keys if the same scalar appears multiple times.
+- Omit the field entirely rather than emitting an empty `p2pk_r`/`pr` array.
+
+## Constraints and errors
+
+- Each $r_i$ MUST satisfy $1 \leq r_i \leq n-1$; randomness MUST be uniform and MUST NOT be reused across blinded keys.
+- Wallets SHOULD reject proofs that advertise `p2pk_r` but do not contain the corresponding blinded keys.
```
robwoodgate:
```suggestion
- Wallets SHOULD reject proofs that advertise `p2pk_r` but do not contain the corresponding blinded keys. There must be one entry in `p2pk_r` for every blinded key.
```

26.md
```diff
@@ -0,0 +1,158 @@
...
+
+## Security considerations
+
+- Mint privacy: mints never learn the blinding scalars or the original pubkeys; spends appear identical to standard P2PK.
+- Freshness: reusing scalars enables linkage across blinded keys. Always use fresh randomness per key.
+- Metadata leakage: wallets MUST keep `p2pk_r` private between sender and receiver (do not leak to the Mint).
+
+## Compatibility
+
+- Backwards compatibility: no mint changes are required. Legacy wallets can't redeem P2BK locked ecash, because they don't know how to produce the right key for the signature.
```
robwoodgate:
```suggestion
- Backwards compatibility: no mint changes are required. Legacy wallets can't redeem P2BK locked ecash, because they don't know how to derive the right private key(s) for the signature.
```

26.md
a1denvalu3: Ah probably can be removed altogether. Clanker sometimes likes to be verbose for no reason.

robwoodgate: Cashu-TS de-duplicates the derived secret keys, but it's not vital. Just saves some signing attempts.
Happy either way.

The important bit is "Wallets MUST preserve ordering of the p2pk_r array"

26.md
```diff
@@ -0,0 +1,158 @@
...
+
+- Scalars MUST be formatted as lowercase 64-character hex strings (32 bytes).
+- Wallets MUST preserve ordering and SHOULD de-duplicate derived signing keys if the same scalar appears multiple times.
+- Omit the field entirely rather than emitting an empty `p2pk_r`/`pr` array.
+
+## Constraints and errors
+
+- Each $r_i$ MUST satisfy $1 \leq r_i \leq n-1$; randomness MUST be uniform and MUST NOT be reused across blinded keys.
+- Wallets SHOULD reject proofs that advertise `p2pk_r` but do not contain the corresponding blinded keys. There must be one entry in `p2pk_r` for every blinded key.
+
```
robwoodgate, line 188:
```suggestion
- Validate P, and later P′, as valid secp256k1 points and not the point at infinity before use.

```

26.md
```diff
@@ -0,0 +1,158 @@
...
+## Security considerations
+
+- Mint privacy: mints never learn the blinding scalars or the original pubkeys; spends appear identical to standard P2PK.
+- Freshness: reusing scalars enables linkage across blinded keys. Always use fresh randomness per key.
+- Metadata leakage: wallets MUST keep `p2pk_r` private between sender and receiver (do not leak to the Mint).
```
robwoodgate, line 193:
```suggestion
- Metadata leakage: wallets MUST keep `p2pk_r` private between sender and receiver (do not leak to the Mint).
- Where a P2BK proof is published in public (e.g: in a NutZap), the Mint _and any observer_ can immediately recover the clear public key $P$ from $P′$ and $r$. This removes the privacy benefits of P2BK, but does not compromise the security of the signing secret key ($p$).

```

26.md
a1denvalu3: I think this is redundant, as the secp library already does this.

26.md
a1denvalu3: This follows logically from the previous point.

lollerfirst:
Closing in favor of [ECDH derived blinding factors](https://github.com/cashubtc/nuts/pull/300)

PR #293, opened 2025-10-10
NUT-XX: Mining Share Payment Method

vnprc:
ehash spec

does not include melting because i haven't written it yet

### DIFF
--- a/XX.md
+++ b/XX.md
@@ -1,0 +1,189 @@
+# NUT-XX: Mining Share Payment Method
+
+`optional`
+
+`depends on: NUT-04 NUT-05 NUT-20`
+
+---
+
+This document describes minting ehash with the `mining_share` payment method. Mining-share quotes represent proof-of-work shares submitted to a mining pool affiliated with the mint. A share is identified by its block header hash and its quote becomes eligible for ehash issuance once the pool validates the share; the resulting ehash token is the authoritative record of the mining share credit until it is redeemed for a mining reward.
+
+## Definitions
+
+- **Ehash**: Ecash issued by a mint to represent the value of a valid mining share. Ehash denominations correspond to powers of two derived from share difficulty.
+- **Share difficulty**: The number of leading zero bits in the 256-bit `header_hash` supplied with a mining-share quote request.
+- **Keyset minimum difficulty**: The minimum number of leading zero bits a mint requires for shares associated with a given keyset and currency unit. This value is configured and advertised by the mint.
+
+All difficulty comparisons in this specification use these definitions.
+
+## Mint Quote
+
+Wallets request a mining-share quote with `POST /v1/mint/quote/mining_share` and the following `PostMintQuoteMiningShareRequest` payload:
+
+```json
+{
+  "amount": <int>,
+  "unit": <str_enum[UNIT]>,
+  "header_hash": <str>,
+  "description": <str|null>,
+  "pubkey": <str>
+}
+```
+
+- `amount` **MUST** be greater than zero and **MUST** equal `2^(share difficulty - keyset minimum difficulty)`, as defined above. If the share difficulty is lower than the keyset minimum difficulty the mint **MUST** reject the quote.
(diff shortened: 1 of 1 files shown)


XX.md
```diff
@@ -0,0 +1,189 @@
...
+  "keyset_id": <str>
+}
+```
+
+Where:
+
+- `quote` is the opaque quote identifier used by subsequent requests.
+- `request` echoes the share `header_hash` supplied in the request.
+- `state` is one of `"UNPAID"`, `"PAID"`, or `"ISSUED"`.
+- `expiry` is the Unix timestamp (in seconds) until which the share credit remains valid.
```
vnprc, line 59: Share credits should remain valid until some time after the mining reward is received. The mint cannot know the expiry time until after the mining reward is received. For ehash, most quotes are expected to be issued before this time. For use cases that do not involve ehash (mining share -> quote -> mining reward -> issue bitcoin-backed ecash) the quote will remain `UNPAID` until the mining reward is found, at which time it will change to `PAID` and the expiry will be set.

vnprc: I haven't thoroughly explored the latter use case and don't have an implementation for it. Should I try to write this spec to solve for both use cases or use a separate specification? Or maybe just spec out ehash for now and punt on this question.

XX.md
```diff
@@ -0,0 +1,189 @@
...
+        "quote": "621ea695-624b-4e53-9643-5e36fc1f8004",
+        "outputs": [
+          {
+            "amount": 256,
+            "id": "009a1f293253e41e",
+            "keyset_id": "01e2f8...",
+            "B_": "035015e6d7ade60ba8426cefaf1832bbd27257636e44a76b922d78e79b47cb689d"
+          },
+          {
+            "amount": 256,
```
vnprc, line 133: Why are there two proofs for the same amount? Seems like an invalid example.

XX.md
```diff
@@ -0,0 +1,189 @@
...
+
+Wallets **MUST** construct mint requests whose summed output amount matches this exponential value exactly. Once the mint marks the quote `"ISSUED"`, wallets **SHOULD** treat it as fully spent and avoid additional mint requests for that quote.
+
+### Example
+
+```bash
+curl -X POST https://mint.host/v1/mint/mining_share \
+  -H "Content-Type: application/json" \
+  -d '{
+        "quote": "621ea695-624b-4e53-9643-5e36fc1f8004",
```
vnprc, line 124: cdk uses UUIDs but they are not required in the spec. Should probably change this to use an ellipsis like other spec examples.

XX.md
```diff
@@ -0,0 +1,189 @@
...
+}
+```
+
+`min_amount` and `max_amount` encode the acceptable difficulty window for share hashes. Each value **MUST** be between `0` and `255`, inclusive, and represents the count of leading zero bits in the 256-bit header hash (difficulty `0` means no leading zero bits; difficulty `255` means all but the final bit are zero). Shares with fewer leading zeros than `min_amount` **MUST** be rejected. Shares that exceed `max_amount` **MUST** be accepted but paid out only up to the reward implied by `max_amount`. Mints **SHOULD** publish windows that align with their pool's payout policy.
+
+## Security Considerations
+
+- Mints **MUST** treat `header_hash` values as unique payment identifiers and **MUST** reject duplicate shares.
+- Wallets **SHOULD** generate a fresh NUT-20 key pair per mining-share quote to avoid linking multiple quotes.
+- Wallets **SHOULD** redeem mining-share quotes before they expire and **SHOULD** discard any quote that remains in the `"UNPAID"` state once its `expiry` elapses, as the mint will reject further minting attempts.
```
vnprc, line 184: Still some open questions around expiry. This comment seems incorrect in light of the intended (but not yet implemented) use of the expiry field.  See [this comment](https://github.com/cashubtc/nuts/pull/293/files#r2424539375)

thesimplekid:
Closing this since it is incomplete and inactive. Cdk supports custom units now so methods like this can be tested without modifying cdk or requiting them to be added to spec. I think it would be good to encourage new methods to first be added in that way to prove some level of market fit before adding them to spec.

PR #294, opened 2025-10-13
feat: Add NUT-XX Payment Request Bech32m Encoding specification

thesimplekid:
Add comprehensive specification for encoding Cashu payment requests using Bech32m format with TLV (Tag-Length-Value) serialization as an alternative to the existing CBOR+base64 format defined in NUT-18.

- New specification file: 26.md defining NUT-26
- Updated README.md to include NUT-26 in the specification table

The new Bech32m encoding format addresses several limitations of the legacy creqA (CBOR+base64) format:

1. **QR Code Efficiency**: Bech32m uppercase encoding is alphanumeric-mode compatible
2. **Better Error Detection**: Built-in Bech32m checksums provide data integrity validation
3. **Improved Human Readability**: Clear `creqb1` prefix identifies the format
4. **Standards Alignment**: Consistent with Bitcoin addresses (BIP-173/350) and Lightning invoices

- TLV-based serialization for flexible extensibility
- Support for all NUT-18 payment request fields (id, amount, unit, mints, description, transports, NUT-10 conditions)
- Compact encoding for common values (e.g., 'sat' as 0x00)
- Nested TLV structures for complex types (transports, spending conditions)
- Forward compatibility through unknown tag handling
- Backward compatibility with legacy creqA format detection
- Case-insensitive decoding with uppercase recommended for QR codes

The specification is marked as optional and depends on NUT-18.

- [ ] update tests based on review changes 

---

- [x] CDK https://github.com/cashubtc/cdk/pull/1190
- [ ] Nutshell 
- [x] cashu-ts https://github.com/cashubtc/cashu-ts/pull/436

### DIFF
--- a/26.md
+++ b/26.md
@@ -6,15 +6,6 @@
 
 This specification defines an alternative encoding format for Payment Requests using Bech32m encoding with TLV (Tag-Length-Value) serialization. This format provides better QR code compatibility and typically 30-60% size reduction compared to the CBOR+base64 encoding defined in NUT-18.
 
-## Motivation
-
-This encoding format addresses several limitations of the legacy `creqA` format:
-
-- **QR Code Efficiency**: Bech32m encoded as uppercase is alphanumeric-mode compatible, resulting in denser QR codes
-- **Human Readability**: The `creqb1` prefix clearly identifies the format
-- **Error Detection**: Bech32m includes built-in checksums for data integrity
-- **Standards Alignment**: Uses the same encoding family as Bitcoin addresses and Lightning invoices
-
 ## Encoded Request Format
 
 Payment requests are serialized using TLV encoding, then encoded with Bech32m:
@@ -23,28 +14,14 @@ Payment requests are serialized using TLV encoding, then encoded with Bech32m:
 
 The human-readable part (HRP) is `"creqb"` and the version separator is `"1"`. The data payload is TLV-encoded as described below, then encoded with Bech32m (not standard Bech32).
 
-> [!NOTE] > **Case Sensitivity and QR Code Optimization**
->
-> The encoding SHOULD be output in uppercase for optimal QR code compatibility. QR codes support three encoding modes, and uppercase alphanumeric characters can be encoded in "alphanumeric mode" which is more space-efficient than "byte mode" required for mixed-case strings.
->
-> **Encoder Requirements:**
->
-> - Implementations SHOULD emit uppercase Bech32m strings (e.g., `CREQB1...`)
-> - This typically results in 30-40% smaller QR codes compared to lowercase
-> - The HRP `creqb` becomes `CREQB` when uppercased
->
-> **Decoder Requirements:**
->
-> - Implementations MUST accept both uppercase and lowercase variants
-> - Implementations MUST accept mixed-case input
-> - The Bech32m checksum validation is case-insensitive
-> - Case conversion should be performed before checksum validation
->
-> **Example:**
->
-> - Uppercase (recommended): `CREQB1QQQSYQQQPJHZP24NGU3...`
-> - Lowercase (valid): `creqb1qqqsyqqqpjhzp24ngu3...`
-> - Mixed-case (valid but not recommended): `CreqB1QqqSyQqqPjhzp24ngu3...`
+> [!NOTE]
+> Implementations SHOULD output uppercase Bech32m strings for optimal QR code compatibility. Uppercase alphanumeric characters use QR "alphanumeric mode" which is more space-efficient than "byte mode" required for mixed-case. Decoders MUST accept both uppercase and lowercase input.
+
+When parsing a `creq` parameter, implementations SHOULD support both formats:
+
+1. If the parameter starts with `creqA` (case-insensitive), parse as NUT-18 CBOR+base64 format
+2. If the parameter is valid Bech32m with HRP `creqb`, parse as NUT-26 format
+3. Otherwise, return an error
 
 ## TLV Structure
 
@@ -68,34 +45,6 @@ The payment request is encoded as a sequence of TLV fields. Each TLV entry consi
 | 0x08 | nut10       | sub-TLV   | NUT-10 spending conditions (corresponds to `nut10` in JSON)          |
 
 All fields are optional. Unknown tags MUST be ignored to maintain forward compatibility.
-
-#### NUT-18 Field Name Mapping
-
-The TLV encoding uses numeric tags while the JSON format (NUT-18) uses single-character field names. The following table provides the complete mapping:
-
-| TLV Tag | Field Name  | JSON Key | Example Value                    |
-| ------- | ----------- | -------- | -------------------------------- |
-| 0x01    | id          | `i`      | `"demo123"`                      |
-| 0x02    | amount      | `a`      | `1000`                           |
-| 0x03    | unit        | `u`      | `"sat"` or `"msat"`              |
-| 0x04    | single_use  | `s`      | `true` or `false`                |
-| 0x05    | mint        | `m`      | `["https://mint.example.com"]`   |
-| 0x06    | description | `d`      | `"Coffee payment"`               |
-| 0x07    | transport   | `t`      | Array of transport objects       |
-| 0x08    | nut10       | `nut10`  | NUT-10 spending condition object |
-
-**Example JSON to TLV Mapping:**
-
-```json
-{
-  "i": "req123",
-  "a": 500,
-  "u": "sat",
-  "d": "Test payment"
-}
-```
-
-Maps to TLV tags: `0x01` (id), `0x02` (amount), `0x03` (unit), `0x06` (description).
 
 ### Unit Encoding (Tag 0x03)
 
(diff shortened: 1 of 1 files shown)


26.md
```diff
@@ -0,0 +1,256 @@
...
+
+This allows encoding arbitrary key-value pairs for extensibility.
+
+### NUT-10 Sub-TLV (Tag 0x08)
+
+NUT-10 spending conditions are encoded as nested TLV structures:
+
+| Sub-Tag | Field     | Type        | Description                                                  |
+| ------- | --------- | ----------- | ------------------------------------------------------------ |
+| 0x01    | kind      | u16         | Secret kind (0=P2PK, 1=HTLC, etc.)                           |
```
a1denvalu3:
```suggestion
| 0x01    | kind      | u8         | Secret kind (0=P2PK, 1=HTLC, etc.)                           |
```

26.md
```diff
@@ -0,0 +1,256 @@
...
+| 0x02    | data      | bytes       | Kind-specific data (UTF-8 encoded)                           |
+| 0x03    | tag_tuple | sub-sub-TLV | Tag tuple (repeatable, uses same encoding as transport tags) |
+
+#### NUT-10 Kind Enumeration
+
+The following kind values are defined for NUT-10 spending conditions:
+
+| Kind Value | Name | Description                                                      |
+| ---------- | ---- | ---------------------------------------------------------------- |
+| 0x0000     | P2PK | Pay to Public Key - requires signature from specified public key |
```
a1denvalu3:
```suggestion
| 0x00     | P2PK | Pay to Public Key - requires signature from specified public key |
```

26.md
```diff
@@ -0,0 +1,256 @@
...
+| 0x0001     | HTLC | Hash Time Locked Contract - requires preimage of hash            |
```
a1denvalu3:
```suggestion
| 0x01     | HTLC | Hash Time Locked Contract - requires preimage of hash            |
```

tests/26-test.md
```diff
@@ -0,0 +1,154 @@
...
+      "g": [["n", "17"]]
+    }
+  ]
+}
+```
+
+Encoded (Bech32m):
+
+```
+CREQB1QYQQSC3HVYUNQVFHXCPQQZQQQQQQQQQQQQ9QXQQPQQZSQ9MGW368QUE69UHNSVENXVH8XURPVDJN5VENXVUQWQRDQYQQZQGZQQSGM6QFA3C8DTZ2FVZHVFQEACMWM0E50PE3K5TFMVPJJMN0VJ7M2TGRQQPQQYGYQQ28WUMN8GHJ7UN9D3SHJTNYV9KH2UEWD9HSGQQHWAEHXW309AEX2MRP0YHRSVENXVH8XURPVDJJ7PQQP4MHXUE69UHKUMMN9EKX7MQ8402ZW
```
a1denvalu3: remember to update this

tests/26-test.md
```diff
@@ -0,0 +1,154 @@
...
+      ]
+    }
+  ]
+}
+```
+
+Encoded:
+
+```
+CREQB1QYQQSE3EXFSN2VTZ8QPQQZQQQQQQQQQQQPJQXQQPQQZSQXTGW368QUE69UHK66TWWSCJUETCV9KHQMR99E3K7MG9QQVKSAR5WPEN5TE0D45KUAPJ9EJHSCTDWPKX2TNRDAKSWQPWQYQQZQGZQQSQQQQQQQQQQQQQQQQQQQQQQQQQQQQQQQQQQQQQQQQQQQQQQQQQQQQRQQZQQYFXQU6YHAEU
```
a1denvalu3: And this

tests/26-test.md
```diff
@@ -0,0 +1,154 @@
...
+    "d": "02c3b5bb27e361457c92d93d78dd73d3d53732110b2cfe8b50fbc0abc615e9c331",
+    "t": [["timeout", "3600"]]
+  }
+}
+```
+
+Encoded:
+
+```
+CREQB1QYQQSCEEV56R2EPJVYPQQZQQQQQQQQQQQ86QXQQPQQZSQXRGW368QUE69UHK66TWWSHX27RPD4CXCEFWVDHK6ZQQTGQSQQSQQQPQQS3SXF3NXC34VF3RYDM9XVMRZDP4XA3NJVNY8YEKGDECV3JRWVMYXDJR2VEHXVERZVFSVGEXXEN98P3R2VRXVF3NQCTZVVMRZDT9893NXVE3QVQQ6PM5D9KK2MM4WSZRXD3SXQQ5SCYD
```
a1denvalu3: and this as well

thesimplekid: updated

tests/26-test.md
thesimplekid: There is no kind here so it is unchanged.

tests/26-test.md
thesimplekid: There is no kind here so it is unchanged.

26.md
```diff
@@ -0,0 +1,256 @@
...
+Maps to TLV tags: `0x01` (id), `0x02` (amount), `0x03` (unit), `0x06` (description).
+
+### Unit Encoding (Tag 0x03)
+
+The unit field uses a compact encoding:
+
+- **Value 0x00**: Represents `sat` (Bitcoin satoshis)
+- **String value**: Any other unit is encoded as a UTF-8 string (e.g., `"msat"`, `"usd"`, `"eur"`)
+
+If an amount (tag 0x02) is present but unit (tag 0x03) is absent, decoders SHOULD default to `sat`.
```
Egge21M: I think this might be overreaching. I don't think other parts of our specifications dictate how this is handled, but most wallets default to 'sat' is unit is missing

26.md
```diff
@@ -0,0 +1,256 @@
...
+
+### Transport Sub-TLV (Tag 0x07)
+
+Transport configurations are encoded as nested TLV structures. Each transport has the following sub-tags:
+
+| Sub-Tag | Field     | Type        | Description                                              |
+| ------- | --------- | ----------- | -------------------------------------------------------- |
+| 0x01    | kind      | u8          | Transport type: 0=in-band, 1=nostr, 2=http_post          |
+| 0x02    | target    | bytes       | Transport target (interpretation depends on kind)        |
+| 0x03    | nips      | u16[]       | Nostr NIPs (sequence of big-endian u16, only for kind=1) |
```
Egge21M: In NUT-18 nips are defined as part of the tags array

26.md
```diff
@@ -0,0 +1,256 @@
...
+| 0x04    | relay     | string      | Nostr relay URL (repeatable, only for kind=1)            |
```
Egge21M: If a relay is required it should be bech32 encoded as part of an nprofile in `target`

26.md
Egge21M: This might be overreaching. I think as of right now we do not specify how wallets should treat missing `unit` in other parts of the specifications. While this might be debatable I think making that decision should not be made by the decoder.

26.md
```diff
@@ -0,0 +1,256 @@
...
+
+**JSON Representation:**
+
+In the NUT-18 JSON format, transports are represented with a `type` field:
+
+```json
+{
+  "t": [
+    { "type": "nostr", "target": "npub1...", "tags": [["n", "17"]] },
+    { "type": "http_post", "target": "https://callback.example.com/pay" }
```
Egge21M: NUT-18 uses `post`

26.md
```diff
@@ -0,0 +1,256 @@
...
+- **Type** (1 byte): Field identifier
+- **Length** (2 bytes, big-endian): Length of value in bytes
+- **Value** (variable): Field data
+
+### Top-Level TLV Tags
+
+| Tag  | Field       | Type      | Description                                                          |
+| ---- | ----------- | --------- | -------------------------------------------------------------------- |
+| 0x01 | id          | string    | Payment identifier (corresponds to `i` in JSON)                      |
+| 0x02 | amount      | u64       | Amount in base units (corresponds to `a` in JSON)                    |
```
Egge21M, line 39: Making this field a u64 means JavaScript implementations need to handle this as BigInt instead of a number. Realistically u32 should be sufficient, right?

26.md
```diff
@@ -0,0 +1,256 @@
...
+```
+
+When encoding to TLV, the `type` string is converted to the corresponding numeric kind value.
+
+#### Transport Target Encoding (Sub-Tag 0x02)
+
+The target field is interpreted based on the transport kind:
+
+- **kind=0 (in-band)**: Empty or omitted
+- **kind=1 (nostr)**: 32-byte X-only public key (raw bytes, not bech32-encoded)
```
Egge21M: Whats the rational behind this? It makes the encoding much more complicated and adds the necessity of additional relay tags. If target was expected to be utf-8 as it is in NUT-18 this would streamline things.

thesimplekid: We do this for space efficiency. An nprofile string is ~150-200 bytes while a raw 32-byte pubkey + relay tag tuples is ~60 bytes. Avoiding nested bech32 gives roughly 3x savings.

26.md
thesimplekid: I intentionally avoid embedding bech32-encoded nprofiles in the target field. Since the payment request itself is bech32m encoded, nesting bech32 inside bech32 inefficient.

26.md
thesimplekid: I removed the nips and just use the tags for the nips and relay in bf62b9b4d39dcb84c027ebe89e78401d0988695b

26.md
Egge21M: I see, that makes sense. I guess with the current amount of different transports and the fact that nostr is common enough to allow a special case this LGTM. But we should probably keep the special cases minimal in the future

26.md
a1denvalu3: Mints already expose keys for amounts up to 2**64. How does cashu-ts handle this?

thesimplekid: looks like in other places we use int not specific as u64 or u32. https://github.com/cashubtc/nuts/blob/779993681c481075f9ed2cd99d2d61ca9e70e835/00.md?plain=1#L61 Maybe we should be more specific in those. Mints can have keysets with amounts that are greater then u32::MAX

26.md
thesimplekid: > But we should probably keep the special cases minimal in the future

Agree

26.md
Egge21M: > Mints already expose keys for amounts up to 2**64. How does cashu-ts handle this?

@lollerfirst It doesn't and its an issue. But mint keys are `string` until they actually get used, so it shouldnt matter. Arithmetic also works fine up to 9,007,199,254,740,992 which should be fine in 99.9999% of all cases

tests/26-test.md
```diff
@@ -0,0 +1,150 @@
...
+```json
+{
+  "i": "f92a51b8",
+  "a": 100,
+  "u": "sat",
+  "m": ["https://mint1.example.com", "https://mint2.example.com"],
+  "t": [
+    {
+      "t": "nostr",
+      "a": "npub1qqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqzqujme",
```
Egge21M: In NUT-18 ever nostr target is a nprofile. When no relays are specified its an nprofile with empty relay list

Egge21M (changes requested):

tests/26-test.md
thesimplekid: updated in https://github.com/cashubtc/nuts/pull/294/commits/887f985b3dbeb65392faaa9ad43809be58021164

26.md
thesimplekid: going to keep it a u64 here

Egge21M:
We are missing a signal for P2BK as @robwoodgate mentioned here: https://github.com/cashubtc/cashu-ts/pull/436#pullrequestreview-3598747826

robwoodgate:
> We are missing a signal for P2BK as @robwoodgate mentioned here: https://github.com/cashubtc/cashu-ts/pull/436#pullrequestreview-3598747826

Wondering if some kind of general "tag" is possible that allows extension of the TLV schema without having to redefine it each time? I think this is what TSK was talking about on the p2bk nut pr?

Egge21M:
This has been finalized and implemented for three weeks. No NACKs remain. I think this should be merged, but my green dont count. @callebtc

callebtc (approved):

PR #296, opened 2025-10-14
NUT-XX: Offline Spilman (unidirectional) channel

SatsAndSports:
With this NUT, Alice and Bob can - using any mint supporting P2PK (NUT-11) - set up a channel where, once this 'cashu channel' is opened, Alice can decrease her own balance and increase Bob's balance. After each update, Bob can verify the new balance without needing to contact the mint for every update of the balance.

This allows a very large number (billions) of balance updates, where each update may be very small (millisat), but with only using a small number (dozens) of proofs prepared in advance.

Some features:
 - blinding the pubkey, for extra privacy, using P2BK (NUT-28)
 - deterministic outputs, as a function of the channel parameters, to minimize the bandwidth
 - the receiver can verify the opening of a channel and the payments without needing to contact the mint. The server contacts the mint only when they decide to close the channel

### DIFF
--- a/XX.md
+++ b/XX.md
@@ -4,10 +4,6 @@
 
 `depends on: NUT-11 (P2PK), NUT-12 (DLEQ), NUT-09 (restore signatures), NUT-07 (token state check) (NUT-17 is beneficial, but not required)`
 
-_TODO: who pays the fees? (see the multi-swap zero-receiver-setup deterministic outputs below) If Alice spends 101 sats to fund the 'funding token' which has a _nominal_ value of 100 sats, and Charlie's exit with the full capacity gives him just nominal 99 sats in his 1-of-1 P2PK outputs, which decrease to 98 sats when he finally swaps them to anyone-can-spend tokens in his wallet, then who should pay those fees? Which of those four numbers is the capacity of the network?_
-
-_TODO: question for the reviewer: Could we use pay-to-blinded-pubkey here? Where exactly? And should we make P2BPK required or optional?_
-
 ---
 
 This describes how Alice can set up a one-way payment channel from herself to Charlie.
@@ -28,7 +24,7 @@ Alice can then immediately claim her balance, even if Charlie does not cooperate
 
 The mint, Bob, is involved only at the start for the initial swap where Alice prepares
 the _funding token_, and at the end where each of the two parties swap their
-final balances into their cashu wallets.
+final balances into their wallets.
 
 # Exiting and fees, and the capacity of the channel
 
@@ -36,30 +32,35 @@ When the channel is closing, there are two stages, and there are fees to be paid
 in each of those two stages.
 First, the _funding token_ token is swapped to create the _deterministic outputs_
 (defined below) for each of the two parties.
+Some of the deterministic outputs are P2PK-locked to Bob's public key, and the remainder
+are of the deterministic outputs are P2PK-locked to Alice's public key.
 
 All the proofs in the _funding token_ are in the same _active_ keyset, with fee rate `input_fee_ppk`.
 If there are `n_funding_proofs` proofs in that funding token, the fees of that swap cost
 `(input_fee_ppk * n_funding_proofs + 999) // 1000`,
 where `//` rounds non-integer results down to an integer.
 
-The deterministic outputs, which are made up of some outputs for Alice and some for Charlie,
-are also in the same keyset and they are P2PK-locked to the intended recipient.
-
 In the second stage of the exit, Alice and Charlie swap these deterministic outputs
-for conventional anyone-can-spend proofs. This also requires paying fees.
+for conventional anyone-can-spend proofs.
+This also requires paying fees. these outputs are also in the same keyset as the funding token.
 
 If there are non-zero fees, i.e. `input_fee_ppk > 0', Alice is responsible for paying all the fees.
 
-The method for constructing the determistic outputs is described in more detail below.
-For now, consider a function `deterministic_value_after_fees(x)` which constructs the
+The method for constructing the determistic outputs is described in more detail later in this document.
+For now, we focus on the fees.
+Consider a function `deterministic_value_after_fees(x)` which constructs the
 deterministic outputs with _nominal_ value `x` and then substracts the fees that would
 result from swapping those deterministic outputs.
 Where `num_deterministic_outputs(x)` is the number of outputs needed,
 
-`deterministic_value_after_fees(x) = x - (input_fee_ppk * num_deterministic_outputs(x) + 999) // 1000`
+```
+deterministic_value_after_fees(x)
+  =
+    x - (input_fee_ppk * num_deterministic_outputs(x) + 999) // 1000
+```
 
 To ensure that Bob's final balance is `bobs_balance`, Alice must compute the inverse
-of 'deterministic_value_after_fees`, i.e. `x = inverse_deterministic_value(bobs_balance)`,
+of `deterministic_value_after_fees`, i.e. `x = inverse_deterministic_value(bobs_balance)`,
 which is the smallest `x` such that `deterministic_value_after_fees(x) = bobs_balance`.
 
 To make a payment which brings Bob's balance to `bobs_balance`, the following summarizes the values and fees:
@@ -92,6 +93,7 @@ Knowing Charlie's public key, and the set of mints and _units_ that are trusted 
  - `unit`: typically `sat` or `msat`
  - `capacity`: the number of sats in the _funding token_ that Alice will create while funding the channel. _**If the fees are non-zero, Charlie's maximum possible balance will be less than the `capacity`**_
  - `active_keyset_id`: An _active_ keyset for that unit at that mint
+ - `input_fee_ppk`: As all the proofs in the funding token, and in the deterministic outputs, are in the same keyset, the `input_fee_ppk` is fixed.
  - `expiry`  unix timestamp: If Charlie doesn't close before this time, Alice can re-claim all the funds after this has expired
  - `setup_timestamp`: unix timestamp: the time when Alice is setting up this channel
  - `sender_nonce`: Random data selected by Alice to add more randomness. May be useful if Alice and Charlie have multiple concurrent channels.
@@ -188,6 +190,8 @@ Deterministic Secret:
   }
 ]
 ```
+
+_TODO? Optional: If Bob, along with his public key, advertizes that he supports pay-to-blinded-pubkey, we could determistically compute an ephemeral private key for blinding, and include the corresponding ephemeral public key in the proof, as per the P2BPK NUT_
 
 The corresponding blinding factor is `SHA256(channel_id || pubkey || amount || "blinding" || index)`,
 and so the deterministic output (BlindedMessage) is contructed by applying that
--- a/XX.md
+++ b/XX.md
@@ -336,14 +336,25 @@ As recommended in NUT-03, the outputs are in ascending order of amount.
 
 The outputs includes all the commitment outputs for Charlie and also
 those for Alice.
-They are ordered for by amount (increasing). For any given amount,
+They are ordered by amount (increasing). For any given amount,
 we have Charlie's deterministic outputs first, ordered by _index_ (increasing),
-and then Alice's similarly ordered
+and then Alice's similarly ordered. For example, if you're distributing
+a few hundred sats to each party, the first few outputs of the swap are:
+
+```
+context    | amount | index
+
+"receiver" | 100    |     0
+"receiver" | 100    |     1
+"sender"   | 100    |     0
+"sender"   | 100    |     1
+...
+```
 
 > [!NOTE]
 > If you have a vector of Charlie's outputs, ordered by amount and index, and then you append
 > Alice's similarly ordered,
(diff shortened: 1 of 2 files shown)


XX.md
```diff
@@ -0,0 +1,56 @@
...
+## Denominations
+
+_This assumes that 1 millisat is the smallest 'resolution' of interest, but another resolution may be agreed_
+
+Alice shall prepare _2_ such proofs with 1-millisat value.
+One of these 1-millisat proofs is a special proof that will be included in every transaction.
+Alice shall prepare one proof for every other power of 2: 2-millisats, 4-millisats, 8-millisats, ..., up to some pre-agreed maximum.
+The total value of all those proofs is the channel capacity and is the maximum amount that Alice can send to Bob through this channel
+
+Bob can verify all of these proofs with the mint, before going offline.
```
c03rad0r: Perhaps Bob doesn't need to check with the mint before going offline if Bob maintains a redeemed stack, because bob can verify that the proofs are locked to his pub key, that they were signed by the mint and that Bob hasn't redeemed them yet.

SatsAndSports: > because bob can verify that the proofs are locked to his pub key, that they were signed by the min

I think I agree.  Do you mean this offline check that Bob can do via Nut-12?

https://github.com/cashubtc/nuts/blob/main/12.md

I don't understand that fully yet myself

c03rad0r: Indeed, based on the following LLM summary of `NUT-12`, it seems to be the right tool for the job. I didn't realize this is so involved. I just assumed anyone would be able to verify that proofs are signed by the mint.. Happy to try to understand it together next time we cowork :)

```
Previously, only the mint itself could verify its own signatures. NUT-12 introduces DLEQ proofs, which allow users like Alice (who interacts with the mint) and Carol (who receives ecash from Alice) to verify that the mint (Bob) used the correct private key for signing.

The core idea is to prove that the private key used to create the mint's public key (A) is the same private key used to sign a blinded message (B').
```

SatsAndSports:
I've updated this now, including various smaller improvements made based on lessons learned in the last few months while implementing demo code

For example, the final outputs are now locked via the blinding method of NUT-28. And I slightly tidied up the channel parameters

SatsAndSports:
As it's been a couple of months since the last change to this PR, and this was mentioned in the Cashu Dev Call a few minutes ago, I'll share a little:

In the last couple of weeks, I've started working again on something I'm calling [MONAD](https://github.com/SatsAndSports/MONAD), which I started a few weeks ago while with some Bitcoiners on Madeira.

It's basically like Tor or VPNs, for onion routing of internet access, but where you pay the relays via these Cashu channels. My hope is to use that an interesting demo to show that the protocol works, and the ability to make many tiny payments to a service providier.

But I'm not sure if that's the right approach though! Maybe I/we should focus on this doc more, and less on demo code. What do you think?

XX.md
```diff
@@ -0,0 +1,733 @@
...
+
+```
+deterministic_nonce(...)
+ =
+   SHA256(channel_secret || "{channel_id}|{output_context}|{amount}|nonce|{index}")
+
+blinding_factor(...)
+ =
+   SHA256(channel_secret || "{channel_id}|{output_context}|{amount}|blinding|{index}")
+```
```
a1denvalu3, line 436: I would advise using an HMAC-SHA256 construction, with channel secret as the key and the rest as the message

XX.md
```diff
@@ -0,0 +1,733 @@
...
+
+where `stage2_context` is `receiver_stage2` or `sender_stage2`, for Charlie's balance and Alice's remainder respectively.
+The `retry_counter` is incremented until the hash yields such a valid scalar
+for use as the ephemeral secret key.
+
+Then an ECDH shared secret is computed between that ephemeral private key and
+the recipient's public key, and the NUT-28 tweak scalar is derived as:
+
+```
+stage2_tweak_scalar = SHA256("Cashu_P2BK_v1" || shared_secret_x || 0x00)
```
a1denvalu3, line 529: I would define `shared_secret_x` at the cost of being redundant. It's better to be exact, so that LLMs can implement more easily.

XX.md
```diff
@@ -0,0 +1,733 @@
...
+# Timing
+
+To improve privacy further, when the channel is closed and the first of the two stages
+is executed, both parties should delay the second stage.
+This delay is to make it more difficult for the mint to correlate the three 'exit swaps' with each other.
+
+
+# proof-of-concept
+
+_As of 2026-03-20, the implementation includes P2BK (Pay-to-Blinded-Key) with per-proof blinding for stage 2 outputs. The funding and commitment outputs are deterministic with blinded pubkeys. Fees are taken into account fully. The token-state-check and token-restore endpoints are used. The implementation is [here](https://github.com/SatsAndSports/demo_of_spillman_cashu_channel/tree/spilman.channel/crates/cdk/src/spilman), with many integration tests._
```
a1denvalu3, line 733: how can per-proof P2BK blinding work with SIG_ALL? IIRC this was a known problem, where SIG_ALL requires all inputs outputs to be signed together by the pubkey(s) and all inputs having the same spending conditions.

XX.md
```diff
@@ -0,0 +1,733 @@
...
+
+_(Question: Does NUT-09 specify what should happen if the outputs are a mixture of unspent outputs and outputs that were never spent? i.e. Does Alice have to call restore once for each possible output, or can she do a batch call where some will fail and some will succeed?)_
+
+# Charlie's verification
+
+The following information should be sent by Alice to Charlie with, or before, the first payment in this channel,
+enabling Charlie to verify everything. He can verify without communicating with the mint.
+
+ - the channel parameters
+ - the funding token, including the DLEQ proofs (NUT-12)
```
a1denvalu3, line 682: With keysets v3, the ecash is issued on the BLS-12-381 curve, which supports special math that allows for public verification. The ecash notes themselves change from being simple algebraic MACs to being proper signatures, so DLEQ is no longer needed.

The nested crypto layer (spending conditions) remains secp-based, so you don't have to change anything.

SatsAndSports: > so DLEQ is no longer needed.

Indeed. So I guess that (unless v2 keysets are fully replace soon), this NUT will have to say that DLEQ proofs are needed only for v2, and that for v3 the channel partipicants should do the new BLS-based verification instead of checking the (now-missing) DLEQ proofs

SatsAndSports:
Another small update on this. First, to recap what I said last month:

> As it's been a couple of months since the last change to this PR, and this was mentioned in the Cashu Dev Call a few minutes ago, I'll share a little:
> 
> In the last couple of weeks, I've started working again on something I'm calling [MONAD](https://github.com/SatsAndSports/MONAD), which I started a few weeks ago while with some Bitcoiners on Madeira.
> 
> It's basically like Tor or VPNs, for onion routing of internet access, but where you pay the relays via these Cashu channels. My hope is to use that an interesting demo to show that the protocol works, and the ability to make many tiny payments to a service providier.


I've added a lot to MONAD since then. Last month, the networking and routing was in good shape, but the payment stuff was a bit immature. Over the last few weeks, I've made the payments much more robust. In normal cases, a single payment over a channel is very easy. But the full lifecycle of a channel involves various swaps to open and close the channel. And any of those swaps could fail, and therefore I've been working on making it more robust, so that both parties in the channel can retry and recover failed swaps. MONAD automatically opens new channels on demand, so users can continue to use MONAD seamlessly for a long time

While the MONAD repo is the main repo, I've put all the non-MONAD-specific payment stuff into [`cashu_spilman_channels` (github is out of date, as I haven't pushed the latest changes)](https://github.com/SatsAndSports/cashu_spilman_channels). That includes all the logic for opening and closing channels, robustly, and it's intended to be a general-purpose library that can be used with any other system. For example, the interfaces and implementations in `cashu_spilman_channels` don't make assumptions about what service you are providing, nor about what pricing you are using.

While I've been busy with that, and I'm happy with this foundation, it means that I haven't yet wrapped it up in an easy-to-use and easy-to-demo project. It won't take long to make it demo-able with a nice user-interface to monitor the connections and manage the channels, probably just a week or so

And thanks to @a1denvalu3 for the review a few weeks ago. They are good suggestions and definitely will be incorporated

XX.md
SatsAndSports: There are two stages in this, each of which involves blinding

The 'channel token' itself can only be spent with SIG_ALL and therefore, as you say, the individual proofs must be identical in certain respects. There are two keys for the 2-of-2 spend path, and a third for the post-expiry refund. Those three keys are different from each other, and different from the user's "real" pubkey, but they are identical across all the outputs in this 'channel token'

The second stage is where the 'channel token' is spent and it creates outputs to distribute the funds between the two parties. The proofs created from those outputs are themselves 1-of-1 P2BK, and they are _not_ SIG_ALL, and therefore each output is blinded differently

SatsAndSports:
An update on what I've been doing in the last month. I've been more on MONAD (see my comment just above from last month), and hope to share more soon as it's close to being ready to be shared.

I have more free time now and for the coming weeks, and so - now and after today's Cashu Dev Call - I will:

 - Act on all the comments already in this PR. I read them before, and they are actionable, I just didn't make time to do anything concrete since reading them.
 - Add test vectors to this doc. Initially just placeholders, but then I'll also fully complete them and implement them in the [`cashu_spilman_channels`](https://github.com/SatsAndSports/cashu_spilman_channels)
 - Make the `cashu_spilman_channels` library a bit more robust, like setting up Github Actions so that all the tests and enforced within Github

In parallel to the above, I'll reach out to potential users of the channel tech, e.g. Routstr and Tollgate, to see if the `cashu_spilman_channels` library might be useful to them. MONAD is a consumer of that library and, even if MONAD itself doesn't become a big success, it might act as a demo of how to use the library

XX.md
```diff
@@ -0,0 +1,733 @@
...
+The secret's _nonce_ and the blind-signature _blinding factor_ are computed as follows:
+
+```
+deterministic_nonce(...)
+ =
+   SHA256(channel_secret || "{channel_id}|{output_context}|{amount}|nonce|{index}")
+
+blinding_factor(...)
+ =
+   SHA256(channel_secret || "{channel_id}|{output_context}|{amount}|blinding|{index}")
```
SatsAndSports, line 435:
```suggestion
   HMAC-SHA256(key = channel_secret, message = "{channel_id}|{output_context}|{amount}|blinding|{index}")
```

XX.md
```diff
@@ -0,0 +1,733 @@
...
+
+As these are all deterministic and based on information known to both parties,
+both parties can construct all these outputs and secrets and blinding factors.
+
+The secret's _nonce_ and the blind-signature _blinding factor_ are computed as follows:
+
+```
+deterministic_nonce(...)
+ =
+   SHA256(channel_secret || "{channel_id}|{output_context}|{amount}|nonce|{index}")
```
SatsAndSports, line 431:
```suggestion
   HMAC-SHA256(key = channel_secret, message = "{channel_id}|{output_context}|{amount}|nonce|{index}")
```

XX.md
SatsAndSports: Done (in the sense that I have - in the last few seconds - made a suggestion in Github to apply this HMAC-256 construction. Is it what you expected?

XX.md
```diff
@@ -0,0 +1,733 @@
...
+where `stage2_context` is `receiver_stage2` or `sender_stage2`, for Charlie's balance and Alice's remainder respectively.
+The `retry_counter` is incremented until the hash yields such a valid scalar
+for use as the ephemeral secret key.
+
+Then an ECDH shared secret is computed between that ephemeral private key and
+the recipient's public key, and the NUT-28 tweak scalar is derived as:
+
+```
+stage2_tweak_scalar = SHA256("Cashu_P2BK_v1" || shared_secret_x || 0x00)
+```
```
SatsAndSports, line 530:
```suggestion
 Then an ECDH shared secret is computed as defined by NUT-28 P2BK. Let `Zx` denote the 32-byte x-coordinate shared secret from NUT-28:
 
   - sender side: `Zx = x(eP)` 
   - receiver side: `Zx = x(pE)`

where `e` is the sender ephemeral private key, `E = eG` is the sender ephemeral public key, `p` is the recipient private key, and `P = pG` is the recipient public key.

`stage2_tweak_scalar = SHA256("Cashu_P2BK_v1" || Zx || 0x00)`
      
```

XX.md
SatsAndSports: Done. I've made a suggestion a moment ago to bring clarity to how this scalar is computed.

As this is just the NUT-28 method, I guess we should just refer to that NUT directly, instead of copying-and-pasting the relevant parts of that spec into this. What do you think?

SatsAndSports:
Mini-update, on v3 keyset support:

I'm making progress with v3 now, writing Rust code that works with the Nutshell branch that has v3 support. My immediate goal isn't to make tidy or usable code; I'm on a branch of `cashu_spilman_channels` that has these hacks. I'm just learning more practical things about how v3 works, which is helping me to understand what is needed for v3 in the channel protocol

Therefore, I think the next changes I make to this PR in coming days will be to add more v3 discussion

As one small example, in v1/v2, this proposal uses a deterministic `Secret.nonce` that is different in each 2-of-2 proof in the funding token. In v3, the same deterministic number will be used, but instead of being used a `Secret.nonce`, it will be used as the offset applied in the NUMS point.

SatsAndSports:
If V3 is formalized soon, we could perhaps remove V1 and V2 entirely from this and make Spilman channels use V3 only

(Not my idea, just writing down an idea that was suggested by someone else today)

Issue #290, opened 2025-09-27
Blinding the locking keys in NUT-11


lollerfirst:
The current state of NUT-11 provides that we input the public key itself into the well known secret structure:

```json
[
  "P2PK",
  {
    "nonce": "da62796403af76c80cd6ce9153ed3746",
    "data": "033281c37677ea273eb7183b783067f5244933ef78d8c3f15b1a77cb246099c26e",
    "tags": [
      ["sigflag", "SIG_INPUTS"],
    ]
  }
]
```
This means that when -during redemption- we provide the unblinded secret, the Mint is able to see the public key. If this public key represents a Nostr identity or is not ephemeral, the Mint is able to tell who is claiming the payment.

## Blinding the key

Let $P$ be the public key `033281c37677ea273eb7183b783067f5244933ef78d8c3f15b1a77cb246099c26e`. 
Instead of locking to $P$, we could lock to $P' = P + r \cdot G$, where $r$ is a random factor.

Then the sender can include $r$ together with the payment, for example it could be included in the `Proof` object:
```json
{
  "amount": int,
  "id": hex_str,
  "secret": str,
  "C": hex_str,
  "r": [hex_str, ...]
}
```

The receiver can calculate the correct signing key by combining their secret key $p$ and $r$: $k = p + r$.
So as you can see this requires no alterations to the verification logic of the Mint. It's all client side.

PR #300, opened 2025-10-19
NUT28 - ECDH-derived Pay-to-Blinded-Key (P2BK)

robwoodgate:
**CLOSES** #290 
**REPLACES:** #291 

### Implementations:

- [x] Cashu-TS 
    - Original: https://github.com/cashubtc/cashu-ts/pull/377
    - Revised: https://github.com/cashubtc/cashu-ts/pull/465
- [x] CDK https://github.com/cashubtc/cdk/pull/1253
- [x] DotNut 
    - Original:  https://github.com/Kukks/DotNut/pull/21 
    - Revised: https://github.com/Kukks/DotNut/pull/28

## Summary:
Defines P2BK as an *ECDH-derived* blinding scheme instead of one using random scalars.

Each proof now includes a per-proof ephemeral pubkey `p2pk_e`, from which both parties can derive the same blinding factor(s) deterministically.

ECDH shared secret works because:

```blockquote
Receiver long-term keypair: (p, P = p·G)
Sender ephemeral keypair: (e, E = e·G)
Shared secret: Zx = x(e·P) = x(p·E)   (32-byte x-coordinate)
Both compute the same point: e·p·G =  e·P = p·E

NOTE: Each receiver pubkey (P) has their own unique shared secret, and can ONLY derive their own.
```

Importantly, 3rd parties and the mint CANNOT derive the original locking pubkeys. Only the sender and the receiver have the secret keys required to calculate the ECDH shared secret, which can derive both the original pubkeys and the signing secret.

Proofs can be locked to a well known public key, posted in public without compromising privacy, and spent by the recipient without needing any side-channel communication.

### Key points:

* Adds `p2pk_e` (33-byte SEC1 pubkey) per proof (stored as `pe` in token v4 format)
* Uses deterministic blinding:
``` rᵢ = SHA-256( b"Cashu_P2BK_v1" || Zx || keyset_id_bytes || i_byte) ```
where Zx is a shared ECDH secret, `keyset_id_bytes` is the hex_to_bytes of keyset ID, and `i_byte` is the P2PK locking key "slot" position.
* No mint or protocol changes required.

### Assumptions

- The explicit 11 slot range (0-10) assumes the limits laid out in #255

### Live Demo: 

- Create P2BK locked tokens using [Cashu NutLock](https://www.nostrly.com/cashu-nutlock/)
- Redeem P2BK using either [Cashu Witness](https://www.nostrly.com/cashu-witness/) or [Cashu Redeem](https://www.nostrly.com/cashu-redeem/)

### DIFF
--- a/28.md
+++ b/28.md
@@ -43,8 +43,6 @@ Where:
 - `i_byte` is the single unsigned byte representation of `i`: (`0x00` to `0x0A`)
 - `||` denotes concatenation
 
-For broader compatibility, `rᵢ` **MUST NOT** be normalised modulo `n`
-
 If `rᵢ` is not in the range `1 ≤ rᵢ ≤ n−1`, retry once with an extra `0xff` byte appended to the hash input as follows:
 
 ```

### CODE REVIEW

File: package-lock.json [hunk 1]
```diff
@@ -0,0 +1,6 @@
+{
```
d4rp4t: this file was probably added by accident

d4rp4t (changes requested):
You've added a package.json file, propably by accident

lollerfirst (approved):

robwoodgate:
> You've added a package.json file, propably by accident

Good catch - thanks @d4rp4t. Fixed

d4rp4t (approved):

lollerfirst:
@robwoodgate  We should derive a `Z_i` for each locking key listed in the secret, so that the knowledge of one blinding factor doesn't necessarily imply knowledge of every other. This way, we can also throw away the slot indices ($i \in [0, 10]$).

Use case example:
```
Alice locks to Carol's public key with a locktime and a refund key. Carol can't unblind Alice's refund key.
```

I've opened a PR [here](https://github.com/robwoodgate/nuts/pull/1)

robwoodgate:
> @robwoodgate  We should derive a `Z_i` for each locking key listed in the secret, so that the knowledge of one blinding factor doesn't necessarily imply knowledge of every other. This way, we can also throw away the slot indices ($i \in [0, 10]$).
> 
> Use case example:
> 
> ```
> Alice locks to Carol's public key with a locktime and a refund key. Carol can't unblind Alice's refund key.
> ```
> 
> I've opened a PR [here](https://github.com/robwoodgate/nuts/pull/1)

We already DO calculate shared secret (Zx) per locking key. 

```
For each receiver key P, compute:
a. Slot index i in [data, ...pubkeys, ...refund]
b. Zx = x(e·P)
c. rᵢ = H("Cashu_P2BK_v1" || Zx || keyset_id || i) mod n
d. P′ = P + rᵢ·G
```

The slot index is just for ADDITIONAL uniqueness, so that if the same key (P) is added to both pubkeys and refund, it will be uniquely blinded by the slot index.

Only the sender knows the ephemeral secret, so only they can derive Zx per locking key (eP)

The receiver(s) only know the ephemeral pubkey (E) and their own secret key (p), so they can only generate the shared secret for their own key (pE).

robwoodgate:
> @robwoodgate I think we can avoid the `p2pk_e` entirely and instead use the `nonce` inside of secret more cleverly by setting it `nonce = e*G`.

Thinking about this some more this afternoon.... a reason to not do this: 

if the Mint knows a P2BK `E` might be held in the secret nonce, it could possibly discriminate against P2BK proofs in some way by checking if the nonce is a valid curve x-coordinate.

robwoodgate:
> Though this could be easily fixed if newer wallets always use EC public keys as nonces, even for normal p2pk.

That would alleviate the discrimination concern for sure. It would also go some way to alleviating the related concern that using the `secret.nonce` would reveal ALL `E`'s to the mint, not just those in proofs posted publicly.

xx.md
```diff
@@ -0,0 +1,232 @@
...
+
+> [!NOTE]
+> Reminder, the shared secret is unique to each receiver public key (`P`), making it the primary blinding factor. The slot index adds additional uniqueness to ensure that if the same receiver public key appears more than once (eg: as a locking AND refund key), it is blinded uniquely.
+
+## Receiver Workflow
+
+1. Read `E` from `proof.p2pk_e`, `keyset_id` from `proof.id`, and the key slot order index `i` from `[data, ...pubkeys, ...refund]`
+2. Calculate your unique shared secret: `Zx = x(p·E)`
+3. For each slot `i`, compute: \
+   a. Blinding scalar: `rᵢ = H("Cashu_P2BK_v1" || Zx || keyset_id || i) mod n` \
```
a1denvalu3: We should be more precise about the format of the stuff that goes in the hash function:
1. `Zx` is raw bytes or the utf-8 encoded hex representation of the bytes?
2. same for `keyset_id`.
3. Is `i` 1 byte or is it a radix-10 utf-8 string

a1denvalu3: Calculating the remainder of the division by `n` requires switching to a representation on a multiple precision library (bigint) which are unfit to do operations on secret values and could also potentially introduce a new dependency on some implementations that are not already relying on them.

I would suggest rejection sampling instead:

1. Sample `rᵢ = H("Cashu_P2BK_v1" || Zx || keyset_id || i || counter)`.
2. Instantiate a scalar. If it fails increase the counter and repeat step 1.

robwoodgate: The Determinism section specifies raw bytes for hash inputs but yes, it could be spelled out here too. 

We should also add a full worked example including secret keys to ensure implementations can test against it.

robwoodgate: Mod n is unavoidable in the derivation step, so not sure this would help much?

That said, there is a retry built in (again in the Determinism section) so can be easily done.

In any case, we should ensure the NUT makes the details laid out in the Cashu-TS reference implementation explicit.

a1denvalu3: > Mod n is unavoidable in the derivation step, so not sure this would help much?

If you meant when deriving `k = p + r (mod n)` then yes... **BUT**. The secp256k1 library (and relative rust bindings, python bindings etc...) have a notion of secret key and scalars, while noble secp (typescript) does not. In cashu-ts you are **FORCED** to use BigInt for secret key tweaking operations, but that doesn't mean it's the case for all other implementations.

a1denvalu3: > That said, there is a retry built in (again in the Determinism section) so can be easily done.

I think it should be laid out here in the derivation scheme by adding a counter at the end.

robwoodgate: The `mod n` is to indicate it is a valid point on the curve. In the cashu-ts reference implementation, I allowed for a single retry in case `r`=0. This is what the Determinism section references: 

```
If `rᵢ = 0`, retry once with an extra `0xff` byte appended to the hash input; abort if still zero.
```

```ts
const Zx = point.multiply(scalar).toBytes(false).slice(1, 33);
const iByte = new Uint8Array([slotIndex & 0xff]);
// Derive deterministic blinding factor (r):
let r = secp256k1.Point.Fn.fromBytes(sha256(Bytes.concat(P2BK_DST, Zx, keysetId, iByte)));
if (r === 0n) {
// Very unlikely to get here (1/n)!
r = secp256k1.Point.Fn.fromBytes(
	sha256(Bytes.concat(P2BK_DST, Zx, keysetId, iByte, new Uint8Array([0xff]))),
);
if (r === 0n) throw new Error('P2BK: tweak derivation failed');
}
return r;
```

a1denvalu3: @robwoodgate 

An invalid key can be `key = 0` or `key >= N`. So it's not really $\frac{1}{N}$. I think it makes most sense to add a counter (not just a 0xFF byte) and retry up to 16 times (just like hash-to-curve does)

```rust
if canonical_slot > 10 {
   return Err(Error::InvalidCanonicalSlot(canonical_slot));
}

// Compute shared point Z = ephemeral_key * pubkey
// Use SharedSecret if available (produces 32 bytes typically equal to x-coordinate)
let shared = SharedSecret::new(pubkey, secret_key);

// SharedSecret exposes 32 bytes (x-coordinate) — ensure this matches your crate
let z_x: [u8; 32] = shared.as_ref().try_into().expect("shared len == 32");

// Build KDF input: domain tag || x-only(Z) || keyset_id (4 bytes BE) || canonical_slot (1 byte) || counter (1 byte)

let mut engine = Sha256::engine();
engine.input(b"Cashu_P2BK_v1");
engine.input(&z_x);
engine.input(&keyset_id.to_bytes());
engine.input(&[canonical_slot]);


// I have to initialize this with a random key otherwise rustc can't figure out
// that there is no possible way this is left uninitialized.
let mut result: Result<SecretKey, Error> = Ok(SecretKey::generate());
for counter in 0..16 {
    let mut engine_clone = engine.clone();
    engine_clone.input(&[counter as u8]);
    let digest = Sha256::from_engine(engine_clone);

    result = SecretKey::from_slice(digest.as_byte_array()).map_err(Error::from);
    match result {
        Ok(_) => break,
        Err(_) => {},
    }
}
```

I think this is what makes most sense.

robwoodgate: Cool, yes using the hash-to-curve counter idea works for me. Want to PR some amends to incorporate this and polish the spec?

a1denvalu3: @robwoodgate I've opened a PR [here](https://github.com/robwoodgate/nuts/pull/2/files) with some small changes I felt necessary.

robwoodgate:
> To summarize the dilemma:
> 

Discussed again off proposal: The general feeling was to go with `Proof.p2pk_e` because:

- Specifying NUT-10 nonces SHOULD be SEC1 compressed pubkeys would be a significant change to protocol
- Precluding P2BK with SIG_ALL would be a big loss 
- Using the nonce field as a critical data carrier feels wrong ("Explicit is better than implicit")

Overall, reason 2 (loss of SIG_ALL compatibility) was seen as the main reason to NOT use the nonce as the carrier.

xx.md
robwoodgate: Resolved - PR was merged, and we removed the requirement to mod n the blinding tweak, preferring instead to check the range for broader compatibility.

lollerfirst:
@robwoodgate We could simplify the parity detection on the receiver side if we compared the x-only of the unblinded public key:
```P' - r*G = -/+ p*G + r*G - r*G = -/+ p*G = -/+ P```
```Compare: x-only(P) == x-only(p*G)```

**Comment by lollerfirst** [comment]
> It's not much of a simplifcation, as the blinded private key still needs to be derived in any case.

```
0. Pre-compute the public key: p*G
1. Derive unblinded P from P' and r: P = P' - r*G
2. See if the x-only matches: x(P) == x(p*G).
    2.a If it does, the first byte of the respective SEC1s tell you whether p or -p is to be used. 
```

To me, this sounds like a semplification.

File: 26.md [hunk 3]
```diff
@@ -0,0 +1,288 @@
...
+  "s": bool <optional>,
+  "m": Array[str] <optional>,
+  "d": str <optional>,
+  "t": Array[Transport] <optional>,
+  "nut10": NUT10Option <optional>,
+  "nut26": bool <optional>,
+}
+```
+
+Where `nut26` indicates whether the receiver supports Pay-to-Blinded-Key and requests the sender to blind the receiver's public keys specified in the `nut10.d` field, as well as any `pubkeys` and `refund` tags in the `nut10.t` field.
```
robwoodgate: I wonder if the `nut26` field should be defined directly inside NUT-18 (as the NUT-10 field is).

Otherwise it can be missed by implementors, if it is only in this NUT.

a1denvalu3: I think it should stay here. Otherwise by the same logic we should modify NUT-00 to include `p2pk_e` in the `Proof` and `Token` objects.

18.md
```diff
@@ -41,6 +43,11 @@ Here, the fields are
 - `t`: The method of `Transport` chosen to transmit the payment (can be multiple, sorted by preference)
 - `nut10`: The required [NUT-10][10] locking condition
 
+> [!NOTE]
+> The payment request object may be extended by other NUTs. These include:
+>
+> - [NUT-26][26]: Pay-to-Blinded-Key - adds `"nut26": bool <optional>,` to signal support.
+
```
thesimplekid: I hate to even suggest this but do we think there will be other use cases in the future that will want to extend the payment request? Would it make sense to add a tags field that is an array of array of strings this would be more consistent with NUT10 where we have the tags in the secret.

26.md
```diff
@@ -0,0 +1,292 @@
...
+> to the even-Y lift. Two interoperable strategies are allowed:
+>
+> 1. **Unblind, then select by parity (RECOMMENDED):** \
+>    a. compute `Rᵢ = rᵢ·G` \
+>    b. unblind `P = P' − Rᵢ` \
+>    c. verify `x(P) == x(p·G)` \
+>    d. use standard derivation if `parity(P) == parity(p·G)`, otherwise use negated derivation
+> 2. **Double-derive and match:** \
+>    a. derive both standard and negated candidates \
+>    b. select the one whose public key reconstructs `P'`
```
thesimplekid: Is there a reason to have both?

18.md
thesimplekid: I see the comment here. I'm okay if we keep it as it just gonna leave my comment about the future extension path.

26.md
robwoodgate: Sadly yes... if we don't want NOSTR users hating Cashu.

The problem is that BIP-340 doesn't require secret keys to be normalized. It only requires the pubkey to be lifted to X-Parity. So a Nostr user with a secret key that derives a natural Y-parity pubkey will not be able to derive a working secret key unless that key was normalized for them.

For regular SEC1 pubkey  it doesn't matter - the parity is always the natural one.

robwoodgate:
Added cross references to NUT-14 (HTLC) and noted that the `data` slot (hash) is not to be blinded for HTLC proofs.

18.md
callebtc: Agreed, I think this deserves its own PR. @robwoodgate maybe we can edit to NUT-18 after a separate discussion?

robwoodgate: I have reversed the changes to NUT-18.

The new NUT-26 still talks about extending it, so we will need to be updated in the new PR  if the signalling schema changes for payment request.

26.md
```diff
@@ -0,0 +1,296 @@
+# NUT-26: Pay-to-Blinded-Key (P2BK)
+
+`optional`
+
+`depends on: NUT-11, NUT-18`
```
callebtc:
```suggestion
`depends on: NUT-11`
```

26.md
```diff
@@ -0,0 +1,296 @@
...
+{
+  "i": str <optional>,
+  "a": int <optional>,
+  "u": str <optional>,
+  "s": bool <optional>,
+  "m": Array[str] <optional>,
+  "d": str <optional>,
+  "t": Array[Transport] <optional>,
+  "nut10": NUT10Option <optional>,
+  "nut26": bool <optional>,
```
callebtc:
```suggestion
  "nut26": NUT26Option <optional>,
```

callebtc: something like this has a better upgradeability

callebtc (changes requested):
I suggest that for future extensions, the setting in the NUT-18 PR should be a JSON.

callebtc:
I applied changes in [f16c9c4](https://github.com/cashubtc/nuts/pull/300/commits/f16c9c4a770ff7af938cb92ccfd5b10929dbc744) before realizing that the NUT-18 specs have moved from the NUT-18 document into the NUT-26 doc. I assumed we're going to define that in a separate PR. I made suggestions because the way it's suggested here will be hard to upgrade in the future. I think it should be left for another day and deserves more thought.

Large parts of the spec read AI generated. **Please remove all AI parts.** The spec could probably be compressed by 50% by taking out the slop.

26.md
```diff
@@ -0,0 +1,296 @@
...
+- **Multi-key tolerance:** in multi-signature or multi-key proofs, receivers **MAY** skip non-matching `P'` for a given private key without error, and proceed to sign any slots that do match.
+- **Caching:** wallets MAY precompute and cache `p·G` and its parity per receiver key to avoid repeated multiplications during spends, without affecting security.
+
+## Compatibility
+
+- NUT-11 protocol is unchanged.
+- Mints see only pure NUT-11 Proofs. The new `p2pk_e` field in Proof is stripped before sending to the mint.
+- Wallets can verify ECDH derivation locally; no mint changes are required.
+- Wallets that do not yet support P2BK will not leak privacy by sending P2BK Proofs to the mint. The spend will simply fail.
+- When used alongside [NUT-14][14] (HTLCs), the first slot (`data`) MUST NOT be blinded.
```
callebtc: redundant

26.md
```diff
@@ -0,0 +1,296 @@
...
+
+## Worked Example
```
callebtc: this should probably be a test vector

26.md
```diff
@@ -0,0 +1,296 @@
...
+- `keyset_id_bytes` is the full Keyset ID from the mint, **hex-decoded** to raw bytes.
+- `i_byte` is a single unsigned byte from 0 to 10 inclusive (`0x00` to `0x0A`).
+- If `rᵢ = 0` or `rᵢ >= n`, retry once with an extra `0xff` byte appended to the hash input; abort if still out of range.
+- All receiver keys **MUST** be in compressed SEC1 format (33 bytes) before ECDH and blinding. The sender **MUST add an '02' prefix** to BIP-340 x only pubkeys (eg Nostr).
+
+## Security Considerations
+
+- **Mint privacy:** the mint never sees `P`, `E`, or `rᵢ`.
+- **Freshness:** each proof uses a new `e`; reuse leaks linkage.
+- **Replay safety:** `keyset_id` prevents cross-mint reuse, `i` slot binding ensures each `rᵢ` is unique.
```
callebtc: not sure if this fits the spec.

26.md
```diff
@@ -0,0 +1,296 @@
...
+
+Where `nut26` indicates whether the receiver supports Pay-to-Blinded-Key and requests the sender to blind the receiver's public keys specified in the `nut10.d` field, as well as any `pubkeys` and `refund` tags in the `nut10.t` field.
+
+### Sender behavior
+
+When a payment request includes a P2PK locking condition in the `nut10` field with `nut26: true`, the sender SHOULD:
+
+1. Blind the receiver's public key(s) specified in the `nut10.d` field, as well as any `pubkeys` and `refund` tags in the `nut10.t` field, before minting/swapping
+2. Include the ephemeral public key `E` in the `p2pk_e` (`pe` for Token V4) field of the resulting proofs
+3. Send the P2BK-enhanced proofs to the receiver
```
callebtc: "P2BK-enhanced"?

26.md
robwoodgate: Removed all NUT-18 references for now. Let's rethink that part

robwoodgate:
> I applied changes in [f16c9c4](https://github.com/cashubtc/nuts/pull/300/commits/f16c9c4a770ff7af938cb92ccfd5b10929dbc744) before realizing that the NUT-18 specs have moved from the NUT-18 document into the NUT-26 doc. I assumed we're going to define that in a separate PR. I made suggestions because the way it's suggested here will be hard to upgrade in the future. I think it should be left for another day and deserves more thought.
> 
> Large parts of the spec read AI generated. **Please remove all AI parts.** The spec could probably be compressed by 50% by taking out the slop.

Have totally re-written from scratch to remove all the cruft and NUT-18 references.

robwoodgate:
@callebtc - can you re-review when you get some time please.

00.md
```diff
@@ -260,6 +265,11 @@ The deserialized `base64_token_cbor` is a JSON of the same form as a TokenV4 but
 
 Note that all fields of the `bytes` type encode hex strings in the original representation of `Proof`'s.
 
+> [!NOTE]
+> The token format may optionally be extended by other NUTs. These include:
+>
+> - [NUT-26][26]: Pay-to-Blinded-Key - adds `"pe": bytes` to individual proofs
```
SatsAndSports: Shouldn't `pe` here be `p2pk_e`?

```suggestion
> - [NUT-26][26]: Pay-to-Blinded-Key - adds `"p2pk_e": bytes` to individual proofs
```

robwoodgate: No, the token v4 format uses abbreviated key names. See NUT-26 - "Proof Object Extension" section:

```
For Token V4 encoding, the `p2pk_e` field is named `pe`, and `E` is encoded as a 33 byte CBOR bstr
```

26.md
```diff
@@ -0,0 +1,183 @@
...
+Per NUT-11, there are up to 11 locking 'slots' in the order: `[data, ...pubkeys, ...refund]`.
+
+Slot 0 is the `data` tag. Slots 1-10 can be any combination of `pubkeys` and `refund` keys.
+
+Each public key in the NUT-11 proof is permanently blinded using a deterministic blinding scalar (`rᵢ`), where `i` is the _slot index_.
+
+The blinding scalar for each slot is calculated as:
+
+```
+rᵢ = SHA-256( DOMAIN_SEPARATOR || Zx || keyset_id_bytes || i_byte)
```
SatsAndSports: Do we need to include the keyset here?

I ask because of a problem in the Spilman Channel, where I think I'd like to remove the keyset_id from this. I have the blinding fully working in the channel today, but not (yet) exactly following the system described in this NUT.

_TLDR: If a SIG_ALL swap locks output to a blinded pubkey, and the swap then sends the outputs to an unexpected keyset, then the unblinding won't work because the keyset is wrong. To clarify: I'm not talking about the inputs to this swap, I'm talking about the outputs of the swap, where the outputs are blinded P2BK_

If I sign a two-party transaction with SIG_ALL then it commits to the amounts in the outputs, but it does _not_ commit to the keyset in those outputs. Imagine the second party adds their signature and performs the swap. The second party decides what keyset those outputs will be in, and this might not be the keyset that I was expecting. They are free to choose any active keyset with the right unit.

If the outputs commit to blinded pubkeys, then things will be strange. I computed the blinded pubkey on one keyset, because I assumed the swap would create outputs in the same keyset as the inputs to the swap. But then the outputs are in a different keyset, and I would need to write special code to unblind and sign those outputs; special code which uses the 'expected keyset' instead of the actual keyset

robwoodgate: The keyset ID ensures the blinding factor is unique between mints (and keysets) to avoid privacy leakage in case of ephemeral key reuse.

SatsAndSports: ~How about (instead) including the secret's `nonce` in this derivation?~

SatsAndSports: > How about (instead) including the secret's `nonce` in this derivation?

actually, ignore that particular question from me about the `nonce`. It would break SIG_ALL, which requires all the proofs to have the same (blinded) public key

SatsAndSports: I would still like to remove the keyset_id from the derivation, for the reason described at the start of this thread

robwoodgate: @callebtc, @a1denvalu3, @thesimplekid  - thoughts?

a1denvalu3: @robwoodgate I think we can make this sacrifice, because  -if there is nonce re-use- putting the keyset ID in the derivation only protects in the case the nonce reuse is across different keysets

a1denvalu3: Realistically we're not going to have nonce re-use.

robwoodgate:
> I think we will now need to change the number of the NUT because bech got merged first.

Yeah 🤦‍♂️

Done.

26.md
robwoodgate: If there is nonce reuse, that is _"dOInG-iT-WroNG"_, so yeah, we probably don't need to protect against that.

robwoodgate (commented):
Have removed keyset ID from the blinding factor calculation, re-calculated all test vectors and renamed to NUT-28 (as BECH32 got merged first).

26.md
robwoodgate: Have removed keyset id as requested

robwoodgate:
Rebaed to main to resolve a merge conflict.

00.md
```diff
@@ -95,6 +95,11 @@ A `Proof` is also called an _input_ and is generated by `Alice` from a `BlindSig
 
 `amount` is the amount of the `Proof`, `secret` is the secret message and is a utf-8 encoded string (the use of a 64 character hex string generated from 32 random bytes is recommended to prevent fingerprinting), `C` is the unblinded signature on `secret` (hex string), `id` is the [keyset id][02] of the mint public keys that signed the token (hex string).
 
+> [!NOTE]
+> A proof may optionally be extended by other NUTs. These include:
+>
+> - [NUT-28][28]: Pay-to-Blinded-Key - adds `"p2pk_e": hex_str` to store an ephemeral pubkey
+
```
callebtc:
```suggestion
```
I suggest to removing this, or we would have to add all possible extensions here (witness for example)

00.md
```diff
@@ -277,6 +282,11 @@ If a short keyset ID resolves to more than one known full keyset ID, the identif
 
 The mint is unaware of the `s_id`. All API endpoints exposed by the mint use the full keyset ID.
 
+> [!NOTE]
+> The token format may optionally be extended by other NUTs. These include:
+>
+> - [NUT-28][28]: Pay-to-Blinded-Key - adds `"pe": bytes` to individual proofs
+
```
callebtc:
```suggestion
```

00.md
```diff
@@ -347,3 +357,4 @@ utf8("craw") || utf8(<token_version>) || <serialised_token>
 [10]: 10.md
 [11]: 11.md
 [12]: 12.md
+[28]: 28.md
```
callebtc:
```suggestion
```

28.md
```diff
@@ -0,0 +1,178 @@
...
+
+---
+
+## Summary
+
+This NUT describes Pay-to-Blinded-Key (P2BK), which extends the [NUT-11][11] (P2PK) spending conditions. By implication, it also extends [NUT-14][14] (HTLC).
+
+P2BK preserves privacy by blinding each NUT-11 receiver pubkey `P` with an ECDH-derived scalar `rᵢ`. Both sides can deterministically derive the same `rᵢ` from their own keys, but a third party cannot.
+
+This brings _"silent payments"_ to Cashu: Proofs can be locked to a well known public key, posted in public without compromising privacy, and spent by the recipient without needing any side-channel communication.
```
callebtc:
```suggestion
```

28.md
```diff
@@ -0,0 +1,178 @@
...
+
+`depends on: NUT-11`
+
+---
+
+## Summary
+
+This NUT describes Pay-to-Blinded-Key (P2BK), which extends the [NUT-11][11] (P2PK) spending conditions. By implication, it also extends [NUT-14][14] (HTLC).
+
+P2BK preserves privacy by blinding each NUT-11 receiver pubkey `P` with an ECDH-derived scalar `rᵢ`. Both sides can deterministically derive the same `rᵢ` from their own keys, but a third party cannot.
```
callebtc:
```suggestion
P2BK preserves privacy by blinding each NUT-11 receiver pubkey `P` with an ECDH-derived scalar `rᵢ`. Both sides can deterministically derive the same `rᵢ` from their own keys, but a third party cannot. This improves user privacy by preventing the mint from linking multiple P2PK spends by the same party.
```

28.md
```diff
@@ -0,0 +1,178 @@
...
+
+This brings _"silent payments"_ to Cashu: Proofs can be locked to a well known public key, posted in public without compromising privacy, and spent by the recipient without needing any side-channel communication.
+
+## ECDH Shared Secret (Zx)
+
+ECDH allows two parties to create an x-coordinate shared secret (`Zx`) by combining their private key with the public key of the other party: `Zx = x(epG) = x(eP) = x(pE)`.
```
callebtc:
```suggestion
Elliptic-curve Diffie–Hellman (ECDH) allows two parties to create an x-coordinate shared secret (`Zx`) by combining their private key with the public key of the other party: `Zx = x(epG) = x(eP) = x(pE)`.
```

callebtc: we should spell it out once before abbreviating

28.md
```diff
@@ -0,0 +1,178 @@
...
+
+For P2BK, the sender creates an ephemeral keypair (private key: `e`, public key: `E`). This protects the privacy of their own long-lived public key. They then calculate the shared secret by combining the ephemeral private key (`e`) and the receiver's long-lived public key (`P`).
```
callebtc: > This protects the privacy of their own long-lived public key.

we didn't define what the long-lived public key is yet (next sentence). we only speak about the receiver's pubkey but the sentence sounds like it protects the sender's pubkey. suggest to just remove the sentence.

callebtc:
```suggestion
For P2BK, the sender creates an ephemeral keypair (private key: `e`, public key: `E`). They then calculate the shared secret by combining the ephemeral private key (`e`) and the receiver's long-lived public key (`P`).
```

28.md
```diff
@@ -0,0 +1,178 @@
...
+
+The receiver calculates the same shared secret using their private key (`p`) and the ephemeral public key (`E`), which is supplied by the sender in the [proof metadata](#proof-object-extension).
```
callebtc:
```suggestion
The receiver calculates the same shared secret `Zx` using their private key (`p`) and the ephemeral public key (`E`), which is supplied by the sender in the [proof metadata](#proof-object-extension).
```

28.md
```diff
@@ -0,0 +1,178 @@
...
+
+This shared secret is then used to derive the blinded public keys.
```
callebtc:
```suggestion
The shared secret `Zx` is then used to derive the blinded public keys.
```

28.md
```diff
@@ -0,0 +1,178 @@
...
+```
+
+Where:
+
+- `DOMAIN_SEPARATOR` constant byte string `b"Cashu_P2BK_v1"`
+- `Zx` is the ECDH shared secret (`eP` for sender, `pE` for receiver).
+- `i_byte` is the single unsigned byte representation of `i`: (`0x00` to `0x0A`)
+- `||` denotes concatenation
+
+For broader compatibility, `rᵢ` **MUST NOT** be normalised modulo `n`
```
callebtc:
```
For broader compatibility, `rᵢ` **MUST NOT** be normalised modulo `n`
```

I know this issue (backwards compat) but why not? Every private key should be normalized.

28.md
```diff
@@ -0,0 +1,178 @@
...
+
+If `rᵢ` is not in the range `1 ≤ rᵢ ≤ n−1`, retry once with an extra `0xff` byte appended to the hash input as follows:
+
+```
+rᵢ = SHA-256( b"Cashu_P2BK_v1" || Zx || i_byte || 0xff )
+```
+
+If `rᵢ` is still not in the range `1 ≤ rᵢ ≤ n−1`, abort and discard the ephemeral keypair.
```
callebtc, line 52: It would allow us to delete this part

28.md
```diff
@@ -0,0 +1,178 @@
...
+
+Finally, the public key (`P`) for slot `i` is blinded (`P'`) as follows:
+
+```
+P' = P + rᵢG
+```
+
+Here is a code example in TypeScript:
```
callebtc:
```suggestion
### Example

Below is an example implementation in TypeScript.
```

28.md
```diff
@@ -0,0 +1,178 @@
...
+  if (r === 0n || r >= secp256k1.Point.CURVE().n) {
+    // Very unlikely to get here!
+    r = bytesToNumber(
+      sha256(Bytes.concat(P2BK_DST, Zx, iByte, new Uint8Array([0xff]))),
+    );
+    if (r === 0n || r >= secp256k1.Point.CURVE().n) {
+      // Astronomically unlikely to get here!
+      throw new Error("P2BK: tweak derivation failed");
+    }
+  }
```
callebtc, line 85: suggest: do not use own normalization, let the crypto lib handle it?

28.md
```diff
@@ -0,0 +1,178 @@
...
+
+Each proof adds a single new metadata field:
+
+```jsonc
+{
+  "amount": int,
+  "id": hex_str,
+  "secret": str,          // still ["P2PK", {...}]
+  "C": hex_str,
+  "p2pk_e": hex_str       // 33-byte SEC1 compressed ephemeral public key E
```
callebtc:
```suggestion
  "p2pk_e": hex_str       // NEW: 33-byte SEC1 compressed ephemeral public key E
```

28.md
```diff
@@ -0,0 +1,178 @@
...
+- For Token V4 encoding, the `p2pk_e` field is named `pe`, and `E` is encoded as a 33 byte CBOR bstr
+
+## Deriving Private Keys
+
+With P2BK, the NUT-11 public locking keys are permanently blinded. The mint sees only the blinded public keys, and expects signatures from the corresponding private key.
+
+The receiver must therefore derive the correct blinded private key. Because BIP-340 lifts public keys to even-Y parity, there are two possible derivation paths:
+
+- Standard derivation: `k = (p + rᵢ) mod n`
+- Negated derivation: `k = (-p + rᵢ) mod n`
```
callebtc, line 122: `k` is undefined

28.md
robwoodgate: I agree. I specced it modulo n in the intial draft, but received feedback that RUST (CDK) would have problems doing modulo n calculations.

So that's why it was re-drafted as _"tweak-and-single-retry, abandon ephemeral key if still out of range"_

Modulo n is the far simpler implementation, if it can be supported consistently.

That's the backwards compatibility issue. Happy to go either way - just make the call.

00.md
robwoodgate: My concern was that a reader has to know all the NUTS to understand the possible shape of a proof. But happy to remove.

28.md
```diff
@@ -0,0 +1,180 @@
...
+- `p2pk_e` contains the sender's ephemeral pubkey (`E`) used for blinding
+- All pubkeys inside the `"P2PK"` secret are the blinded forms `P'`
+- The mint sees standard P2PK data and remains unaware of the blinding
+- For Token V4 encoding, the `p2pk_e` field is named `pe`, and `E` is encoded as a 33 byte CBOR bstr
+
+## Deriving Private Keys
+
+With P2BK, the NUT-11 public locking keys are permanently blinded. The mint sees only the blinded public keys, and expects signatures from the corresponding private key.
+
+The receiver must therefore derive the correct blinded private key. Because BIP-340 lifts public keys to even-Y parity, there are two possible derivation paths:
```
robwoodgate:
```suggestion
The receiver must therefore derive the correct blinded private key (`k`). Because BIP-340 lifts public keys to even-Y parity, there are two possible derivation paths:
```

28.md
robwoodgate: If RUST can handle the `modulo n`, it's a thumbs up from me. (@thesimplekid )

28.md
robwoodgate: The idea here was to highlight to sender that using an ephemeral keypair protects their own (long lived) public key from being linked to the proofs.

We can remove the sentence, but I think it's an important feature of P2BK... it provides privacy for both sender AND receiver.

robwoodgate: That's exactly right - using an ephemeral keypair does protect the senders pubkey :)

28.md
SatsAndSports: modulo `n` could technically give us `0` right?

It's so unlikely, I guess we could ignore it

But if we want to avoid that unlikely event, we could do modulo (n-1) and then add one

i.e. `r = 1 + sha256(...) % (n-1)` , where `%` is modulo, in order to ensure that we get a number between `1` and `n-1` (inclusive). Although hopefully the cryptography library does something like this already

robwoodgate: I believe the usual practice is to do standard modulo n and discard/retry if zero.

28.md
robwoodgate: We agreed offline to keep tweak and retry, as not all implementations can handle modulo n reduction

28.md
robwoodgate: We agreed offline to keep tweak and retry, as not all implementations can handle modulo n reduction

28.md
robwoodgate: clarified

28.md
robwoodgate: We agreed offline to keep tweak and retry, as not all implementations can handle modulo n reduction

PR #311, opened 2025-11-19
NUT-12 DLEQ: 'r' refers to the mint's random nonce, and also to Alice's blinding factor. Rename one of them to avoid confusion

SatsAndSports:
_Update: 2025-11-25: instead of the note, this PR now just renames one of the variables, so that we no longer have two different variables with the same name_

In NUT-00, `r` refers to the _blinding factor_ which Alice's uses to blind and unblind. This is also used in this NUT-12 when Alice sends her blinding factor to Carol. This is _not_ shared with the mint.

But the DLEQ proof here in NUT-12 also reuses `r` to refer to the random nonce selected by Bob the mint, and Bob doesn't share this with Alice or Carol.

Two different variables with the same name (`r`) is confusing

This PR rename's Bob's (the mint) `r` to `m` to ensure that this isn't confused with Alice's `r`

### DIFF
--- a/12.md
+++ b/12.md
@@ -18,11 +18,11 @@ The complete DLEQ proof reads
 (These steps occur when Bob returns C')
 
 Bob:
-m = a random nonce, selected by Bob the mint and not shared
-R1 = m*G
-R2 = m*B'
+k = a random nonce, selected by Bob the mint and not shared
+R1 = k*G
+R2 = k*B'
 e = hash(R1,R2,A,C')
-s = m + e*a
+s = k + e*a
 return e, s
 
 Alice:


lollerfirst:
@SatsAndSports it's the same `r`. Bob doesn't decide `r`. Bob produces the DLEQ without it. `r` is sent to Carol so that she can compute $B' = Y + r\cdot G$

NACK

lollerfirst:
@SatsAndSports  I don't think there is any room for confusion, because in the context of the creation of the ZK proof, Bob can't possibly refer to Alice's `r` as he doesn't know it. The blinding factor is not a concept for the Mint, at all. It only exists for the wallet.

12.md
```diff
@@ -18,11 +18,11 @@ The complete DLEQ proof reads
 (These steps occur when Bob returns C')
 
 Bob:
-r = random nonce
-R1 = r*G
-R2 = r*B'
+k = a random nonce, selected by Bob the mint and not shared
```
callebtc, line 21:
```suggestion
k = a random nonce
```
this is self-evident

PR #325, opened 2026-01-08
NUT-01: return keyset metadata

### LINKED ISSUE #289: /v1/keys endpoint should include input_fee_ppk
In #182, I raised a related point that the `/v1/keys` endpoint should ideally return ALL keyset fields required for active wallet operations. 
This allows a consumer to call just ONE endpoint and get data needed for active wallet ops - fees, expiry, keys etc etc

That way, the '/v1/keysets' endpoint can just be called when non-operational, historic or inactive keyset metadata is required.

@callebtc made the comment:

> Nevertheless: wallets should hit `/keysets` for metadata, and `/keys` for the actual keys.

I agree in principle - if there was a lot of incidental metadata - but in the v1 API, both are practically identical:

```ts
/v1/keysets
    {
      "id": <hex_str>,
      "unit": <str>,
      "active": <bool>,
      "input_fee_ppk": <int|null>,
      "final_expiry": <int|null>
    }

/v1/keys
    {
      "id": <keyset_id_hex_str>,
      "unit": <currency_unit_str>,
      "final_expiry": <unix_timestamp_int|null>
      "keys": {
        <amount_int>: <public_key_str>,
        ...
      }
    }
```

`active` is implied in /v1/keys - as only active keysets are returned.
Which means the only "missing" metadata is `input_fee_ppk`, which is vital to wallet operations.

It seems sub-optimal to make a consumer lookup two endpoints just to get the keyset fee, which is as as vital to wallet ops as `final_expiry`.

callebtc:
Expose `active` and `input_fee_ppk` fields on keysets returned from `/v1/keys` endpoint and align the `/v1/keys/{keyset_id}` example accordingly.

This allows wallets to compute the keyset ID and access associated keyset metadata directly from the `/keys` response without requiring an additional round-trip to the `/keysets` endpoint.

closes #289 

- [ ] CDK: https://github.com/cashubtc/cdk/pull/1502
- [ ] nutshell: https://github.com/cashubtc/nutshell/pull/868
- [ ] nutmix


callebtc:
> We probably need the `final_expiry` field too, right?

yes, that would have to wait for or be merged into the keyset v2 PR

PR #329, opened 2026-01-12
NUT-XX: get locked mint quotes by public keys

average-gary:
The primary use case I have in mind is ehash mining but examples use bolt11 LN for simplicity and broader knowledge of that flow for Cash protocol engineers.

### DIFF
--- a/28.md
+++ b/28.md
@@ -1,4 +1,4 @@
-# NUT-28: Third-Party Mint Quotes
+# NUT-28: Mint Quote Lookup by Public Key
 
 `optional`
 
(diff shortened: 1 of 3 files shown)


28.md
```diff
@@ -0,0 +1,320 @@
...
+
+```json
+{
+  "quote": <str>,
+  "request": <str>,
+  "state": <str_enum[STATE]>,
+  "expiry": <int>,
+  "pubkey": <str>
+}
+```
```
thesimplekid: This is the same as nut20 currently is, the mint does not need to know that its intended for a 3rd party.

28.md
```diff
@@ -0,0 +1,320 @@
...
+
+The `signature` follows the [NUT-20][20] message aggregation scheme:
+
+```
+msg_to_sign = quote || B_0 || ... || B_(n-1)
+```
+
+The mint verifies the signature against the `pubkey` stored with the quote and responds with blind signatures as in [NUT-04][04].
```
thesimplekid: This is the same as current melt flow I don't think we need to redefine it.

28.md
```diff
@@ -0,0 +1,320 @@
...
+Quote states follow [NUT-04][04]:
+
+- `UNPAID`: Quote created, Lightning invoice not yet paid
+- `PAID`: Invoice paid, waiting for recipient to mint
+- `ISSUED`: Tokens have been minted
+- `EXPIRED`: Quote expired before completion
+
+### Expiry policy
+
+Quotes **SHOULD** respect the keyset's `final_expiry` ([NUT-01][01]). Mints define their own retention policy for `PAID` quotes that have not been redeemed.
```
thesimplekid: Don't think this should be in this spec either.

28.md
```diff
@@ -0,0 +1,320 @@
...
+  }
+}
+```
+
+| Field | Type | Description |
+| --- | --- | --- |
+| `supported` | `bool` | Whether third-party quote lookup is supported |
+| `methods` | `str[]` | Payment methods supporting third-party quotes (e.g., `["bolt11"]`) |
+| `timestamp_tolerance_secs` | `int\|null` | Tolerance window for signature timestamps in seconds (default: 60) |
+| `quote_retention_days` | `int\|null` | How long `PAID` quotes are retained in days (`null` = indefinite) |
```
thesimplekid: We don't delete these.

28.md
```diff
@@ -0,0 +1,320 @@
...
+```json
+{
+  "type": "cashu_third_party_quote",
+  "mint": "https://mint.host:3338",
+  "method": "bolt11",
+  "pubkey": "03d56ce4e446a85bbdaa547b4ec2b073d40ff802831352b8272b7dd7a4de5a7cac",
+  "amount": 100,
+  "unit": "sat"
+}
+```
```
thesimplekid: This also seems like an application specific thing and not for the nut. Also should not use nut04

PR #331, opened 2026-01-19
Deterministic pay to public key generation

lescuer97:
Adds a standard way to generate keys for wallets to generate private keys for usage in P2PK operations. 

- [x] cdk:  https://github.com/cashubtc/cdk/pull/1466
- [x] coco: https://github.com/cashubtc/coco/pull/121
- [x] cashu-ts: https://github.com/cashubtc/cashu-ts/pull/697
- [ ] macadamia


13.md
```diff
@@ -39,6 +39,16 @@ The HMAC-SHA256 KDF is built as the following:
...
 
+#### P2PK Derivation
+
+Wallet are able to generate private keys to then have proofs be locked to them.
+
+The following path is used for derivation of those keys: `m/129372'/10'/0'/0'/{counter}`:
+
+- 129372': Registered SLIP-0044 coin type for Cashu.
+- 10': Purpose for generating private keys for usage in P2PK.
+- {counter}: Incrementing counter encoded as an unsigned 64-bit integer in big-endian format.
```
a1denvalu3: is this using BIP32? Should be more clear.

13.md
```diff
@@ -39,6 +39,16 @@ The HMAC-SHA256 KDF is built as the following:
 2. `hmac_digest = HMAC_SHA256(seed, message)`, where `HMAC_SHA256` is the [hash-based message authentication code](https://en.wikipedia.org/wiki/HMAC) using SHA-256 as the hashing algorithm.
 3. `secret = hmac_digest` and `blinding_factor = hmac_digest % N`.
 
+#### P2PK Derivation
+
+Wallet are able to generate private keys to then have proofs be locked to them.
+
+The following path is used for derivation of those keys: `m/129372'/10'/0'/0'/{counter}`:
```
a1denvalu3:
```
`m/129372'/10'/0'/0'/{counter}`:
```

not using per-keyset derivation. are we sure about this?

lescuer97: yes, because it's for locking up proofs  it should be more general

13.md
lescuer97: yes, I'll make it clearer

robwoodgate (approved):
Suggestion below to tighten the language, otherwise ACK.

robwoodgate (commented):
Suggestion to tighten language, otherwise concept ACK

13.md
```diff
@@ -39,6 +39,18 @@ The HMAC-SHA256 KDF is built as the following:
 2. `hmac_digest = HMAC_SHA256(seed, message)`, where `HMAC_SHA256` is the [hash-based message authentication code](https://en.wikipedia.org/wiki/HMAC) using SHA-256 as the hashing algorithm.
 3. `secret = hmac_digest` and `blinding_factor = hmac_digest % N`.
 
+#### P2PK Derivation
+
+Wallet are able to generate private keys is a deterministic way to have proofs locked to them.
+
+The following BIP32 derivation path for derivation of the key: `m/129372'/10'/0'/0'/{counter}`:
```
prusnak: Can you explain why the last element is not hardened?

Is the xpub of `m/129372'/10'/0'/0'` ever shared anywhere?

If it is not shared, then instead of hardening it, I propose we change the scheme to `Cashu_KDF_HMAC_SHA256` (same as used in Keyset v2, since BIP32 used in Keyset v1 is deprecated).

lescuer97: We are following what Bitcoin does when it comes to key derivation for locking.  

there is no reason right now for sharing an XPUB but can't guarantee that in the future. 

The reason we want  to use bip32 is because Bip32 is specifically made for this case.The `Cashu_KDF_HMAC_SHA256` scheme is used because of aggregation to avoid certain issues when generating the keyset id. 

This keys are never aggregated and are use individually.

prusnak: > The reason we want to use bip32 is because Bip32 is specifically made for this case.

No, it's not. BIP32 usecase is when you need to share a pubkey for a certain key subtree. Unless you need this requirement, going with `Cashu_KDF_HMAC_SHA256` is simply faster and easier.

Egge21M: @prusnak if we get the option to query the mint for quotes connected to a pubkey via NUT-20, public key derivation would become more meaningful than it currently is. I can not think of a usecase right now, but with this, extended keys would be able to essentially create a watch-only wallet of a wallets quotes, without spending from it.

So either we think about proper usecases for this for 2 weeks and go with HMAC if we can't find any, or we take the performance L, go with BIP32 and hope someone finds a usecase some day in the future

prusnak: if we want to have a watch-only wallet eventually which should show p2pk tokens too, this seems like a good use-case. ACK

13.md
```diff
@@ -39,6 +39,18 @@ The HMAC-SHA256 KDF is built as the following:
...
+
+- 129372': Registered SLIP-0044 coin type for Cashu.
```
thesimplekid: Think coco and cdk are not using 129372
```suggestion
The following BIP32 derivation path for derivation of the key: `m/129372'/10'/0'/0'/{counter}`:

- 129373': Registered SLIP-0044 coin type for Cashu.
```

13.md
```diff
@@ -39,6 +39,18 @@ The HMAC-SHA256 KDF is built as the following:
 2. `hmac_digest = HMAC_SHA256(seed, message)`, where `HMAC_SHA256` is the [hash-based message authentication code](https://en.wikipedia.org/wiki/HMAC) using SHA-256 as the hashing algorithm.
 3. `secret = hmac_digest` and `blinding_factor = hmac_digest % N`.
 
+#### P2PK Derivation
```
thesimplekid, line 42: I wonder of this should just be a new nut would be easier to track who supports it and think we should avoid changing existing nuts where we can especially widely supported ones like nut13. Though it would be a very small nut.

13.md
```diff
@@ -39,6 +39,18 @@ The HMAC-SHA256 KDF is built as the following:
...
+
+Wallet are able to generate private keys is a deterministic way to have proofs locked to them.
```
thesimplekid:
```suggestion
Wallet are able to generate private keys in a deterministic way to have proofs locked to them.
```

13.md
lescuer97: I thought about the same but I don't think it's needed in this case. Maybe if we start making more deterministic generation we would move it there.

13.md
lescuer97: Thank you, I forgot about changing this I also changed the tests to the correct ones.

robwoodgate (changes requested):
Honestly, I think we should use NUT-13's v2 HMAC-SHA256 KDF rather than introducing a new BIP32 path.

The KDF defined for v2 exists precisely because BIP32 path traversal is really expensive, and a P2PK private key only needs a uniform secp256k1 scalar, which the same single-HMAC + single-subtraction reduction (bias ~2⁻¹²⁸) gives you directly.

Quick numbers from a pure-JS noble benchmark (cashu-ts, per counter):

```
HMAC-SHA256                                       ~165 µs    1.0×
```

In a restore situation, the wallet can cache the parent path derivation and get close to parity with HMAC, but if using just a cached master key (for regular wallet ops), the slowdown is 5x. In other words, BIP-32 needs careful handling to avoid performance tanking. 

There is also no special need for BIP32 here: P2PK keys are derived wallet-side from a hot seed, so there's no hardware-wallet or chain-code benefit.

Mirroring the v2 construction — something like:

```
HMAC_SHA256(seed, "Cashu_KDF_HMAC_SHA256_NUT11" || counter_u64_be) reduced mod n
```

would keep a single KDF across all of NUT-13 and lines up with the direction v2 was already moving.

13.md
robwoodgate: The use case is double hypothetical - relying on wallets to export XPUBS and an endpoint that doesn't exist. BIP32 is a slowdown vs HMAC, a performance L to soak up forever.

And any lookup by pubkey would either be limited (pubkeys are SUPPOSED to be single use for privacy) or would likely allow the mint to link pubkeys (eg in a batch lookup scenario).

robwoodgate:
I have proposed a [HMAC-SHA256 KDF derived alternative](https://github.com/cashubtc/nuts/pull/384), which would close this PR

robwoodgate (approved):
Following offline discussion, and the fact this is already live, I've closed #384.

thesimplekid (approved):

robwoodgate:
Cashu-TS support now added:
- https://github.com/cashubtc/cashu-ts/pull/697

13.md
```diff
@@ -39,6 +39,18 @@ The HMAC-SHA256 KDF is built as the following:
 2. `hmac_digest = HMAC_SHA256(seed, message)`, where `HMAC_SHA256` is the [hash-based message authentication code](https://en.wikipedia.org/wiki/HMAC) using SHA-256 as the hashing algorithm.
 3. `secret = hmac_digest` and `blinding_factor = hmac_digest % N`.
 
+#### P2PK Derivation
+
+Wallet are able to generate private keys in a deterministic way to have proofs locked to them.
```
robwoodgate:
```suggestion
Wallets are able to generate private keys in a deterministic way to have proofs locked to them.
```

13.md
```diff
@@ -39,6 +39,18 @@ The HMAC-SHA256 KDF is built as the following:
...
+
+The following BIP32 derivation path for derivation of the key: `m/129373'/10'/0'/0'/{counter}`:
+
+- 129373': Purpose picked for P2PK derivation.
+- 10': Account for generating private keys for usage in P2PK.
+- {counter}: Incrementing counter encoded as an unsigned 64-bit integer in big-endian format.
+
+This will allow wallets to swap proof that are still locked to a public key during a restore process.
+
```
robwoodgate, line 53:
```suggestion

In line with BIP-32, if the resulting private key is out of range (`> N`), it should be discarded.
```

robwoodgate:
@ye0man - this is fully in production (CDK/Coco led, CTS recently), and has greens, so should be ready to merge baring acceptance of any final wording tweaks (I have one suggestion outstanding).

callebtc (approved):

13.md
```diff
@@ -39,6 +39,18 @@ The HMAC-SHA256 KDF is built as the following:
...
 
+#### P2PK Derivation
+
+Wallets are able to generate private keys in a deterministic way to have proofs locked to them.
+
+The following BIP32 derivation path for derivation of the key: `m/129373'/10'/0'/0'/{counter}`:
+
+- 129373': Purpose picked for P2PK derivation.
+- 10': Account for generating private keys for usage in P2PK.
+- {counter}: Incrementing counter encoded as an unsigned 64-bit integer in big-endian format.
```
robwoodgate:
```suggestion
- `{counter}` is an incrementing, non-hardened BIP32 child index.
```

BIP-32 counters are not u64, they are an unsigned integer in the range `0 <= counter < 2^31`

PR #337, opened 2026-02-07
NUTs for prediction markets

joemphilips:
This is my ongoing attempt to support encoding Prediction Market Token with cashu.
My primary motivation is to

- Emulate [Conditional Token Framework](https://conditional-tokens.readthedocs.io/en/latest/motivation.html) for flexible markets
- Support existing oracle infrastructure from [DLC](https://github.com/discreetlogcontracts/dlcspecs) as much as possible.

There have been previous attempts to support DLCs on Cashu, for example:
https://github.com/cashubtc/nuts/pull/128

The purpose of those attempts was essentially to emulate a two-party DLC transaction using eCash tokens.

This approach is completely different. Each outcome is encoded as a separate token, and the mint is agnostic to the oracles. I believe this is a more natural representation of a prediction-market security token and enables a more liquid market.

This PR contains three NUTs

- NUT-CTF: Conditional token redemption via oracle-attested outcome keysets
- NUT-CTF-split-merge: Conditional token split and merge for trading positions
- NUT-CTF-numeric: Numeric outcome conditions with digit decomposition

Done self-reviewing, it is now ready for review from others. I will try to have a reference implementation ready when I have time. I may update the spec based on what I've learn during implementation.

It is possible to split CTF.md (Conditional Token Framework) into an independent PR if the maintainer wishes. Since this is quite a huge PR.

Reference implementation: https://github.com/cashubtc/cdk/pull/1666

### DIFF
--- a/CTF.md
+++ b/CTF.md
@@ -1,0 +1,533 @@
+# NUT-CTF: Conditional Token Framework
+
+`optional`
+
+`depends on: NUT-02, NUT-06`
+
+---
+
+This NUT defines conditional tokens and conditional keysets for oracle-attested events.
+
+A **conditional token** is a regular Cashu token ([NUT-00][00]) signed under a conditional keyset. It can be transferred and swapped like any other Cashu token, with one additional ability: it can be redeemed for regular ecash via `POST /v1/redeem_outcome` by providing a DLC oracle's attestation signature as a witness.
+
+A **conditional keyset** is a per-outcome-collection signing keyset ([NUT-02][02]) that the mint creates during partition registration. Each outcome collection gets a unique keyset with different signing keys.
+
+The oracle signature scheme is compatible with the [DLC specification](https://github.com/discreetlogcontracts/dlcspecs/blob/master/Oracle.md), allowing Cashu mints to leverage existing DLC oracle infrastructure.
+
+Caution: Applications that rely on oracle resolution must verify that the oracle is trustworthy and check via the mint's [info][06] endpoint that NUT-CTF is supported.
+
+**Related specifications:** [NUT-CTF-split-merge][CTF-split-merge] defines split/merge operations for creating and dissolving complete sets of conditional tokens. [NUT-CTF-numeric][CTF-numeric] extends this framework with numeric outcome conditions.
+
+## Terminology
+
+- **Condition**: A question with defined outcomes, resolved by an oracle. Identified by a `condition_id`. Equivalent to "condition" in the [Gnosis Conditional Token Framework](https://docs.gnosis.io/conditionaltokens/).
+- **Outcome**: A single atomic result that an oracle attests to (e.g., `"YES"`, `"ALICE"`).
+- **Outcome collection**: A subset of outcomes, defined by a partition element (e.g., `"YES"`, `"ALICE|BOB"`). Each gets its own conditional keyset. Redeemable if the oracle attests to ANY outcome it contains.
+- **Partition**: A division of all outcomes into disjoint, complete outcome collections.
+- **Condition ID** (`condition_id`): 32-byte tagged hash uniquely identifying a condition. Partition-independent. See [Condition ID](#condition-id).
+- **Outcome collection ID** (`outcome_collection_id`): 32-byte x-only public key uniquely identifying an outcome collection within a condition. See [Outcome Collection ID](#outcome-collection-id).
+
+## Outcome Collections
+
+Outcome collections allow tokens to represent one or more outcomes. An outcome collection is either a single outcome or an OR-combination joined by `|` (e.g., `"ALICE|BOB"` = "Alice or Bob wins"). If an outcome name contains `|`, it MUST be escaped as `\|`.
+
+### Partition Rules
+
+Partition keys MUST form a valid partition of all outcomes:
+
+1. **Disjoint**: No outcome appears in multiple outcome collections
+2. **Complete**: Every outcome appears in exactly one outcome collection
+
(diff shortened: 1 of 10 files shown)


joemphilips:
Reference implementation in [CDK(WIP)](https://github.com/cashubtc/cdk/pull/1648)
and [cashu-ts](https://github.com/cashubtc/cashu-ts/pull/682)

joemphilips:
In case anyone's interested, I've been building a PoC application based on this protocol. 

- Mockup website (very much a WIP): https://frontend-bitcaster-staging.azurewebsites.net/markets
- documentation: https://bitcasterdoc.com/user-guide/getting-started/overview/

CTF.md
```diff
@@ -0,0 +1,706 @@
...
+[09]: 09.md
+[10]: 10.md
+[11]: 11.md
+[12]: 12.md
+[14]: 14.md
+[21]: 21.md
+[22]: 22.md
+[CTF-split-merge]: CTF-split-merge.md
+[CTF-numeric]: CTF-numeric.md
+[NIP-88]: https://github.com/nostr-protocol/nips/pull/1681
```
joemphilips, line 534: This link should be replaced with an actual NIP after https://github.com/nostr-protocol/nips/pull/1681 gets merged.

joemphilips:
squashed and rebased

CTF.md
```diff
@@ -0,0 +1,533 @@
...
+
+The oracle signature scheme is compatible with the [DLC specification](https://github.com/discreetlogcontracts/dlcspecs/blob/master/Oracle.md), allowing Cashu mints to leverage existing DLC oracle infrastructure.
+
+Caution: Applications that rely on oracle resolution must verify that the oracle is trustworthy and check via the mint's [info][06] endpoint that NUT-CTF is supported.
+
+**Related specifications:** [NUT-CTF-split-merge][CTF-split-merge] defines split/merge operations for creating and dissolving complete sets of conditional tokens. [NUT-CTF-numeric][CTF-numeric] extends this framework with numeric outcome conditions.
+
+## Terminology
+
+- **Condition**: A question with defined outcomes, resolved by an oracle. Identified by a `condition_id`. Equivalent to "condition" in the [Gnosis Conditional Token Framework](https://docs.gnosis.io/conditionaltokens/).
```
sktr: The Gnosis Conditional Token Framework link appears to be broken.
It currently shows a "Page Not Found" page.

joemphilips: Thanks. It seems like the website is not well maintained. Perhaps I should link to the contract itself (https://github.com/gnosis/conditional-tokens-contracts)

Or referencce the Polymarket's version (https://github.com/Polymarket/ctf-exchange-v2) but this one has slightly different terminology so I must be careful. Let me think

joemphilips: broken link fixed.
It seems to be referencing Polymarket version CTF is not a good idea for now since it has a much broader scope than this (off-chain CLOB matching, UMA oracle, etc.)

MarkovianProtocol:
This is a clean design, and the oracle-agnostic split (mint stays agnostic, outcome tokens redeem against an external attestation) is the right seam.

One attestation source that fits the redemption step without adding a trusted signer: a deterministic-replay resolver. Markovian'"'"'s RESOLVE produces a verdict by replaying public data through a fixed function, so any party (the mint, either counterparty, a third agent) can recompute the same outcome and check the commitment against Bitcoin. The attestation is not "trust this signer," it is "here is the public input, the function, and the anchor, replay it yourself." That keeps the mint agnostic and removes the single-oracle trust assumption in the redemption path.

Honest scope: this only covers outcomes decidable from public data by a deterministic function, which is where DLC-style markets already sit. Anything needing private or subjective input still needs a conventional attestor.

Happy to put up a small reference adapter against whatever attestation interface NUT-CTF settles on, so an outcome token can redeem against a replayable verdict as one option alongside DLC oracles. Would that be useful for the spec at this stage?

joemphilips:
@MarkovianProtocol Thanks for the comment. NUT-CTF is intentionally oracle-agnostic at the redemption layer: any attestation source (DLC oracle, multi-sig oracle set, or a deterministic-replay resolver like yours) that can bind a verdict to a `condition_id` and an outcome keyset can serve as a redemption source, without the spec naming it.

I want to keep the v1 surface small, so the core spec can be reviewed; this PR is already sizable. The right place for a specific resolver like RESOLVE is a companion NUT or supplementary doc once the core is merged and there's a working adapter demonstrating the attestation→keyset mapping. I'd welcome that as a follow-up — please open a separate issue or draft so it can be evaluated on its own.

MarkovianProtocol:
Agreed on keeping v1 oracle-agnostic and the surface small, that is the right call for review. The redemption layer staying attestation-source-neutral is exactly what lets a deterministic-replay resolver plug in without the core spec naming it.

We will do it as you suggested: once the core merges, a separate companion draft with a working adapter that demonstrates the attestation-to-keyset mapping concretely (verdict bound to `condition_id`, replayed from public data, mapped to the outcome keyset), so it can be evaluated on its own rather than expanding this PR. Will open it as its own issue/draft then. Thanks for pointing to the right venue.

MarkovianProtocol:
Opened #411 as the separate draft. Scoped to making the verdict history single and timestamped. Adapter once the core and the cdk reference merge.

perplexiphone-cmd (approved):

perplexiphone-cmd (approved):

CTF.md
perplexiphone-cmd: Merge

PR #339, opened 2026-02-09
Feat(melt): replace Prefer header with prefer_async field in POST data

robwoodgate:
This PR replaces the `Prefer: respond-async` header in NUT-05 with a `"prefer_async": <bool>` field in the POST data.

## Implementations

- CDK: https://github.com/cashubtc/cdk/pull/1618
- Cashu-TS: https://github.com/cashubtc/cashu-ts/pull/500
- Nutshell:
    - https://github.com/cashubtc/nutshell/pull/887 (tested against Cashu-TS PR above)
    - https://github.com/cashubtc/nutshell/pull/801 (being updated)

## Detailed Description

This is a non-breaking change, as mints only honor an async preference if they choose to - there is no mint signalling or guarantee.

The reason for the change is simple - `Prefer` is not on the CORS safelist, so browsers require a preflight check before sending a request with this header. If a mint does not add `Prefer` to their CORS `Access-Control-Allow-Headers` response, the preflight will fail with an unspecified network error.

This creates a "fail unsafe" scenario, because a mint that does not support async must still allow the header to avoid an error. This new approach sidesteps CORS issues, by adding a POST data field the mint can ignore if unsupported.

### DIFF
--- a/05.md
+++ b/05.md
@@ -28,7 +28,7 @@ The melting process follows these steps for all payment methods:
 
 **Synchronous (Default):** The melt request blocks until payment completion. This ensures immediate finality but may result in long request times for slow payment methods.
 
-**Asynchronous (Optional):** When the `Prefer: respond-async` header is included and supported by the mint, the request returns immediately after validation with a `"PENDING"` state. The wallet must then monitor the payment progress through polling or websocket notifications.
+**Asynchronous (Optional):** When the `prefer_async: true` param is included and supported by the mint, the request returns immediately after validation with a `"PENDING"` state. The wallet must then monitor the payment progress through polling or websocket notifications.
 
 ## Common Request and Response Formats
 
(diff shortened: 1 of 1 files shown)


05.md
```diff
@@ -112,28 +111,19 @@ When the `Prefer: respond-async` header is included:
...
-
-```json
-{
-  "quote": <str>,
-  "inputs": <Array[Proof]>
-  // Additional method-specific fields may be required
-}
-```
-
-where `quote` is the melt quote ID and `inputs` are the proofs with a total amount sufficient to cover the requested amount plus any fees.
```
a1denvalu3, line 128: why was this still removed?

robwoodgate: It's probably better to read the consolidated doc (view file), as the original had duplication that makes the diff confusing. This was moved higher up to lines 94-100.

PR #341, opened 2026-02-16
NUT-XX: Get quotes by pubkeys

callebtc:
# Implementations

- [ ] CDK: https://github.com/cashubtc/cdk/pull/1834
- [ ] Nutshell: https://github.com/cashubtc/nutshell/pull/1123
- [ ] Cashu-TS: https://github.com/cashubtc/cashu-ts/pull/782

supersedes #329
Get NUT-20 quotes by `pubkeys`. Requires signatures to prove possession of the corresponding private keys.


xx.md
```diff
@@ -0,0 +1,60 @@
...
+{
+  "pubkeys": <Array[str]>,
+  "pubkey_signatures": <Array[str]>
+}
+```
+
+- `pubkeys` is an array of hex-encoded compressed secp256k1 NUT-20 public keys (33 bytes each)
+- `pubkey_signatures` is an array of hex-encoded Schnorr signatures on `pubkeys` in the same order (64 bytes each)
+
+The wallet **MUST** provide a valid signature in `pubkey_signatures` for each public key in `pubkeys` with the corresponding private key in the same order as the `pubkeys` array. The message to sign is the byte representation of the public key.
```
robwoodgate: I think signature on the plain public key is pretty risky, as it allows signature replay in any "proof of key ownership" context, or even for future uses on other endpoints and other mints.

It would be better defined as a signature on a domain separated message like we do in P2BK / keyset v2

```
eg: SHA-256( b"Cashu_NUTXX_QUOTE_v1" || MINT_URL || method || pubkey )
```

or

```
eg: SHA-256( b"Cashu_NUTXX_QUOTE_v1" || NONCE || pubkey )

where NONCE is a random shared nonce, supplied in a nonce param:

{
  "pubkeys": <Array[str]>,
  "pubkey_signatures": <Array[str]>,
  "nonce": str
}
```
)

xx.md
```diff
@@ -0,0 +1,60 @@
...
+
+Where `MintQuoteResponse` is the quote response type defined in [NUT-04][04].
+
+## Settings
+
+The settings for this NUT are part of the mint info response ([NUT-06][06]):
+
+```json
+{
+  "28": {
```
robwoodgate: 28 now taken... 29?

```suggestion
  "29": {
```

xx.md
```diff
@@ -0,0 +1,60 @@
...
+## Response
+
+The mint responds with a `PostMintQuotesByPubkeyResponse`:
+
+```json
+{
+  "quotes": <Array[MintQuoteResponse]>
+}
+```
+
```
robwoodgate, line 50: We should probably specify what happens if any signature is invalid. Does entire op fail, or does mint return only the valid ones? What happens if more/fewer sigs are provided than pubkeys?

xx.md
```diff
@@ -0,0 +1,60 @@
+# NUT-28: Mint Quote Lookup by Public Key
```
robwoodgate: 28 now taken... 29?

```suggestion
# NUT-29: Mint Quote Lookup by Public Key
```

xx.md
```diff
@@ -0,0 +1,60 @@
...
+```http
+POST https://mint.host:3338/v1/mint/quote/bolt11/pubkey
+```
+
+The wallet includes the following `PostMintQuotesByPubkeyRequest` data:
+
+```json
+{
+  "pubkeys": <Array[str]>,
+  "pubkey_signatures": <Array[str]>
```
robwoodgate, line 24: Would it be useful to group quotes by pubkey?

robwoodgate (changes requested):
The signature scheme is a bit worrying, but otherwise just some nit/questions for clarity.

xx.md
thesimplekid: in the mint response? I think best to not define the ordering and let the wallet figure it out.

robwoodgate: Yeah, I don't mind, a flat array is easy to sort by a wallet. Simpler response. And pubkey is in the quote.

xx.md
thesimplekid: I think we should match the plan for batched minting where if any are invalid the whole op fails.

robwoodgate: Agree. It lowers the abuse surface too, because a mint can reject at first invalid signature (or on mismatch key/sig count), rather than continue processing.

xx.md
robwoodgate:
```suggestion
The wallet **MUST** provide a valid signature in `pubkey_signatures` for each public key in `pubkeys` with the corresponding private key in the same order as the `pubkeys` array. The message to sign is the byte representation of the public key.

### Signature Validation Failure

If there is a mismatch between signature and pubkey counts, or **any signature is invalid**, the mint MUST reject the request and return an error.
```

xx.md
robwoodgate: I added a suggestion to the request section above, but it could be moved to response section if preferred.

xx.md
callebtc: good point!

a1denvalu3 (approved):

d4rp4t (approved):

xx.md
```diff
@@ -0,0 +1,60 @@
+# NUT-29: Mint Quote Lookup by Public Key
```
TheMhv: This needs a new NUT number, NUT-29 is already defined in [NUT-29: Batched Minting](https://github.com/cashubtc/nuts/blob/main/29.md)

xx.md
```diff
@@ -0,0 +1,60 @@
...
+
+Where `MintQuoteResponse` is the quote response type defined in [NUT-04][04].
+
+## Settings
+
+The settings for this NUT are part of the mint info response ([NUT-06][06]):
+
+```json
+{
+  "29": {
```
TheMhv: > This needs a new NUT number, NUT-29 is already defined in [NUT-29: Batched Minting](https://github.com/cashubtc/nuts/blob/main/29.md)

Same thing here

xx.md
thesimplekid: Generally we us XX and just define before merge

xx.md
```diff
@@ -0,0 +1,60 @@
...
+{
+  "pubkeys": <Array[str]>,
+  "pubkey_signatures": <Array[str]>
+}
+```
+
+- `pubkeys` is an array of hex-encoded compressed secp256k1 NUT-20 public keys (33 bytes each)
+- `pubkey_signatures` is an array of hex-encoded Schnorr signatures on `pubkeys` in the same order (64 bytes each)
+
+The wallet **MUST** provide a valid signature in `pubkey_signatures` for each public key in `pubkeys` with the corresponding private key in the same order as the `pubkeys` array. The message to sign is the byte representation of the public key.
```
a1denvalu3: A static schnorr signature allows replay attacks: if I observe a signature i can use it to query the status of  a quote again and again.

xx.md
```diff
@@ -0,0 +1,60 @@
...
+
+`depends on: NUT-04, NUT-20`
+
+---
+
+This NUT adds an endpoint for wallets to get all NUT-20 locked mint quotes associated with a set of public keys. Queries require a valid signature from the owner of the corresponding private keys.
+
+## Request
+
+To query quotes assigned to a public key, the wallet makes a `POST /v1/mint/quote/{method}/pubkey` request.
```
TheMhv, line 13:
```suggestion
To query quotes assigned to a public key, the wallet needs to
1. Get a nonce from `GET /v1/mint/quote/{method}/pubkey` request. 
2. Makes a `POST /v1/mint/quote/{method}/pubkey` request.
```

xx.md
```diff
@@ -0,0 +1,60 @@
...
+  "pubkeys": <Array[str]>,
+  "pubkey_signatures": <Array[str]>
+}
+```
+
+- `pubkeys` is an array of hex-encoded compressed secp256k1 NUT-20 public keys (33 bytes each)
+- `pubkey_signatures` is an array of hex-encoded Schnorr signatures on `pubkeys` in the same order (64 bytes each)
+
+The wallet **MUST** provide a valid signature in `pubkey_signatures` for each public key in `pubkeys` with the corresponding private key in the same order as the `pubkeys` array. The message to sign is the byte representation of the public key.
+
```
TheMhv, line 40:
````suggestion
```json
{
  "pubkeys": <Array[str]>,
  "pubkey_signatures": <Array[str]>,
  "nonce": <hex_str>
}
```

- `pubkeys` is an array of hex-encoded compressed secp256k1 NUT-20 public keys (33 bytes each)
- `pubkey_signatures` is an array of hex-encoded Schnorr signatures on `pubkeys` in the same order (64 bytes each)
- `nonce` is an hex-encoded nonce fetched from previous request.

The wallet **MUST** provide a valid signature in `pubkey_signatures` for each public key in `pubkeys` with the corresponding private key in the same order as the `pubkeys` array. The message to sign is the byte representation of the following hash:

```
SHA-256( DOMAIN_SEPARATOR || nonce || pubkey)
```

Where:
- `DOMAIN_SEPARATOR` constant byte string `b"Cashu_MINTQUOTES_v1"`
- `nonce` is the previous nonce fetched by `GET` request
- `pubkey` is the byte representation of the public key
- `||` denotes concatenation
````

File: xx.md [hunk 11]
```diff
@@ -0,0 +1,60 @@
...
+## Response
+
+The mint responds with a `PostMintQuotesByPubkeyResponse`:
```
TheMhv, line 43:
````suggestion
On GET request, the mint responds with a `GetMintQuotesByPubkeyResponse`:
```json
{
  "nonce": <hex_str>
}
```

On POST request, the mint responds with a `PostMintQuotesByPubkeyResponse`:
````

**Review by TheMhv — CHANGES_REQUESTED** [review]
Following the [suggestion](https://github.com/cashubtc/nuts/pull/341#discussion_r2812465774) from @robwoodgate :

My suggestion is create a GET request for the same path to mint generate a nonce that wallet can be use for sign a hash message, this will prevent replay attacks.

**Comment by a1denvalu3** [comment]
I've opened a PR with the updated signature logic and test vectors here: #363

File: xx.md (hunk 9 above)
Comment (robwoodgate) [reply xx.md:13]: A nonce doesn't need to be mint-supplied. It can just be a random string.

**Comment by robwoodgate** [comment]
> I've opened a PR with the updated signature logic and test vectors here: #363

I support @a1denvalu3 's approach in #363 - using a timestamp also allows a mint to prevent reuse of the signature after xx minutes too.

It would be even better with a domain prefix, in case a similar scheme is used later on elsewhere:

```
SHA-256( b"Cashu_NUTXX_QUOTE_v1" || pubkey || timestamp || mint_pubkey )
```

But with a timestamp max age specified, the reuse window would be minimal in any case.

**Comment by c03rad0r** [comment]
Thanks for opening this — it's great to see this direction being taken seriously. I'm building an ESP32-based captive portal (TollGate) that provides WiFi access in exchange for ecash payments. My use case: a mining proxy on the ESP32 connects to a hashpool translator via SV1 stratum. The translator mints ehash tokens for the miner's locking pubkey and currently pushes them downstream via a custom `mining.token` notification.

This NUT ("Get quotes by pubkeys") would let me eliminate the translator middleman — the ESP32 could directly query the mint for quotes attributed to its pubkey, receive tokens without the push notification hack. This is especially important for resource-constrained devices where running a full translator is impractical.

The domain-separated signature scheme in #363 (timestamp + mint pubkey) looks solid. I'm really looking forward to using this functionality — happy to test with my ESP32 implementation once the spec stabilizes.

**Comment by TheMhv** [comment]
How it's going?

**Comment by robwoodgate** [comment]
To hopefully move this forward, I've proposed some amendments in the PR below, which gives us time-bounded replay protection. It is intended to be merged into this PR.

- https://github.com/cashubtc/nuts/pull/391

File: xx.md (hunk 11 above)
Comment (robwoodgate) [reply xx.md:43]: outdated, marking as resolved

File: xx.md (hunk 10 above)
Comment (robwoodgate) [reply xx.md:40]: outdated, marked resolved

**Review by robwoodgate — APPROVED** [review]
ACK. LGTM

File: xx.md (hunk 1 above)
Comment (robwoodgate) [reply xx.md]: An equivalent has been added at line 39

**Comment by robwoodgate** [comment]
@ye0man - this looks ready for implementations to build on it

**Comment by TheMhv** [comment]
This still vulnerable by replay attacks, why removed `timestamp` field from signed hash?

**Comment by callebtc** [comment]
> This still vulnerable by replay attacks, why removed `timestamp` field from signed hash?

replay can't be fully prevented without a mint state or perfectly synchronized clocks (i.e. time windows to compensate), so we dropped it

**Comment by callebtc** [comment]
ACK

**Comment by TheMhv** [comment]
> replay can't be fully prevented without a mint state or perfectly synchronized clocks (i.e. time windows to compensate), so we dropped it

To prevent replay attack, we can make a request before everything to setup some nonce that can be used for validate the `mint_quote` request.

I suggested it before: 
https://github.com/cashubtc/nuts/pull/341#pullrequestreview-4093726509

File: xx.md [hunk 12]
```diff
@@ -0,0 +1,68 @@
...
+
+```json
+{
+  "quotes": <Array[MintQuoteResponse]>
+}
+```
+
+Where `MintQuoteResponse` is the quote response type defined in [NUT-04][04].
```
robwoodgate, line 51: Presumably, each quote will be the shape specified for the method it represents? So maybe a better wording is something like:


```suggestion
Where `MintQuoteResponse` is the quote response type defined in [NUT-04][04], or an extension of it based on the method the quote relates to. 
```

xx.md
```diff
@@ -0,0 +1,68 @@
...
+
+`depends on: NUT-04, NUT-20`
+
+---
+
+This NUT adds an endpoint for wallets to get all NUT-20 locked mint quotes associated with a set of public keys. Queries require a valid signature from the owner of the corresponding private keys.
+
+## Request
+
+To query quotes assigned to a public key, the wallet makes a `POST /v1/mint/quote/{method}/pubkey` request.
```
robwoodgate, line 13: Do we NEED a `{method}` now quotes have `method` in their structs (https://github.com/cashubtc/nuts/pull/387)?

The endpoint could return quotes for ANY method?


```suggestion
To query quotes assigned to a public key, the wallet makes a `POST /v1/mint/quote/pubkey` request.
```

TheMhv: IMO we don't need to separate those request.
**Concept ACK** for this suggestion

thesimplekid: I think we should keep it. It is more inline with batch minting and the responses are still different shapes so I would prefer to return the uniform response of a list of one response type.

vnprc:
Suggestion: drop `{method}` from the path. It's already been removed in https://github.com/cashubtc/cdk/pull/1834. The endpoint answers "what does this key own," and a NUT-20 key locks quotes of any payment method — so the segment forces one request per method and buys nothing. The signature doesn't cover it, and every `MintQuoteResponse` already carries a `method` field (NUT-04), so a mixed-method response is unambiguous without it.

vnprc:
Suggestion: bound the pubkey count. The mint must verify every signature before it can tell whether the caller is entitled to anything, so request length decides how much work an anonymous caller can demand. Add a MUST to limit it, error 11017 on overflow, and an optional `max_pubkeys` in the NUT-06 settings — the same shape NUT-29 already uses for `max_batch_size`.

vnprc:
Suggestion: let the request bound the response. The query returns a pubkey's full quote history in every state, with no way to request a subset so the response grows unbounded with the number of quotes a key accumulates. Two optional request fields fix it: mintable quotes (`amount_issued < amount_paid`), and an `updated_at` filter. Omitting both preserves current behavior.

robwoodgate:
> Suggestion: drop `{method}` from the path. 

Previously discussed and mint side prefers to keep:
https://github.com/cashubtc/nuts/pull/341#discussion_r3644500473

> Suggestion: bound the pubkey count.

Good idea. I wonder if instead of _yet-another-mintinfo-flag_, we just cap it in the NUT spec to (say) 100?

> Suggestion: let the request bound the response.

In an ideal world, pubkeys would not be constantly reused for privacy, but there is no bound in the real world. I wonder if a mint returning just NON-EXPIRED and mintable quotes (`amount_issued < amount_paid`) is enough to cap and avoid the complexity of filter flags, pagination etc?

vnprc:
> I wonder if instead of yet-another-mintinfo-flag, we just cap it in the NUT spec to (say) 100?

Fine with me. Before we pick a number perhaps we should do some load testing to help inform our choice.

> avoid the complexity of filter flags, pagination etc

I agree in principle that we shouldn't build it before we need it. My use case does a lot more volume than the rest of the cashu ecosystem. This is why Hashpool has been useful in the past for load testing cdk. I plan to rotate locking keys on a timer at the miner wallet, not implemented yet.

I could lean on vardiff or progressive mint fees to limit the number of quotes the mint processes, the latter is probably the best solution (market solutions ftw). Lots of knobs I can turn to shift complexity outside of the cashu protocol definition.

robwoodgate:
> Suggestion: bound the pubkey count. The mint must verify every signature before it can tell whether the caller is entitled to anything, so request length decides how much work an anonymous caller can demand. Add a MUST to limit it, error 11017 on overflow, and an optional `max_pubkeys` in the NUT-06 settings — the same shape NUT-29 already uses for `max_batch_size`.

For cross-reference, this PR would solve the pubkey bound issue in practice (mints already cap array lengths using the value internally): https://github.com/cashubtc/nuts/pull/425

thesimplekid:
> > Suggestion: bound the pubkey count. The mint must verify every signature before it can tell whether the caller is entitled to anything, so request length decides how much work an anonymous caller can demand. Add a MUST to limit it, error 11017 on overflow, and an optional `max_pubkeys` in the NUT-06 settings — the same shape NUT-29 already uses for `max_batch_size`.
> 
> For cross-reference, this PR would solve the pubkey bound issue in practice (mints already cap array lengths using the value internally): #425

There is another issue of not only max pubkey but max quotes per pubkey. If a wallet creates many quotes for a pubkey and then does a look up of just one pubkey but over time has generated many quotes on the mint this could be an expensive look up for the mint especially if we add logic like is mintable etc. We don't have anyway to limit this right now.

callebtc:
> There is another issue of not only max pubkey but max quotes per pubkey. If a wallet creates many quotes for a pubkey and then does a look up of just one pubkey but over time has generated many quotes on the mint this could be an expensive look up for the mint especially if we add logic like is mintable etc. We don't have anyway to limit this right now.

Good point. We would probably have to limit the length of the returned arrays as well.

callebtc:
@thesimplekid @robwoodgate are both implementation PRs gtg?

robwoodgate:
> @thesimplekid @robwoodgate are both implementation PRs gtg?

CTS: Yes, just need to gate the mint info once nut number is finalized and add integration tests once a mint release supports it. So available pretty quickly once nut and cdk merge it

TheMhv:
We can assign a number for this thing?

robwoodgate:
Cross link: 

- https://github.com/cashubtc/nuts/pull/416

thesimplekid:
> > There is another issue of not only max pubkey but max quotes per pubkey. If a wallet creates many quotes for a pubkey and then does a look up of just one pubkey but over time has generated many quotes on the mint this could be an expensive look up for the mint especially if we add logic like is mintable etc. We don't have anyway to limit this right now.
> 
> Good point. We would probably have to limit the length of the returned arrays as well.

We can't just limit it though other wise quotes will be lost if a user only has the pubkey, but i don't see a great solution other then limiting the number of quotes that can be created for a pubkey or adding pagination to the look up by pubkey both not good.

robwoodgate:
> We can't just limit it though other wise quotes will be lost if a user only has the pubkey, but i don't see a great solution other then limiting the number of quotes that can be created for a pubkey or adding pagination to the look up by pubkey both not good. 

Is there an issue if the mint simply rejects a create mint quote request if it uses a pubkey that has been used more than the max_array_inputs number of times?

Then it can always return a max number of quotes, no pagination

robwoodgate (changes requested):
This still needs replay protection to stop an onlooker following the lifecycle of a pubkey forever. Suggest we include timestamp in the request, signed over, and the mint with its NTP server clock can accept or reject it +/- a defined window.

This avoids needing an extra challenge request, and also avoids mint needing to store or track nonces

Request:

```json
{
  "pubkeys": [
    "03d56ce4e446a85bbdaa547b4ec2b073d40ff802831352b8272b7dd7a4de5a7cac",
    "02a4c1d2f3e5b6a7c8d9e0f1a2b3c4d5e6f7a8b9c0d1e2f3a4b5c6d7e8f9a0b1c2"
  ],
  "timestamp": 1788825600,
  "signatures": [
    "d9be080b33179387e504bb6991ea41ae0dd715e28b01ce9f63d57198a095bb2f...",
    "3f2a8c1e5b7d9f0a2c4e6b8d1f3a5c7e9b0d2f4a6c8e1b3d5f7a9c0e2b4d6f8a..."
  ]
}
```

Signature scheme:

```
For each pubkey P_i in the request, the wallet computes:

    msg_i = len32(P_i) || P_i || len32(t) || t

where:
    P_i    = the 33-byte compressed public key
    t      = the timestamp as an 8-byte big-endian unsigned integer
             (Unix seconds)
    len32  = the byte length of the following field as a 4-byte
             big-endian unsigned integer
    ||     = byte concatenation

The wallet signs:

    sig_i = BIP340_sign(k_i, TaggedHash("Cashu_QuoteLookup_v1", msg_i))

where k_i is the private key for P_i and TaggedHash is the BIP-340
tagged hash: SHA256(SHA256(tag) || SHA256(tag) || msg) with the tag
as UTF-8 bytes.

signatures[i] MUST be the signature for pubkeys[i] (same order,
same length). All signatures in one request commit to the same
timestamp.

The mint MUST reject the request if:
- |now - timestamp| > TIMESTAMP_WINDOW (suggested: 300 seconds),
  checked before any signature verification; or
- the pubkeys and signatures arrays differ in length; or
- any signature fails BIP-340 verification against the
  x-coordinate of its pubkey.

Rejection is all-or-nothing; a stale timestamp returns a distinct
error code so wallets can retry with a fresh timestamp rather than
treating it as an authorization failure.
```

Issue #301, opened 2025-10-21
Improved recovery


a1denvalu3:
## Improved Recovery (Efficiency and Privacy)

The way recovery is currently performed is a privacy disaster: the wallet leaks the entire spending history in an attempt to recover unspent user funds. This issue describes the current recovery method, followed by a revised approach which reduces the number of leaked **unissued** notes by a logarithmic factor and limits the number of leaked **issued** notes to an arbitrary $d$.

## Current Recovery Modal

Let $T$ be the index of the nonce that was issued in the last `BlindMessage` (meaning there is no other issued note whose nonce index $O > T$).

```
[0] |--------------------------------------------------------------------------------- [T] >                                                                 
```

On the first iteration our current approach sets a batch size $b$, then computes the nonces and relative `BlindMessage`s in $[0, b-1]$ and queries the Mint find out which one of them are issued and spent.

```
[0] |xxxxxxxxxx [b-1] --------------------------------------------------------------- [T] > 
                                                                                 
```

On the second iteration, we check the next batch of `BlindMessages`s in $[b, 2b-1]$.

```
[0] |xxxxxxxxxxxxxxxxxxxx [2b-1] --------------------------------------------------- [T] > 
                                                                                 
```

This process continues until the $i$-th iteration, when we find that the last batch was returned empty or partially empty.

```
[0] |xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx [T] > ***************  [i*b-1]
```
After which it follows that if $u$ is the number of un-issued notes in the last batch, then $T = i\cdot b-u-1$, and we conclude.
Obviously, this procedure is terrible for privacy because we're linking together our entire wallet activity history and -on top of that- we're revealing nonces for which `BlindMessage`s were unissued (the segment part marked with `*`).

Other than being non-privacy friendly, this approach is also relatively in-efficient: it takes $\mathcal{O}(\frac{n}{b})$ network requests, with $n$ being the total number of issued notes.

## Improved Approach

We subdivide our improved approach into two "phases" to make it more digestable:
1. we enstablish a property for spent and issued notes.
2. we introduce a binary search method to find $T$ with logarithmically fewer queries.

### 1. Spent-Issued Invariant

[...]

PR #342, opened 2026-02-18
NUT-XX: Efficient wallet recovery

Forte11Cuba:
Motivated by #301 . Replaces the NUT-13 linear scan with binary search + a depth invariant that bounds unspent proofs to the last `d` nonce indices.

Reduces BlindedMessages revealed to the mint from O(T) to ~182 (constant), and requests from O(T/b) to ~35. No mint changes required.

Builds on the binary search and d-invariant ideas from #301 . Credit to @a1denvalu3  for the original proposal and algorithm design, @robwoodgate  for the halfway-house approach and the concerns around coin-selection that shaped the final invariant, @gandlafbtc  for the local-search alternative that framed the trade-offs, and @davidcaseria  for raising the practicality constraints that led to the relaxed invariant.

Open for discussion this is a draft NUT, not final. Feedback welcome especially on: the Depth Invariant definition, the gap-tolerance approach, and the default value of `d`.

### DIFF
--- a/342.md
+++ b/342.md
@@ -1,4 +1,4 @@
-# NUT-XX: Efficient Wallet Recovery
+# NUT-342: Efficient Wallet Recovery
 
 `optional`
 
@@ -202,15 +202,15 @@ This NUT is fully backwards-compatible:
 
 ### Nostr mint backups (NUT-27)
 
-To signal to a restoring wallet that the original wallet maintained the Depth Invariant, wallets that implement this NUT and perform [NUT-27][27] backups **SHOULD** include a `nut_xx: true` field in the unencrypted JSON payload.
+To signal to a restoring wallet that the original wallet maintained the Depth Invariant, wallets that implement this NUT and perform [NUT-27][27] backups **SHOULD** include a `nut_342: true` field in the unencrypted JSON payload.
 
-Example of a [NUT-27][27] backup payload with the `nut_xx` hint:
+Example of a [NUT-27][27] backup payload with the `nut_342` hint:
 
 ```json
 {
   "mints": ["https://mint.example.com", "https://another-mint.org"],
   "timestamp": 1703721600,
-  "nut_xx": true
+  "nut_342": true
 }
 ```
 
--- a/342.md
+++ b/342.md
@@ -6,7 +6,7 @@
 
 ---
 
-This NUT defines an improved recovery algorithm for [NUT-13][13] wallets that reduces the number of `BlindedMessages` revealed to the mint from O(T) to a constant (~182), improving both privacy and efficiency. It combines a binary search to locate the last issued nonce index T with a **Depth Invariant** that confines all unspent tokens to the last `d` nonce indices.
+This NUT defines an improved recovery algorithm for [NUT-13][13] wallets that reduces the number of `BlindedMessages` revealed to the mint from O(T) to a constant (~182 = log₂(2³²) + g + d with default parameters), improving both privacy and efficiency. It combines a binary search to locate the last issued nonce index T with a **Depth Invariant** that confines all unspent tokens to the last `d` nonce indices.
 
 ## Depth Invariant
 
--- a/342.md
+++ b/342.md
@@ -1,4 +1,4 @@
-# NUT-342: Efficient Wallet Recovery
+# NUT-342: Efficient Wallet Recovery with Dynamic Gap Backup
 
 `optional`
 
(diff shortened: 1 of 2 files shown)


a1denvalu3:
> This NUT reduces leakage to O(log N + g + d) BlindedMessages

It also speeds up recovery, because you send log N requests.

XX.md
```diff
@@ -0,0 +1,248 @@
...
+`depends on: NUT-03, NUT-07, NUT-09, NUT-13`
+
+---
+
+This NUT defines an improved recovery algorithm for [NUT-13][13] wallets. It is motivated by the privacy and efficiency limitations of the current linear scan approach described in [nuts#301](https://github.com/cashubtc/nuts/issues/301).
+
+The recovery procedure defined in [NUT-13][13] reveals the wallet's entire transaction history to the mint. During recovery, the wallet sends every `BlindedMessage` it has ever generated — including nonces for tokens that were never issued — to the mint in sequential batches. This has two consequences:
+
+1. **Privacy**: The mint can correlate all of the user's past ecash activity retroactively, defeating the unlinkability guarantees of ecash.
+2. **Efficiency**: Recovery requires O(T/b) network requests, where T is the total number of issued notes and b is the batch size.
```
a1denvalu3: I would cut this and instead put in the PR description. This is more of a motivation.

Forte11Cuba: Good point trimmed it down to a two-line summary. The detailed  motivation is already in the PR description.

thesimplekid: Not specific to this pr but we should also make sure we include context like this not just in the PR description but in the git log so we don't lose it the day github enviably doesn't work.

342.md
```diff
@@ -0,0 +1,241 @@
...
+### Maintaining the invariant
+
+To maintain the invariant, wallets that implement this NUT **MUST**:
+
+1. **On receive (mint/swap-in)**: After receiving new tokens, T increases. If any existing unspent `Proof` now has a nonce index `i ≤ T - d`, the wallet **MUST** consolidate those proofs by re-issuing them via a [NUT-03][03] swap. The re-issued proofs receive new nonce indices near T, restoring the invariant.
+
+2. **On send (swap-to-send)**: A send operation involves a [NUT-03][03] swap that creates new outputs (send proofs + change proofs), which increases T. If this increase causes any remaining unspent proofs (those not involved in the swap) to have index `i ≤ T - d`, the wallet **MUST** consolidate them. Note: a pure melt (paying a Lightning invoice without swap) does not increase T and therefore cannot violate the invariant.
+
+3. **Coin selection**: Wallets **MAY** use arbitrary coin selection. After any operation that increases T, the wallet **MUST** check whether any unspent proof now has index `i ≤ T - d` and consolidate if so. This preserves free coin selection while enforcing the invariant.
+
```
a1denvalu3, line 55: There is also the case where a wallet already has `d` unspent proofs, and adding any more would violate the constraint no matter if you reissue the old ones. In that case, the operation should fail and compaction should be done to compress all of the existing unspent proofs into fewer ones.

XX.md
```diff
@@ -0,0 +1,241 @@
...
+2. **On send (swap-to-send)**: A send operation involves a [NUT-03][03] swap that creates new outputs (send proofs + change proofs), which increases T. If this increase causes any remaining unspent proofs (those not involved in the swap) to have index `i ≤ T - d`, the wallet **MUST** consolidate them. Note: a pure melt (paying a Lightning invoice without swap) does not increase T and therefore cannot violate the invariant.
+
+3. **Coin selection**: Wallets **MAY** use arbitrary coin selection. After any operation that increases T, the wallet **MUST** check whether any unspent proof now has index `i ≤ T - d` and consolidate if so. This preserves free coin selection while enforcing the invariant.
+
+> [!NOTE]
+> The consolidation swap is a standard [NUT-03][03] swap — it does not reveal any additional information about the wallet's history beyond what is already implicit in the swap operation itself.
+
+### Consolidation strategy
+
+Since recovery can be triggered at any time by data loss, the invariant **MUST** hold at all times — not just "before recovery". Wallets **MUST** consolidate violating proofs immediately after any operation that increases T.
```
a1denvalu3:
```suggestion
Since recovery can be triggered at any time by data loss, the invariant **MUST** hold at all times — not just "before recovery". Wallets **MUST** consolidate violating proofs immediately before any operation that increases T.
```

XX.md
```diff
@@ -0,0 +1,241 @@
...
+
+### Breakeven points
+
+This NUT reveals **more** nonces than [NUT-13][13] for small wallets:
+
+- **Privacy breakeven**: T ≈ 132. For wallets with T < 132, [NUT-13][13] reveals fewer nonces.
+- **Efficiency breakeven**: T ≈ 825 (with batched gap scan). For wallets with T < 825, [NUT-13][13] makes fewer requests.
+
+For wallets below these thresholds, the overhead of binary search (32 sequential round-trips) and the fixed recovery window (`d` nonces) exceeds the cost of a simple linear scan. Wallets **MAY** use a heuristic: if a previous recovery or local state suggests T < 200, fall back to the [NUT-13][13] linear scan.
+
```
a1denvalu3: really good, although one doesn't know T until found.

XX.md
```diff
@@ -0,0 +1,241 @@
...
+### Privacy properties
+
+The privacy gain for wallets above the breakeven comes from two properties:
+
+1. **Bounded leakage**: The number of revealed nonces is constant (182) regardless of wallet history size. A wallet with T = 1,000 reveals the same number of nonces as one with T = 100,000.
+2. **Non-sequential probes**: The 32 binary search probes are spread across the entire nonce space and are non-contiguous, making transaction history correlation significantly harder for the mint compared to the sequential batches of [NUT-13][13].
```
a1denvalu3: I think this information is not necessary to implementors and can be documented elsewhere.

XX.md
```diff
@@ -0,0 +1,231 @@
+# NUT-XX: Efficient Wallet Recovery
+
+`optional`
+
+`depends on: NUT-03, NUT-07, NUT-09, NUT-13`
+
+---
+
+This NUT defines an improved recovery algorithm for [NUT-13][13] wallets that reduces the number of `BlindedMessages` revealed to the mint from O(T) to a constant (~182), improving both privacy and efficiency. It combines a binary search to locate the last issued nonce index T with a **Depth Invariant** that confines all unspent tokens to the last `d` nonce indices.
```
a1denvalu3: > [...] reduces the number of `BlindedMessages` revealed to the mint from O(T) to a constant (~182)

Why costant 182?

thesimplekid:
On the cdk call we discussed that a wallet birthday would be useful so a wallet knows if it need to use the old linear scan or can use the new method. We could use [NUT27](https://github.com/cashubtc/nuts/blob/main/27.md) to store a birthday in addition to the mint list. Currently there is a timestamp in the body of the backup but I think that's the time stamp of the backup, adding a optional wallet birthday to that json structure would be a simple backwards compatible change and helpful here.

a1denvalu3:
@thesimplekid Thinking more about it, the birthday doesn't really tells us if we can use the fast recovery. Maybe storing an explicit hint that the wallet respects the properties defined in `nut-xx` would be a more helpful tip.

thesimplekid:
Good point. I think that would be reasonable to add to the event as well. Since the event body is just json we could specify that it can be extended by other nuts with optional json fields this way nut-27 does not define it or need to change for possible future nuts and each nut can define that.

a1denvalu3:
@Forte11Cuba made a PR to your PR, addressing what we talked about.

callebtc:
I think the idea is great. However, this introduces significant complexity in wallet proof management that hasn't been considered enough. Building a wallet that makes sure that all proofs are always within a band of the derivation index is not trivial. 

Additionally, this PR should rather be seen as a suggestion to wallet developers rather than a spec change because it does not improve interoperability as long as we do not require all wallets to use this exact same proof management system. It's highly unlikely that we will be able to convince all wallet developers to use the exact same proof management system and agree on meta parameters to make restore for all wallets predictable. 

I suggest to close this PR and recommended the method for implementations rather than a protocol requirement.

a1denvalu3:
@Forte11Cuba There is an alternative (simpler) way we can avoid maintaining the d-invariant, and I strongly suggest you to consider it.
First, we could allow redefine `d` to be `the gap between the last unspent note and the first`, and let it be arbitrarily big. At the same time, we "back up" this number by storing it together with the `BlindedMessage` object that we send to the Mint as an output  (for a swap, mint or melt request). Example: 
```json
{
  "amount": int,
  "id": hex_str,
  "B_": hex_str,
  "d_gap": int, # <-- gap between this particular note's derivation index and the derivation index of the first unspent note in the wallet
}
```

Then, during recovery, we do only the first pass of binary-search to find the last unspent note. The last unspent note will tell us what `d_gap` is, we then proceed to do a linear scan recovery (normal recovery) from `last_unspent-d` to `last`.
Furthermore, `d_gap` could even be stored encrypted with a symmetric encryption scheme like AES-128 using the `r` blinding factor with which `BlindedMessage` was generated as key.

Lastly, it is extremely easy to keep track of `d_gap` in the wallet (or calculated it dynamically), if you store the secret derivation index within the `Proof` object in the wallet:

```json
{
  "amount": int,
  "id": hex_str,
  "secret": str,
  "C": hex_str,
  "derivation_index": int
}
```

Then `d_gap` is just `last_unspent.derivation_index - first_unspent.derivation_index`

a1denvalu3:
@Forte11Cuba  I've applied the changes in my comments, here. The recovery process is now fully backwards compatible without the need to extend the nostr backup, and the d-invariant is gone: https://github.com/Forte11Cuba/nuts/pull/2

a1denvalu3:
@Forte11Cuba Please review the PR

342.md
```diff
@@ -1,248 +1,202 @@
...
-### Maintaining the invariant
+To prevent the mint from knowing the wallet's activity patterns or the exact number of active coins, the `d_gap` value **SHOULD** be encrypted by the wallet before sending.
 
-To maintain the invariant, wallets that implement this NUT **MUST**:
+#### Encryption Procedure
 
-1. **On receive (mint/swap-in)**: After receiving new tokens, T increases. If any existing unspent `Proof` now has a nonce index `i ≤ T - d`, the wallet **MUST** consolidate those proofs by re-issuing them via a [NUT-03][03] swap. The re-issued proofs receive new nonce indices near T, restoring the invariant.
+1. **Derive key**: Let `r` be the 32-byte private blinding factor used to construct the corresponding `BlindedMessage`'s `B_` (`B_ = Y + rG`). Derive a 128-bit key using SHA-256:
+   ```
+   K_aes = SHA256(r)[0:16]
```
a1denvalu3: Maybe this could have additional domain separation

a1denvalu3:
image

robwoodgate:
https://github.com/cashubtc/cashu-ts/pull/788

Got a working demo in Cashu-TS, using Nutshell @ https://github.com/cashubtc/nutshell/pull/1081

Screenshot 2026-07-14 at 22 28 00

robwoodgate:
Been experimenting with the search phase, and the results are worth bringing back into the spec discussion. Everything below is wallet-local: no changes to the mint requirements, the d_gap mechanism, or the wire format. Code, benchmarks and a runnable comparison script are in cashubtc/cashu-ts#794.

## The observation: Binary search maps the space, not the wallet.
Counters live in a 2^32 space, but real wallets cluster at the very bottom of it. For scale: a daily-use wallet I sampled has nine keysets whose last counters are 2024, 728, 69, 53, 19, 4, 4, 2 and 2.

Binary search spends the same ~31 sequential questions whether T is 20 or 20 million, because its cost is set by the size of the space.

And since each probe depends on the previous answer, none of it can be batched, even though NUT-09 happily accepts many blinded messages per request.

## A batched alternative.

The variant I ended up with ("exponential ladder-grid-tile") exploits both facts:

**Invariant**

> A probe window containing any signature proves T is at or beyond its start; an empty window proves T is below its start, provided derivation gaps stay shorter than the window.

**Worked trace** (w = 25, wallet's unknown T = 5000; `##` = signatures returned, `--` = empty):

```
Round 1 - ladder (one request): windows at 0 and w*2^k, doubling across the space
 start:  0   25   50  100  200  400  800 1600 3200 6400 12800 ... 2^31
         ##  ##   ##  ##   ##   ##   ##   ##   ##   --    --       --
                                       last hit^    ^first empty
 => T is in (3224, 6400)

Round 2 - grid (one request): 12 windows spread evenly across the bracket
 start:  3469 3713 3957 4202 4446 4690 4934 5179 5423 5667 5911 6155
         ##   ##   ##   ##   ##   ##   ##   --   --   --   --   --
 => T is in (4958, 5179)

Round 3 - tile (one request): the remaining 220 counters, wall to wall
         4959 ............... 5178
         ############---------
 => T = 5000, and its BlindSignature already carries the encrypted d_gap

Round 4 - fetch [T - d_gap, T] in chunks, filter spent per NUT-07
```

Full recovery lands at ~5 requests at any wallet age (binary search: ~31, always). The tables in cashubtc/cashu-ts#794 include an exact reproduction of the benchmark posted earlier in this thread, so the numbers are directly comparable.

**On privacy:** replaying the nine-keyset wallet above through both searches, a tuned ladder came within 20 linked signatures of binary search while using 230 fewer requests; per-keyset breakdown in the PR.

## Suggested spec changes:

1. Keep `find_t` non-normative and note explicitly that NUT-09 places no constraint on probe placement: probes may be batched, non-contiguous windows. Present the batched search as the recommended strategy, with the cost table split by wallet age rather than a single constant.
2. Upgrade `d_gap` encryption from SHOULD to MUST. Plaintext gaps leak activity patterns at issuance time, on every swap, not just at recovery.
3. Note that the searchable space depends on the derivation scheme: keysets restored via legacy hardened BIP-32 paths cap counters at 2^31, not 2^32.
4. Have mints advertise their maximum restore/checkstate batch length in the nut info block. Recovery has to chunk the window (a `d_gap` larger than the request cap is routine for unconsolidated wallets), and today the cap is discovered by error.
5. Name the privacy lever explicitly: recovery links `d_gap + 1` signatures, so wallets **SHOULD** keep the gap small by consolidating opportunistically, folding old proofs into swaps they are already making. Correctness never depends on it.

a1denvalu3:
@robwoodgate I wonder if the NUT should just describe the protocol addition of `d_gap`. The rest is just an algorithmic strategy and it could vary from wallet to wallet, given they that they are all still compatible.

robwoodgate:
> @robwoodgate I wonder if the NUT should just describe the protocol addition of `d_gap`. The rest is just an algorithmic strategy and it could vary from wallet to wallet, given they that they are all still compatible.

Yes, that is true, and was what I was leaning towards in my comment - _Keep find_t non-normative._ 

The `d_gap` insight is what really makes the magic work. And the `d_gap` is the big privacy lever for wallets (smaller gap, less mandatory privacy leakage). The search strategy for `T` is, as you say, irrelevant to inter-wallet compatibility.

robwoodgate:
> @robwoodgate I wonder if the NUT should just describe the protocol addition of `d_gap`. The rest is just an algorithmic strategy and it could vary from wallet to wallet, given they that they are all still compatible.

It probably does no harm to give an **ILLUSTRATIVE** T-search strategy - either binary search, as now, or here's a pseudocode for the exponential ladder-grid-tile:

```
Probe window (w, default 25). The number of consecutive counters checked at each probe position.
A window containing any signature proves T lies at or beyond its start; an empty window proves T
lies below it. NOTE: w must exceed the wallet's longest run of never-used counters.

Probe budget (budget, default 12). The maximum number of probe windows sent in one search
request. It sets the narrowing rate: each grid round splits the unknown zone into budget + 1
segments, so the search needs one request per ~13x of zone width at the default.

MAX = top of counter space (2^31 for legacy v0 keysets, else 2^32)

 function find_t(keyset_id) {
      // one request: a w-wide window at 0 and at each doubling position
      hits = restore([window(0)] + [window(w * 2^k) for k = 0 .. log2(MAX/w)])
      if (hits is empty) return undefined                  // caller falls back to linear scan
      floor   = highest counter with a signature           // T >= floor
      ceiling = lowest rung start above floor, else MAX+1  // T <  ceiling
      while (floor + 1 < ceiling) {
          zone = (floor + 1, ceiling)                      // width = ceiling - floor - 1
          if (width <= budget * w) {                       // default: width < 300 (12 * 25)
              hits = restore(tile(zone))                   // wall to wall
              floor = highest counter with a signature
              break                                        // zone fully probed: floor is T
          }
          hits = restore(spread(budget, zone))             // 12 probes, ~13x narrowing
          update floor and ceiling from hits
      }
      return floor    // = T; its BlindSignature carries d_gap
  }

```

But if any is included, it SHOULD NOT be mandatory. Just illustrative.

robwoodgate:
Some more experimentation today - using demo in https://github.com/cashubtc/cashu-ts/pull/794, but would apply to any T-Search strategy.

## d_gap is the overriding factor

The T-search is cheap at any wallet age (~4 requests batched, ~31 bisected); the recovery window costs `d_gap`, and the wallet controls it.

Three live runs of the bundled demo against the dockerized Nutshell mint (`input_fee_ppk` 100; each regime is an env knob on the demo) differ only in wallet policy:

| Wallet history | Method | Restore requests | Messages | Linked | Wall |
| --- | --- | --- | --- | --- | --- |
| unmanaged, small spends (T=4,409, d_gap = T) | T-search | 13 | 6,155 | 4,409 | 32s |
| | NUT-09 scan | 12 | 6,000 | 4,409 | 33s |
| pending token sent mid-history, refresh on (T=16,176, d_gap=8,489) | T-search | 22 | 10,377 | 8,739 | 60s |
| | NUT-09 scan | 36 | 18,000 | 16,176 | 115s |
| oldest-refresh policy (T=16,285, d_gap=135) | T-search | 5 | 1,510 | 586 | 1.8s |
| | NUT-09 scan | 36 | 18,000 | 16,285 | 114s |

 (The 13 and 22 each include one extra request from a retried dropped connection.)

## Observations

**Lump Sum funding with small spend maximizes`d_gap`.** The window is set by the wallet's single oldest live proof, and nothing about normal usage moves it. Small spends never break a lump-funded wallet's large reserve proofs: in the first row the counter-0 reserve survived every round, the window is the whole history, and the search sits at its floor: scan cost plus three search requests.

**Unclaimed sends pin the window** at their own counter even while consolidation runs (second row: the refresh policy was active throughout), and dust that never clears its own fee pins it the same way.

**Randomized coin selection keeps old proofs in play indefinitely**. Only heavy turnover consolidates by accident (a run swapping ~half the balance per round landed at d_gap 172 with no policy at all). 

**A cheap periodic refresh bounds `d_gap`** by the refresh cadence (the demo rolls its oldest proofs forward every 20 swaps): one extra swap per cycle, a few counters (third row). 

## Conclusion

Wallets should actively manage the window by periodically refreshing their oldest live proofs (folding dust in), rather than relying on spending patterns; long-unclaimed sends set a floor no wallet policy can cross.

a1denvalu3:
@robwoodgate Ran some simulations on a CDK implementation I'm working on:

```bash
target_index=10000 trial=1 nonce=4fe9efa69ae75c70 swaps=1017 min=22 median=302 max=898 final=273 shrinks=203
target_index=10000 trial=2 nonce=d937071fc64d38cc swaps=1018 min=21 median=335 max=985 final=82 shrinks=194
target_index=10000 trial=3 nonce=1dc99293a81fe450 swaps=1027 min=20 median=356 max=927 final=461 shrinks=192
target_index=50000 trial=1 nonce=b3d10aa67fa156fa swaps=5088 min=16 median=333 max=1029 final=283 shrinks=1031
target_index=50000 trial=2 nonce=6d39fc20c0a1389a swaps=5104 min=20 median=438 max=1359 final=474 shrinks=1011
target_index=50000 trial=3 nonce=149e7b26f3e86db9 swaps=5125 min=22 median=338 max=1446 final=182 shrinks=1057
target_index=100000 trial=1 nonce=46cdee83fe734a55 swaps=10196 min=10 median=347 max=1270 final=268 shrinks=2056
target_index=100000 trial=2 nonce=bb4b074a82b80f83 swaps=10200 min=24 median=319 max=1294 final=1270 shrinks=2014
target_index=100000 trial=3 nonce=f291618cd8000aa5 swaps=10199 min=11 median=335 max=1132 final=405 shrinks=2011
target_index=250000 trial=1 nonce=28ada00ff96669ba swaps=25576 min=13 median=335 max=1251 final=494 shrinks=5082
target_index=250000 trial=2 nonce=26dcdcd66482c64e swaps=25622 min=11 median=336 max=1247 final=377 shrinks=5141
target_index=250000 trial=3 nonce=0482ef89edcdf0fc swaps=25574 min=10 median=358 max=1570 final=249 shrinks=5120

```

It looks like using a "oldest first" biased coinselection, where the algorithm takes in the `derivation_index` of the proofs and -at parity of value- it prefers proofs that are "older" is enough to make `d_gap` a median ~350 across pretty much any length.

a1denvalu3:
Also, in i've changed the shape of `BlindedMessage` and `BlindedSignature` to this:

```json
{
  "amount": int,
  "id": hex_str,
  "B_": hex_str,
  "metadata": {
    "342": hex_str
  }
}
```

robwoodgate:
Given the desire to avoid storing extra data in the mint, I've been playing with a different approach to recovery that doesn't rely on storing `d_gap`, and the speedup is pretty impressive so far: cashubtc/cashu-ts#964

Recovery in 2.5 seconds vs 89 seconds for a wallet with 13k counters used.

It's a refactoring of the original algo, so it walks the same range and touches the same counters as the NUT-09 approach. The difference is, it queries on `Y`, which has lower linkability than `B_` (only the original wallet can reproduce its `B_`, whereas `Y` is known to anyone who's ever held the token), and only fetches a `B_` for proofs that turn out to still be live. On that run the mint recognised 22 blinded messages instead of 13,013.

The walk being the same, the win is that there's far less intensive cryptographic work for both wallet and mint. It also means the heavy per-signature work tracks the live proof count rather than the history, so it stays cheap however fragmented the wallet is.

Obviously the windowed approach in this NUT draft is superior for a wallet with good `d_gap` hygiene, but this seems to handle the `[0, T]` cases like lump sum, small spend much better overall. And doesn't require any new NUT specification, because it uses the existing NUT-07/09 APIs.

@a1denvalu3 would value your eyes on it.

PR #343, opened 2026-02-18
Define canonical JSON encoding for unsigned 64-bit integers

robwoodgate:
The NUTs define amounts as unsigned 64-bit integers, but the JSON wire encoding is ambiguous.

JSON has no integer type, and many clients parse JSON numbers via float-like defaults. This can cause cross-language divergence and precision loss for large values (e.g, in environments where JSON numbers use IEEE-754).

This PR defines a canonical `UInt64Json` format, which is backwards compatible and precision safe across runtimes.

Simply:

> `UInt64Json` **MAY** be represented as a JSON integer number or a base-10 decimal string.
> 
> Implementations **MUST** accept both forms.
>
> Values greater than `9007199254740991 (2^53-1)` **MUST** be represented as decimal strings.

This applies to `int` / `u64` fields across the protocol, including amounts in Proofs, Blinded Messages, and request/response payloads.

Fee fields are unchanged to keep the change minimal, but we could include all int/u64 JSON fields for consistency.


prusnak:
When exactly are the amounts bigger than `(2^53-1)`?

```
>>> 21e6 * 1e8 < 2**53
True
```

00.md
```diff
@@ -56,11 +56,24 @@ These specifications use elliptic curve public key cryptography over the field a
...
 - Byte sequences
+- `int` / `u64`: unsigned 64-bit integer in the range: `0..2^64-1`.
+
+#### UInt64Json
+
+For JSON fields, `int` / `u64` values are encoded as `UInt64Json`.
+
+`UInt64Json` **MAY** be represented as a JSON integer number or a base-10 decimal string.
+
+Implementations **MUST** accept both forms.
```
robwoodgate, line 70: If we want to encourage a single standard (decimal string) over time, we could tweak:

```suggestion
Implementations **MUST** accept both forms, and **SHOULD** encode values as decimal strings.
```

We already consistently use string amounts in the `/v1/keys` api, so this would be a continuation of that precedent.

callebtc:
I think this is too excessive. NUT-00 could state that all amounts are uint64 and we could leave the rest as is.

SatsAndSports:
> NUT-00 could state that all amounts are uint64

When you say 'uint64', do you mean _non-quoted_ integers?

As mentioned by others, key amounts are quoted currently , but fees are non-quoted. Both of those, I think, make sense as fees can't realistically get very big


```
$ curl https://mint.minibits.cash/Bitcoin/v1/keys | jq .                                                                                             
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current                                                                                              
                                 Dload  Upload   Total   Spent    Left  Speed                                                                                                
100  5341  100  5341    0     0   1045      0  0:00:05  0:00:05 --:--:--  1100                                                                                               
{                                                                                                                                                                            
  "keysets": [                                                                                                                                                               
    {                                                                                                                                                                        
      "id": "00107937db0cc865",                                                                                                                                              
      "unit": "sat",                                                                                                                                                         
      "active": true,                                                                                                                                                        
      "input_fee_ppk": 0,                                                                                                                                                    
      "keys": {                                                                                                                                                              
        "1": "03be63a0f422f8db6297fb3018bc3e626751010434c19c9c8c990e1c4e438f03dc",  
...
        "9223372036854775808": "02d900df693b1505cc5cbdef13a7ee793b89ebfedb628f4dc5fbbe49a7ef8cd07a"
```

robwoodgate:
> I think this is too excessive. NUT-00 could state that all amounts are uint64 and we could leave the rest as is.

I have trimmed it down to essentially the NUT-00 change. LMK if you want other changes.

robwoodgate:
@ye0man - A lot of things have changed since this PR was raised. Cashu-TS, for example now implements a `JSONInt` parser to handle decimal integers over the safe limit, and an `Amount` VO to normalise amount handling regardless of initial type (string, integer, bigint). 

What started as a hard blocker in the JSON safe integer limit (2 ^ 53 - 1) is now mostly resolved in the ecosystem. 

We should still discuss whether wire forms should accept dual string / decimal integer forms, but it isn't as imperative as it was back in February.

PR #147, opened 2024-07-05
NUT-00: Remove Base64 padding from tokens

zeugmaster:
Add small mention of `base64_urlsafe` padding characters and test vector

PR #353, opened 2026-03-16
Specify NUT-22 authA serialization

robwoodgate:
NUT-22 does not specify the base64 encoding variant to use.

CDK uses [general_purpose::URL_SAFE](https://github.com/cashubtc/cdk/blob/8ed5c86625ef6e238692761c8a64231ebd560a57/crates/cashu/src/nuts/auth/nut22.rs#L244), which requires base64 URL _**with padding**_ .

This PR requires mints to support both padded and unpadded forms (as per https://github.com/cashubtc/nuts/pull/147).

# Implementation
- [x] Cashu-TS: https://github.com/cashubtc/cashu-ts/pull/529 (sends padded)
- [x] CDK: https://github.com/cashubtc/cdk/issues/1740
- [x] Nutshell: https://github.com/cashubtc/nutshell/issues/919

**NOTE:** This was not an issue prior to keysets v2, as keyset v1 IDs (16 chars) produced serialized BAT JSON of exactly 174 bytes, and 174 % 3 = 0 — so base64 encoding never produced any padding before.

### DIFF
--- a/22.md
+++ b/22.md
@@ -170,13 +170,15 @@ The wallet checks the `MintBlindAuthSetting` of the mint to determine which endp
 
 ### Serialization
 
-To add a blind authentication token (BAT) to the request header, we need to serialize a single `AuthProof` JSON as URL-safe base64 encoding **with padding** (RFC 4648 §5) with the prefix `authA`:
+To add a blind authentication token (BAT) to the request header, we need to serialize a single `AuthProof` JSON as base64url with the prefix `authA`:
 
 ```sh
-authA[base64urlpadded_authproof_json]
+authA[base64url_authproof_json]
 ```
 
 This string is a BAT.
+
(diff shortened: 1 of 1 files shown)


thesimplekid:
@robwoodgate looks like what I told you was wrong. We did not restrict the spec for the normal tokens. So I think we should do the same here and cdk should be changed to be less restrictive. 

https://github.com/cashubtc/nuts/pull/147

robwoodgate:
> @robwoodgate looks like what I told you was wrong. We did not restrict the spec for the normal tokens. So I think we should do the same here and cdk should be changed to be less restrictive.
> 
> #147

@thesimplekid  - that is even better. I've tweaked the spec to say mints MUST support both forms. 

I'll leave Cashu-TS on padded, as that is backwards compatible with CDK <= 0.15.2

Please can you raise a ticket to sort CDK and pop it on this PR when ready.

thesimplekid (approved):

PR #355, opened 2026-03-26
feat(nut-18, nut-26): add preferred_mints to payment requests

a1denvalu3:
## Summary
- Introduces `preferred_mints` field (`pm` in JSON, `0x09` in TLV) to payment requests in NUT-18 and NUT-26.
- This allows the originator of the request to specify mints they support.
- Paying wallets can optionally use these preferred mints for faster and cheaper transactions without bridging through the lightning network.

### DIFF
--- a/18.md
+++ b/18.md
@@ -37,8 +37,8 @@ Here, the fields are
 - `a`: The amount of the requested payment
 - `u`: The unit of the requested payment (MUST be set if `a` is set)
 - `s`: Whether the payment request is for single use
-- `m`: A set of mints from which the payment is requested
-- `pm`: A set of preferred mints the receiver supports. The sender can choose to pay from one of these mints for faster and cheaper transactions
+- `m`: A set of mints from which the payment is requested. If this field is defined, the receiver SHOULD only accept proofs from these mints.
+- `pm`: A set of preferred mints the receiver supports. If this field is defined, the receiver will accept proofs from any mint but prefers these mints for faster and cheaper transactions.
 - `d`: A human readable description that the sending wallet will display after scanning the request
 - `t`: The method of `Transport` chosen to transmit the payment (can be multiple, sorted by preference)
 - `nut10`: The required [NUT-10][10] locking condition
--- a/18.md
+++ b/18.md
@@ -24,6 +24,7 @@ A Payment Request is defined as follows
   "u": str <optional>,
   "s": bool <optional>,
   "m": Array[str] <optional>,
+  "pm": Array[str] <optional>,
   "d": str <optional>,
   "t": Array[Transport] <optional>,
   "nut10": NUT10Option <optional>,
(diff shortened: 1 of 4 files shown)


18.md
```diff
@@ -37,6 +38,7 @@ Here, the fields are
 - `u`: The unit of the requested payment (MUST be set if `a` is set)
 - `s`: Whether the payment request is for single use
 - `m`: A set of mints from which the payment is requested
```
thesimplekid: Should we be more clear on when to use mints vs preferred mints. If mints is defined then only tokens from those mints will be accepted and receive wallets SHOULD not redeem proofs of a mint not in the list. While if preferred mints is defined any mints tokens will be accepted and they will be melted.

robwoodgate: I was just writing the same comment. it's not clear to me when to use `m` vs `pm`

18.md
```diff
@@ -36,7 +37,8 @@ Here, the fields are
 - `a`: The amount of the requested payment
 - `u`: The unit of the requested payment (MUST be set if `a` is set)
 - `s`: Whether the payment request is for single use
-- `m`: A set of mints from which the payment is requested
+- `m`: A set of mints from which the payment is requested. If this field is defined, the receiver SHOULD only accept proofs from these mints.
```
robwoodgate: Maybe better to put the onus on the sender? And is it only a SHOULD, or should it be a MUST?
What happens if the sender sends proofs from a different mint?

```suggestion
- `m`: A set of mints from which the payment is requested. If this field is defined, the sender SHOULD only send proofs from these mints.
```

realmeylisdev:
Hi — this is useful, thanks. Following my comment on #357 earlier today, I'm engaging from a wallet-implementer angle (Flutter bindings in `cdk-flutter`).

One small gap in the current text: **is the `pm` array ordered?**

NUT-18's `t` (Transport) field is explicitly spec'd as "can be multiple, **sorted by preference**". The proposed `pm` says "preferred mints the receiver supports" but doesn't state whether the array order conveys ranking. Since "preferred" implies a relative preference, senders will inevitably start treating the array as a priority list — and without a canonical answer in the spec, implementations will diverge (alphabetical, HashSet iteration order, insertion order, etc.).

Suggested small wording, modeled on `t`:

> `pm`: A set of preferred mints the receiver supports, **sorted by preference**. If this field is defined, the receiver will accept proofs from any mint but prefers these mints for faster and cheaper transactions. …

This gives the sender a deterministic selection heuristic when the wallet has balance at multiple mints in the list.

Orthogonal to the `m` vs `pm` clarification @thesimplekid and @robwoodgate are discussing above — happy to submit this as a small editorial follow-up PR if it's a useful direction.

robwoodgate:
I have cherry-picked the commits from this PR into https://github.com/cashubtc/nuts/pull/381, which also adds an "extra fee" field for non preferred mint payments.

thesimplekid:
Closing for https://github.com/cashubtc/nuts/pull/381

PR #358, opened 2026-04-01
Tighten nut-11 spending conditions

robwoodgate:
This PR removes ambiguity by tightening the spending conditions to specifically reject malformed P2PK secrets.

- Nutshell: https://github.com/cashubtc/nutshell/pull/969
- CDK: https://github.com/cashubtc/cdk/pull/1880
- Cashu-TS: https://github.com/cashubtc/cashu-ts/pull/578

### DIFF
--- a/11.md
+++ b/11.md
@@ -82,6 +82,8 @@ Supported tags are:
 - `refund: <hex_str>` are additional public keys that can provide signatures after `locktime` (_allows multiple entries_)
 - `n_sigs_refund: <int>` specifies the minimum number of [Refund Multisig](#refund-multisig) public keys providing valid signatures after `locktime` expires
 
+Each of the above tags may appear exactly **ONCE** in a secret. If a tag appears more than once, the P2PK secret is malformed and the Proof **MUST** be rejected as invalid.
+
 > [!NOTE]
 >
 > The tag serialization type is `[<str>, <str>, ...]` but some tag values are `int`. Wallets and mints must cast types appropriately for de/serialization.
@@ -96,6 +98,8 @@ Signature flags are defined in the tag `Secret.tags['sigflag']`. Currently, ther
 If any one input has the signature flag `SIG_ALL`, then all inputs are required to have the same kind, the flag `SIG_ALL` and the same `Secret.data` and `Secret.tags`, otherwise an error is returned.
 
 `SIG_INPUTS` is only enforced if no input is `SIG_ALL`.
+
+If a proof has any other signature flag value, the P2PK secret is malformed and the Proof **MUST** be rejected as invalid.
 
 ### Signature flag `SIG_INPUTS`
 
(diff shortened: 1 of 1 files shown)


11.md
```diff
@@ -82,6 +82,8 @@ Supported tags are:
 - `refund: <hex_str>` are additional public keys that can provide signatures after `locktime` (_allows multiple entries_)
 - `n_sigs_refund: <int>` specifies the minimum number of [Refund Multisig](#refund-multisig) public keys providing valid signatures after `locktime` expires
```
KvngMikey, line 83: This PR tightens tag-count rules well, but n_sigs and n_sigs_refund values themselves are still underspecified. Three edge cases remain unaddressed:

`n_sigs = 0`:  is this a valid anyone can spend shorthand or a malformed proof?

`n_sigs > total available keys`:  unsatisfiable condition, should the proof be immediately rejected at mint time or only at spend time?

A non-integer string value like `"two"`: the existing spec already says wallets/mints must cast types, but there's no rejection rule here comparable to the new duplicate-tag rule.

i suggest adding a line such as: "If `n_sigs` or `n_sigs_refund` is not a positive integer, or exceeds the total number of keys in its pathway, the P2PK secret is malformed and the Proof MUST be rejected as invalid."

11.md
```diff
@@ -253,6 +257,41 @@ The signature flag `sigflag` indicates that signatures are necessary on the `inp
 ]
 ```
 
+## Public Key Canonicalisation
+
+Public keys **MUST** use the [compressed Secp256k1 public key format](https://learnmeabitcoin.com/technical/public-key#public-key-format).
+
+A public key may appear in both the [Locktime Multisig](#locktime-multisig) and [Refund Multisig](#refund-multisig) pathways at the same time, but only **ONCE** in each pathway.
```
KvngMikey: the sentence "A public key may appear in both the Locktime Multisig and Refund Multisig pathways at the same time, but only ONCE in each pathway" could be interpreted as blanket permission for duplicates as long as they span two pathways. 

I suggest we reword to something like this:
"Within a single pathway (Locktime or Refund), each key MUST be unique by its x-coordinate. A key that is unique within each pathway MAY appear in both pathways simultaneously."

this makes it constraint first, then the permission, which is the safer reading order for implementers.

11.md
robwoodgate: > `n_sigs = 0`: is this a valid anyone can spend shorthand or a malformed proof?

Good point. I think malformed.

> `n_sigs > total available keys`: unsatisfiable condition, should the proof be immediately rejected at mint time or only at spend time?

Secret is blinded at mint time, so only rejectable at spend time. This condition is "unspendable" (malformed).

> A non-integer string value like `"two"`: the existing spec already says wallets/mints must cast types, but there's no rejection rule here comparable to the new duplicate-tag rule.

Malformed - the spec clearly defines these as "integer"

> i suggest adding a line such as: "If `n_sigs` or `n_sigs_refund` is not a positive integer, or exceeds the total number of keys in its pathway, the P2PK secret is malformed and the Proof MUST be rejected as invalid."

I like that. Will add.

11.md
robwoodgate: Thanks @KvngMikey - I cleaned it up and went constraint first

KvngMikey (approved):
lgtm

11.md
```diff
@@ -82,6 +82,10 @@ Supported tags are:
 - `refund: <hex_str>` are additional public keys that can provide signatures after `locktime` (_allows multiple entries_)
 - `n_sigs_refund: <int>` specifies the minimum number of [Refund Multisig](#refund-multisig) public keys providing valid signatures after `locktime` expires
 
+Each of the above tags may appear exactly **ONCE** in a P2PK secret. If a tag appears more than once, the P2PK secret is malformed and the Proof **MUST** be rejected as invalid.
```
thesimplekid: I think rejected is not clear that it means the proof is unspendable.

robwoodgate: so if I change to 'be rejected as unspendable`?

Or just 'and the Proof **MUST** be treated as unspendable`?

robwoodgate: changed invalid to unspendable.

robwoodgate:
@thesimplekid @callebtc - we have merged implementations across the core repos.
Can we consider merging, unless any other wording tweaks required?

11.md
```diff
@@ -82,6 +82,10 @@ Supported tags are:
 - `refund: <hex_str>` are additional public keys that can provide signatures after `locktime` (_allows multiple entries_)
 - `n_sigs_refund: <int>` specifies the minimum number of [Refund Multisig](#refund-multisig) public keys providing valid signatures after `locktime` expires
 
+Each of the above tags may appear exactly **ONCE** in a P2PK secret. If a tag appears more than once, the P2PK secret is malformed and the Proof **MUST** be rejected as unspendable.
```
a1denvalu3, line 85: As per discussion, we could enforce split handling on the mint and wallet sides.

1. The mint could treat malformed NUT-11 as anyone-can-spend
2. The wallets could reject malformed NUT-11.

robwoodgate: What change in wording would work for you?

thesimplekid (approved):

PR #359, opened 2026-04-02
nut10: Add witness section and warning

robwoodgate:
Mints currently treat Proofs with a witness field but a plain (non-NUT-10) secret differently.

Nutshell rejects, CDK ignores the witness.

This PR codifies this behavior with a warning.

### DIFF
--- a/10.md
+++ b/10.md
@@ -73,7 +73,7 @@ Signatures or other witness data are provided in the `Proof.witness` as a serial
 
 > [!IMPORTANT]
 > `Proof.witness` is only expected for proofs containing a NUT-10 Well-known Secret.
-> A mint MAY reject a Proof containing a witness field if its secret is not NUT-10 compliant
+> A mint MAY reject a Proof containing a witness field if its secret cannot be parsed as a [Well-known Secret](#well-known-secret)).
 
 ## Examples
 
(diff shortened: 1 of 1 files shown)


robwoodgate:
@ye0man - This PR still highlights a real problem, but this far down the road, I think the real solution is to tweak Nutshell to ignore witnesses it doesn't need (follow CDK).

With keysets v3 on the horizon, there will be plenty of non-NUT10 shaped secrets that DO require a witness so the comment, as written, will age badly.

KvngMikey:
With the proposal for Nutshell to adapt CDK's behavior (ignore witnesses it doesn't need), I think the normative text in this PR should flip from "MAY reject" to "SHOULD ignore":

> `Proof.witness` is only expected for proofs containing a NUT-10 Well-known Secret.
> A mint SHOULD ignore a `witness` field on a Proof whose secret cannot be parsed as a Well-known Secret.

Worth keeping/adding alongside this:
- Mints SHOULD still reject malformed JSON secrets that declare a NUT-10 shape or known kind (genuine malformedness, not plainness).
- Mints MUST still reject plain secrets inside SIG_ALL batches, where every input must be a spending condition.
- Optionally: mints MAY log when they drop a witness on a plain secret, so the behavior is observable without being a hard failure.

PR #362, opened 2026-04-09
fix nut-29 test vector, clarify concatenation rules.

robwoodgate:
The NUT-29 test vector was not correctly using NUT-20 message-to-sign rules.

This PR fixes the test vector, clarifies the rules, and adds the intermediate steps for verification (eg message hash)

Note for reviewers:

I confirmed that cashu-ts and cdk both were doing NUT-29 message aggregation correctly. It was only the test vector which was incorrect.


thesimplekid:
> I confirmed that cashu-ts and cdk both were doing NUT-29 message aggregation correctly. It was only the test vector which was incorrect.

Do they already use these test vectors then or we need to add them probably?

robwoodgate:
> > I confirmed that cashu-ts and cdk both were doing NUT-29 message aggregation correctly. It was only the test vector which was incorrect.
> 
> Do they already use these test vectors then or we need to add them probably?

Cashu-ts uses the specific NUT-29 vector (how I found out it was wrong), cdk uses other tests, but not that vector specifically

robwoodgate:
Superseded by #375

PR #363, opened 2026-04-17
Update NUT-29 signature message to include mint pubkey and timestamp

a1denvalu3:
## Summary
- Update msg_to_sign for the pubkey quote lookup to include a timestamp and the mint's pubkey
- Add test vectors for NUT-29 request signature verification

### DIFF
--- a/xx.md
+++ b/xx.md
@@ -21,14 +21,16 @@ The wallet includes the following `PostMintQuotesByPubkeyRequest` data:
 ```json
 {
   "pubkeys": <Array[str]>,
-  "pubkey_signatures": <Array[str]>
+  "pubkey_signatures": <Array[str]>,
+  "timestamp": <int>
 }
 ```
 
 - `pubkeys` is an array of hex-encoded compressed secp256k1 NUT-20 public keys (33 bytes each)
-- `pubkey_signatures` is an array of hex-encoded Schnorr signatures on `pubkeys` in the same order (64 bytes each)
+- `pubkey_signatures` is an array of hex-encoded Schnorr signatures in the same order (64 bytes each)
+- `timestamp` is the current Unix timestamp in seconds
 
-The wallet **MUST** provide a valid signature in `pubkey_signatures` for each public key in `pubkeys` with the corresponding private key in the same order as the `pubkeys` array. The message to sign is the byte representation of the public key.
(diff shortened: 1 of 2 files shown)


a1denvalu3:
This PR includes updates to the signature scheme and test vectors, intended to be merged into the branch for #341.

TheMhv (approved):
That's better approach to fix reply attack on https://github.com/cashubtc/nuts/pull/341

Concept ACK 2fd727aef4550625aa194ec18cd7c4ad5a997172

xx.md
```diff
@@ -21,14 +21,16 @@ The wallet includes the following `PostMintQuotesByPubkeyRequest` data:
 ```json
 {
   "pubkeys": <Array[str]>,
-  "pubkey_signatures": <Array[str]>
+  "pubkey_signatures": <Array[str]>,
+  "timestamp": <int>
```
robwoodgate, line 25: We should add a hard rule about acceptable timestamp age - eg, mint will reject any timestamp older than xx minutes, and any timestamp greater than current epoch

xx.md
```diff
@@ -21,14 +21,18 @@ The wallet includes the following `PostMintQuotesByPubkeyRequest` data:
...
-- `pubkey_signatures` is an array of hex-encoded Schnorr signatures on `pubkeys` in the same order (64 bytes each)
+- `pubkey_signatures` is an array of hex-encoded Schnorr signatures in the same order (64 bytes each)
+- `timestamp` is the current Unix timestamp in seconds
 
-The wallet **MUST** provide a valid signature in `pubkey_signatures` for each public key in `pubkeys` with the corresponding private key in the same order as the `pubkeys` array. The message to sign is the byte representation of the public key.
+The wallet **MUST** provide a valid signature in `pubkey_signatures` for each public key in `pubkeys` with the corresponding private key in the same order as the `pubkeys` array. The message to sign is the SHA-256 hash of the concatenated string `pubkey || timestamp || mint_pubkey`. Where `||` denotes concatenation, `pubkey` is the hex-encoded public key, `timestamp` is the UTF-8 string representation of the Unix timestamp, and `mint_pubkey` is the hex-encoded public key of the mint as defined in [NUT-06][06].
+
+The mint **MUST** evaluate the `timestamp` field and reject the request if it is more than 60 seconds in the past or future.
```
thesimplekid: I think 60 seconds may be too strict. We've had reports of mints having the time in the mint info off by 15 minutes.

robwoodgate: With NTP there should be little reason for clocks to be that far out really. I think maybe up to 300 seconds is more reasonable - still a 10 minute window overall.

a1denvalu3:
NUT-29  signature message is already being changed in https://github.com/cashubtc/nuts/pull/375

PR #368, opened 2026-05-04
feat(nut12): deterministic DLEQ nonce derivation

robwoodgate:
- [x] Cashu-TS: https://github.com/cashubtc/cashu-ts/pull/696
- [x] CDK: https://github.com/cashubtc/cdk/pull/2122
- [x] Nutshell: https://github.com/cashubtc/nutshell/pull/1010

## Problem

In a DLEQ proof, the mint publishes *challenge* and *response* values `e` and `s`, related by:

$$s = r + e \cdot a \pmod{n}$$

* `a` - the mint’s long-lived private key (from its keyset).
* `r` - a temporary, per-proof nonce used in the DLEQ signature.

If the same `r` is ever reused with two different challenges `e₁` and `e₂`, the private key can be recovered immediately:

$$a = (s_1 - s_2) \cdot (e_1 - e_2)^{-1} \pmod{n}$$

It’s the same class of failure that affected Bitcoin Android wallets in 2013, where repeated ECDSA nonces (`k`) allowed attackers to derive private keys and drain funds.

> [!IMPORTANT]
> Current Cashu implementations are secure in practice, because they use modern CSPRNGs, such as `crypto.getRandomValues()` in browsers, `crypto.randomBytes()` in Node.

However, the current spec does not "fail safe". A poorly written RNG wrapper, re-used seed, or language-level entropy issue could silently break key safety while remaining spec-compliant.

## Solution - The spec should "Fail Safe"

To remove the RNG "weak link", RFC 6979 and BIP-340 both moved from "random nonce" to deterministic nonce derivation: the nonce is derived from the private key and the signing context rather than sampled fresh each time.

This PR brings the same approach to NUT-12's DLEQ proof.

The nonce `r` is now derived via HMAC-SHA256, keyed on the mint's private key and bound to the proof context:

```
r = HMAC-SHA256(key=a, data="Cashu_DLEQ_R_v1" || A || B' || C' || ctr) mod n
```

HMAC-SHA256 must be used because the r must be SECRET (keyed to mint secret key `a`), and the inputs to the message to sign are all public knowledge.

Mints SHOULD use this derivation. Mints using random nonces MUST use a CSPRNG.

This is not a breaking change, as verifiers reconstruct R1 and R2 from (e, s) and never see r.

All existing proofs continue to verify.

Also adds a test vector section to `tests/12-tests.md` with fixed inputs and exact expected (e, s) output, so implementations can verify byte-for-byte conformance.

## Proof

Here is a simplified TS proof showing how nonce reuse allows private key recoverability:

```typescript
import { invert, mod } from "@noble/curves/abstract/modular.js";
import { expect, test } from "vitest";

const n = BigInt(
  "0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFEBAAEDCE6AF48A03BBFD25E8CD0364141",
);

const a = BigInt(
  "0x1234567890abcdef1234567890abcdef1234567890abcdef1234567890abcdef",
);

const reusedNonce = BigInt(
  "0x4242424242424242424242424242424242424242424242424242424242424242",
);

const proof1 = {
  e: BigInt("0x1111111111111111111111111111111111111111111111111111111111111111"),
  s: BigInt("0xc49a2d11da2194415dfd54aa2a948adf95f5b1168422bb6386df3170690f28ca"),
};

const proof2 = {
  e: BigInt("0x3333333333333333333333333333333333333333333333333333333333333333"),
  s: BigInt("0xc94a02b109e0383f95737979fb391c1b82adb1d8589b0d6a5046b13fe672b499"),
};

test("reused DLEQ nonce reveals the mint private key", () => {
  // Prove s values were calculated using same private key
  expect(proof1.s).toBe(mod(reusedNonce + proof1.e * a, n));
  expect(proof2.s).toBe(mod(reusedNonce + proof2.e * a, n));

  const sDiff = mod(proof1.s - proof2.s, n);
  const eDiff = mod(proof1.e - proof2.e, n);
  const recoveredA = mod(sDiff * invert(eDiff, n), n);
  // Prove private key was recovered
  expect(recoveredA).toBe(a);
});

test("the same proof used twice reveals nothing (0 = a · 0)", () => {
    // Nonce reuse alone is not sufficient, the two challenges must differ.
    // With one proof, both diffs are zero: recovery solves 0 = a · 0,
    // which every key in the field satisfies, so no information leaks.
    const sDiff = mod(proof1.s - proof1.s, n); // 0n
    const eDiff = mod(proof1.e - proof1.e, n); // 0n

    expect(sDiff).toBe(0n);
    expect(eDiff).toBe(0n);

    // invert(0, n) has no solution; recovery throws rather than
    // returning a key (BigInt has no Infinity / NaN sentinel).
    expect(() => invert(eDiff, n)).toThrow();
  });
```


a1denvalu3:
Why are we hashing the key `a` before inputting into the HMAC?

robwoodgate:
> Why are we hashing the key `a` before inputting into the HMAC?

Mostly for defense in depth - it keeps the raw scalar out of direct HMAC-key use, and ensures the HMAC key is always fixed-width 32-byte, regardless of keygen.

So just key-derivation hygiene (it doesn't add entropy or protect weak mint keys etc)

a1denvalu3:
@robwoodgate I think it's an unnecessary measure and also `a` is always 32 bytes

robwoodgate:
> @robwoodgate I think it's an unnecessary measure and also `a` is always 32 bytes

Agree it's a bit belt and braces - am happy we just use the scalar without hashing if you think it adds nothing.

robwoodgate:
> @robwoodgate I think it's an unnecessary measure and also `a` is always 32 bytes

Addressed in [e57b141](https://github.com/cashubtc/nuts/pull/368/commits/e57b1412414f5686df554f4ba0cba96cd244a312)

TheMhv (approved):
LGTM.

ACK e57b1412414f5686df554f4ba0cba96cd244a312

a1denvalu3:
@robwoodgate 

The HMAC output `x` is uniformly distributed in the range `[0, 2^256 - 1]`. The secp256k1 curve order `SECP256K1_N` is approximately `2^256 - 2^128`. When `x >= SECP256K1_N`, the code subtracts `SECP256K1_N` from `x`. This maps values in the upper range `[SECP256K1_N, 2^256 - 1]` down to the lower range `[0, 2^256 - 1 - SECP256K1_N]`.

Because of this re-mapping, the values in the lower range `[0, 2^256 - 1 - SECP256K1_N]` have **two preimages** (generated when the HMAC output is either `r` or `r + SECP256K1_N`). In contrast, values in the upper range `[2^256 - SECP256K1_N, SECP256K1_N - 1]` only have **one preimage**. This means that nonces in the lower range are twice as likely to be generated as nonces in the higher range.

Over a large number of DLEQ proofs, an attacker could collect these signatures and apply lattice reduction techniques (Hidden Number Problem) to extract the mint's private key `a` due to the non-uniform distribution of nonces.

robwoodgate:
> @robwoodgate
> 
> The HMAC output `x` is uniformly distributed in the range `[0, 2^256 - 1]`. The secp256k1 curve order `SECP256K1_N` is approximately `2^256 - 2^128`. When `x >= SECP256K1_N`, the code subtracts `SECP256K1_N` from `x`. This maps values in the upper range `[SECP256K1_N, 2^256 - 1]` down to the lower range `[0, 2^256 - 1 - SECP256K1_N]`.
> 
> Because of this re-mapping, the values in the lower range `[0, 2^256 - 1 - SECP256K1_N]` have **two preimages** (generated when the HMAC output is either `r` or `r + SECP256K1_N`). In contrast, values in the upper range `[2^256 - SECP256K1_N, SECP256K1_N - 1]` only have **one preimage**. This means that nonces in the lower range are twice as likely to be generated as nonces in the higher range.
> 
> Over a large number of DLEQ proofs, an attacker could collect these signatures and apply lattice reduction techniques (Hidden Number Problem) to extract the mint's private key `a` due to the non-uniform distribution of nonces.

Thank you for raising this. It sounds scarier than it is - the bias between secp `n` and a 256 bit hash is 2^-128, which is considered cryptographically negligible, and a tradeoff we have made elsewhere in the spec (eg NUT-13 blinding factors).

It's a 1-bit bias (bad) in 1 out of 2^128 of the generated `r`s , which makes even SEEING a biased `r` highly improbable.

That said, the theoretical risk is totally valid, and DLEQ matters much more than NUT-13 as the mint key is long-lived and the risk of fund loss would be catastrophic.

So I agree it's a belt and braces change we should make, so have switched to rejection sampling instead of mod n for the nonce.

12.md
```diff
@@ -18,11 +18,11 @@ The complete DLEQ proof reads
 (These steps occur when Bob returns C')
 
 Bob:
-r = random nonce
+r = HMAC-SHA256(key=a, data="Cashu_DLEQ_R_v1" || A || B' || C' || 0x00) # Deterministic nonce
```
prusnak: nitpick: maybe let's rename `0x00` to `ctr` to avoid implementations just use 0x00 by not reading the spec thoroughly?

and then use this below?

``` patch
-0x00 = nonce counter byte; increment and retry if r == 0 or r >= n (max 256 tries)
+ctr = nonce counter byte; set to 0x00 initially, increment and retry if r == 0 or r >= n (max 256 tries)
```

robwoodgate: I was trying to save words, but yes, that is clearer

robwoodgate: Done

robwoodgate:
Current reference implementations now use rejection sampling too.

robwoodgate:
One final clarification added (https://github.com/cashubtc/nuts/pull/368/commits/4affebec14a026c85d2f7a8dc892b96c2d583274)  to the legend. This does not change the derivation, just clarifies the public points are uncompressed raw bytes.

@callebtc - we now have three compliant implementations, please can you do a final review and consider merging.

robwoodgate:
@a1denvalu3 - please can you also do a final review if you get a chance

a1denvalu3 (approved):

thesimplekid (approved):

PR #371, opened 2026-05-20
NUT-00: add BLS12-381 (v3) protocol

robwoodgate:
## Implementations

- [x] Cashu-TS: https://github.com/cashubtc/cashu-ts/pull/949
       (core v3 already merged in https://github.com/cashubtc/cashu-ts/pull/661)
- [ ] CDK: https://github.com/cashubtc/cdk/pull/2194
- [ ] Nutshell: https://github.com/cashubtc/nutshell/pull/999

**NOTE:** Length framed keyset ID derivation added in 68840ccd6bd71fa2c0a104353d741c78505fd7d1

## Summary

Adds **Pairing-based BDHKE (BLS12-381)** as the v3 Cashu blind-signature protocol for keysets with version byte `02`, alongside the legacy secp256k1 protocol for `00`/`01` keysets.

- **NUT-00** defines the BLS12-381 protocol: `hash_to_curve_G1` with the Cashu RFC 9380 DST, multiplicative blinding `B_ = r·Y`, blind signing `C_ = a·B_`, unblinding `C = r^-1·C_`, and verification by `e(C, G2) == e(Y, K)`.
- **NUT-00** also defines deterministic weighted batch verification using a Fiat-Shamir transcript, rejection-sampled weights in `Fr*`, and explicit point-validation requirements to reject non-canonical encodings, identity points, off-curve points, and non-prime-order subgroup points before signing or verification.
- **NUT-01** version-gates key serialization: `00`/`01` keysets use compressed secp256k1 keys, while `02` keysets use compressed BLS12-381 G2 keys.
- **NUT-02** adds V3 keyset ID derivation with version byte `02`, G2 public keys in the preimage, lowercased units, and updated V3 test vectors.
- **NUT-12** scopes DLEQ to secp256k1 keysets only. V3 proofs and blind signatures do not carry `dleq`; offline verification uses the pairing equality from NUT-00.
- **NUT-13** extends deterministic recovery for V3 keysets: secrets remain HMAC-SHA256-derived as in V2, while blinding factors use rejection sampling against `BLS_FR_ORDER` instead of modular reduction.
- **tests/00, 02, 13** add vectors for BLS round-trip signing, weighted batch verification, V3 keyset IDs, and V3 deterministic blinding-factor rejection sampling.

The JSON/wire shape of `BlindedMessage`, `BlindSignature`, `Proof`, and TokenV4 remains unchanged; the keyset version selects the curve and byte widths for `B_`, `C_`, `C`, and mint public keys.


Egge21M:
ChatGPT Pro Review

### 1. Blocking: externally supplied BLS points need explicit subgroup validation

The new BLS section says compressed encodings are fixed-width and that infinity is never valid, but it does **not** explicitly require subgroup checks for `B_`, `C_`, `C_`, and `K`. That is too loose for BLS12-381, because the serialized curve format validates curve points but does not, by itself, guarantee prime-order subgroup membership; the BLS draft defines `subgroup_check(P)` separately and requires public keys to be valid, non-identity points in the correct subgroup, and `CoreVerify` performs a signature subgroup check before pairing verification.  ([[IETF Datatracker](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-bls-signature)][1])

I’d add normative text along these lines in NUT-00:

> Implementations **MUST** deserialize all v3 `B_`, `C_`, `C`, and `K` values using the canonical compressed BLS12-381 encoding, reject non-canonical encodings, reject infinity, reject points not on the relevant curve, and reject points not in the prime-order subgroup. Mints **MUST** perform this validation on `B_` before computing `C_ = a·B_`; wallets and receivers **MUST** perform it on mint public keys and signatures before pairing checks.

This matters especially for `B_`, because it is attacker-controlled input to the mint’s secret scalar multiplication.

### 2. Blocking/high: batch “rejection sampling” is actually modulo reduction

The batch section says weights are derived “by rejection sampling,” but the pseudocode does:

```text
h   = SHA256(challenge || u32_BE(i) || u8(ctr))
r_i = OS2IP(h) mod BLS_FR_ORDER
```

and only rejects `r_i == 0`. That is not rejection sampling; it is modulo reduction. With BLS12-381’s 255-bit scalar order and a 256-bit SHA-256 output, the bias is not as negligible as with secp256k1: `2^256 = 2*q + rem`, so some residues have three preimages and the rest have two. The total variation distance is about 7.5% by this derivation. 

I’d change this to one of:

```text
x = OS2IP(h)
if x == 0 or x >= BLS_FR_ORDER: continue
r_i = x
```

or use a wider expand, for example 64 bytes via XMD/HKDF/expand_message, then reduce modulo `BLS_FR_ORDER` and describe it as wide reduction, not rejection sampling. If you choose true 32-byte rejection sampling, `u8(ctr)` is also unnecessarily bounded; use `u32_BE(ctr)` or define behavior after 255 retries.

The same issue is worth revisiting in NUT-13: the v3 deterministic blinding factor is currently `OS2IP(HMAC-SHA256(...)) mod BLS_FR_ORDER`. That may be acceptable if the project explicitly accepts biased deterministic blinders, but it should be a deliberate choice, not an accidental carry-over from secp256k1 where the order is much closer to `2^256`. 

### 3. High: v3 keyset-ID test vectors reuse the same G2 key for multiple amounts

The new NUT-02 v3 test vectors use the exact same G2 public key for amount `1` and `2`, and again for `1`, `2`, `4`, and `8`. 

That is risky as a canonical example. In this protocol, verification of a proof checks `e(C, G2) == e(Y, K)` for the amount’s selected mint key `K`; the amount itself is not cryptographically inside the signature.  If two denominations share the same private/public key, a proof signed for one amount can be relabeled as another amount using the same `K`.

I’d either update the vectors to use distinct valid G2 public keys per amount, or add a strong rule in NUT-01/NUT-02:

> Within a keyset, public keys for distinct amounts **MUST** be distinct. Mints **MUST NOT** reuse the same signing scalar/public key for multiple amounts.

The safer option is to do both.

### 4. Medium: the new test vectors do not actually expose `Y` or `K`

The NUT-00 v3 vector says it covers `hash_to_curve_G1`, blinding, signing, and unblinding, but it only lists `secret`, `r`, `a`, `B_`, `C_`, and `C`; it does not list `Y = hash_to_curve_G1(secret)` or `K = a·G2`. 

That makes it hard for independent implementations to isolate failures in hash-to-curve, serialization, signing, or unblinding. I decoded the provided G1 points and verified the scalar relations `C_ = 2·B_` and `C_ = 3·C`; the implied `Y` from `B_ = 3·Y` is:

```text
Y = 860d58e5aeda1376185436ed96412313424cc079e056d1dab595e6db4c2c9685fec7da052c8db68d88985b75a42388ad
```

That still should be independently confirmed as the RFC 9380 hash-to-curve output for `test_message` with the Cashu DST. RFC 9380 defines `hash_to_curve` as hashing to field, mapping, adding, and clearing the cofactor, and requires domain separation tags for these encodings. ([[RFC Editor](https://www.rfc-editor.org/rfc/rfc9380.html)][2])

I’d add:

```text
Y:      ...
K:      ...   # compressed G2 for a = 2
```

and at least one batch-vector containing the transcript, challenge, per-proof weights, and expected pass/fail result.

### 5. Medium: version gating is inconsistent and may accidentally bind future versions

NUT-00 says the BLS scheme applies to keysets with version byte `02` “onwards,” while NUT-01 says version byte `02` uses compressed BLS12-381 G2 keys.

I’d avoid “onwards” unless the intent is to permanently reserve all future keyset versions for this exact BLS12-381 protocol and encoding. Safer wording:

> Applies to keysets with version byte `02`. Future version bytes MUST specify their own cryptographic protocol or explicitly inherit this one.

### 6. Low: the serialization reference is pinned to an older draft

NUT-00 links to `draft-irtf-cfrg-pairing-friendly-curves-11`. The Datatracker currently shows `draft-irtf-cfrg-pairing-friendly-curves-12` as the latest revision, with revision date 2025-11-02.  ([[IETF Datatracker](https://datatracker.ietf.org/doc/draft-irtf-cfrg-pairing-friendly-curves/)][3])

Since the PR relies on the Zcash-style BLS12-381 compressed encoding, I’d either link the unversioned Datatracker document or explicitly cite the serialization format being used. The curves draft’s appendix describes compressed G1 as 48 bytes, compressed G2 as 96 bytes, and the three metadata bits, which matches the PR’s intended format. ([[IETF Datatracker](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-pairing-friendly-curves-11)][4])

### 7. Low: the NUT-02 sample code is ambiguous around `serialize()`

The V3 derivation text says to concatenate lowercase compressed G2 public key hex strings, but the Python sample uses:

```python
f"{k}:{v.serialize()}".encode("utf-8")
```

If `serialize()` returns bytes, Python will hash the `b'...'` representation rather than the lowercase hex string. 

I’d make the sample explicit:

```python
pubkey_hex = v.serialize().hex()  # or v.to_hex(), depending on the type
```

and also apply `.lower()` to the unit in the sample, since the prose says the unit is lowercased.


[1]: https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-bls-signature "
            
                draft-irtf-cfrg-bls-signature-06
            
        "
[2]: https://www.rfc-editor.org/rfc/rfc9380.html "RFC 9380: Hashing to Elliptic Curves"
[3]: https://datatracker.ietf.org/doc/draft-irtf-cfrg-pairing-friendly-curves/ "
            
    
        draft-irtf-cfrg-pairing-friendly-curves-12 - Pairing-Friendly Curves
    

        "
[4]: https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-pairing-friendly-curves-11 "
            
                draft-irtf-cfrg-pairing-friendly-curves-11
            
        "

Egge21M:
I would still request changes due to the new Mint Scalar Construction section.

The formula `a = (H_G1 * a_raw) mod BLS_FR_ORDER` does not make the resulting integer representative generally divisible by `H_G1`; reducing modulo the prime subgroup order destroys the property needed to annihilate torsion on arbitrary full-curve G1 inputs. Since `B_` is attacker-controlled input to the signing oracle, the spec should make mint-side subgroup validation of `B_` a MUST before computing `C_ = a·B_`.

Suggested fix: remove the normative Mint Scalar Construction section and replace the critical note with an unambiguous requirement:

"Mints MUST deserialize and validate v3 `B_` using canonical compressed BLS12-381 G1 encoding, and MUST reject infinity, non-curve points, non-canonical encodings, and points not in the prime-order G1 subgroup before signing."

The other updates look good: batch rejection sampling, NUT-13 v3 rejection sampling, distinct v3 amount keys, Y/K vectors, and batch challenge/weight vectors are all substantial improvements.

robwoodgate:
@Egge21M - Thank you for running the review bot. All issues raised have been addressed and this should be ready to review now.

cc: @callebtc @a1denvalu3 @thesimplekid

02.md
```diff
@@ -77,10 +112,11 @@ An example implementation in Python:
 ```python
 def derive_keyset_id_v2(keys: Dict[int, PublicKey], unit: str, input_fee_ppk: Optional[int], final_expiry: Optional[int]) -> str:
+    # v.serialize() returns the compressed secp256k1 bytes; .hex() yields the lowercase 66-char hex.
     keyset_id_bytes = b",".join(
-        [f"{k}:{v.serialize()}".encode("utf-8") for k, v in sorted_keys.items()]
+        [f"{k}:{v.serialize().hex()}".encode("utf-8") for k, v in sorted_keys.items()]
```
a1denvalu3, line 117: is this a correction or are we changing how we derive keysets ID v2?

02.md
```diff
@@ -77,10 +112,11 @@ An example implementation in Python:
...
     )
-    keyset_id_bytes += f"|unit:{unit}".encode("utf-8")
+    keyset_id_bytes += f"|unit:{unit.lower()}".encode("utf-8")
```
a1denvalu3, line 119: same here

02.md
robwoodgate: Unless I've misunderstood the python, it is just a correction to the example code to bring it in line with spec. Clanker says:

```
v.serialize().hex() fixes the Python example to hash amount:publickey_hex;
plain v.serialize() inside an f-string would render Python bytes syntax like b'...', which is the wrong preimage
```

02.md
robwoodgate: This is also intended to bring the python example into line with the existing spec. Although we assume unit is normalized lowercase elsewhere, I think the example should explicitly normalize.

00.md
```diff
@@ -10,7 +10,144 @@ This document details the notation and models used throughout the specification
 - Receiving user: `Carol`
 - Mint: `Bob`
 
-# Blind Diffie-Hellmann key exchange (BDHKE)
+Cashu uses Blind Diffie-Hellmann key exchange (BDHKE) for its blind signature protocol. Two variants are defined: a pairing-based scheme on BLS12-381 (the current protocol) and a legacy scheme on secp256k1 (deprecated).
```
callebtc:
```suggestion
Cashu uses a blind BLS signatures scheme on BLS12-381. The deprecated Blind Diffie-Hellmann key exchange (BDHKE) scheme was based on secp256k1 (legacy).
```

00.md
```diff
@@ -10,7 +10,144 @@ This document details the notation and models used throughout the specification
...
+
+# Pairing-based BDHKE (BLS12-381)
```
callebtc:
```suggestion
# Pairing-based BLS blind signatures
```

.DS_Store
robwoodgate, line 1: Stray file

xx+1.md
robwoodgate, line 1: @callebtc - one of your commits added this. I'm not sure it should be in this PR, but didn't want to remove arbitrarily.

robwoodgate (commented):
@callebtc - your recent commit seem to have included a bunch of stray files

robwoodgate:
Have resolved the overlap in symbol definitions as `r` (blinding factor) and `n` (SECP group order) were already defined.

- Weights: `r_i` is now `w_i`
- BLS Curve Order: `r` (was briefly changed to `n`) is now consistently `BLS_FR_ORDER`

13.md
```diff
@@ -37,7 +38,27 @@ The HMAC-SHA256 KDF is built as the following:
...
+message_v3 = b"Cashu_KDF_HMAC_SHA256" || keyset_id_bytes || counter_k_bytes || 0x01 || u32_BE(attempt)
+hmac_digest = HMAC_SHA256(seed, message_v3)
+x = OS2IP(hmac_digest)
+if x == 0 or x >= BLS_FR_ORDER:
+    continue   # try next attempt
+blinding_factor = hmac_digest        # 32 bytes; equals OS2IP⁻¹(x) since x < 2^256
+break
+```
+
+`attempt` is encoded as an unsigned 32-bit big-endian integer appended to the V2 message format (after the `0x01` derivation-type byte). With ≈ 0.45 acceptance per iteration, the expected number of attempts is ≈ 2.2; needing many retries is highly unlikely.
```
a1denvalu3:
```suggestion
`attempt` is encoded as an unsigned 32-bit big-endian integer appended to the V2 message format (after the `0x01` derivation-type byte).
```

13.md
```diff
@@ -51,52 +72,57 @@ Python:
...
     if keyset_version == "00":
-        # Use legacy BIP32 derivation for version 00
         return derive_secret_and_r_bip32(seed, keyset_id, counter_k)
     elif keyset_version == "01":
-        # Use HMAC-SHA256 derivation for version 01
-        return derive_secret_and_r_hmac(seed, keyset_id, counter_k)
+        # secp256k1 keyset: modular reduction (n ≈ 2^256, bias is ~2^-128)
+        return derive_secret_and_r_hmac_v2(seed, keyset_id, counter_k)
+    elif keyset_version == "02":
+        # BLS12-381 keyset: rejection sampling (BLS_FR_ORDER ≈ 0.45·2^256, mod would bias ~7.5%)
```
a1denvalu3:
```suggestion
```

13.md
```diff
@@ -51,52 +72,57 @@ Python:
...
-    # Determine keyset version from first two characters
     keyset_version = keyset_id[:2]
 
     if keyset_version == "00":
-        # Use legacy BIP32 derivation for version 00
         return derive_secret_and_r_bip32(seed, keyset_id, counter_k)
     elif keyset_version == "01":
-        # Use HMAC-SHA256 derivation for version 01
-        return derive_secret_and_r_hmac(seed, keyset_id, counter_k)
+        # secp256k1 keyset: modular reduction (n ≈ 2^256, bias is ~2^-128)
```
a1denvalu3:
```suggestion
```

00.md
```diff
@@ -10,7 +10,149 @@ This document details the notation and models used throughout the specification
...
+
+> _Applies to keysets with version byte `02` (see [NUT-02][02])._
+
+## Groups and Scalars
+
+- `G1` generator point of BLS12-381 G1 curve group of prime order `BLS_FR_ORDER` (where `Y`, `B_`, `C_`, `C` live); points come from `hash_to_curve_G1`.
+- `G2` generator point of BLS12-381 G2 of the same order
+- `Fr` finite field of integers modulo `BLS_FR_ORDER`; scalars live here.
+
+`BLS_FR_ORDER` is the order of the BLS12-381 prime-order subgroup (the order of `Fr`):
```
a1denvalu3: Redundant
```suggestion
```

00.md
```diff
@@ -10,7 +10,149 @@ This document details the notation and models used throughout the specification
...
+Cashu uses a blind BLS signatures scheme on BLS12-381. The deprecated Blind Diffie-Hellmann key exchange (BDHKE) scheme was based on secp256k1 (legacy).
+
+# Pairing-based BLS blind signatures
+
+> _Applies to keysets with version byte `02` (see [NUT-02][02])._
+
+## Groups and Scalars
+
+- `G1` generator point of BLS12-381 G1 curve group of prime order `BLS_FR_ORDER` (where `Y`, `B_`, `C_`, `C` live); points come from `hash_to_curve_G1`.
+- `G2` generator point of BLS12-381 G2 of the same order
```
a1denvalu3: You are naming `G1` and `G2` both the generators and the groups they generate.

00.md
```diff
@@ -10,7 +10,149 @@ This document details the notation and models used throughout the specification
...
+### Types
+
+Values are one of:
+
+- Scalars: integers in `Fr` (the prime field of order `BLS_FR_ORDER`)
+- G1 points: compressed 48 bytes / 96 hex characters
+- G2 points: compressed 96 bytes / 192 hex characters
+- Byte sequences
+
+Compressed point encoding follows the standard BLS12-381 serialisation. Hex encodings in JSON fields are lowercase, with no `0x` prefix and the fixed widths above. The identity point (point at infinity) is never a valid signature, mint key, or commitment.
```
a1denvalu3: What is a commitment in this context?

00.md
```diff
@@ -10,7 +10,149 @@ This document details the notation and models used throughout the specification
...
+### `hash_to_curve_G1(x: bytes) -> G1 point Y`
+
+Maps `x` deterministically to a point in `G1` using the BLS12-381 G1 random-oracle suite defined in [RFC 9380](https://www.rfc-editor.org/rfc/rfc9380) §8.8.1 (`BLS12381G1_XMD:SHA-256_SSWU_RO_`), with Cashu domain separation tag (DST):
+
+```
+DST = b"CASHU_BLS12_381_G1_XMD:SHA-256_SSWU_RO_"
+```
+
+> [!NOTE]
+> Under the discrete-log hardness assumption, `hash_to_curve_G1(x)` is one-way and yields a NUMS point in `G1`, meaning no one is assumed to know a scalar `y` such that `Y = y·G1`.
```
a1denvalu3:
```suggestion
```

00.md
robwoodgate: Is it redundant? It's the core definition for `BLS_FR_ORDER`. Otherwise the number below it is just a number.

00.md
```diff
@@ -10,7 +10,149 @@ This document details the notation and models used throughout the specification
...
+```
+h = SHA256(challenge || u32_BE(i) || u32_BE(ctr))
+x = OS2IP(h)
+if x == 0 or x >= BLS_FR_ORDER:
+    continue   # try next ctr
+w_i = x
+```
+
+> [!IMPORTANT]
+> Rejection sampling **MUST** be used to avoid bias. `BLS_FR_ORDER ≈ 0.45 · 2^256`, so each iteration succeeds with probability ≈ 0.45 (about 2.2 expected iterations per proof). This yields `w_i` uniformly distributed over `Fr*`. Plain `mod BLS_FR_ORDER` would over-represent some values by ~36%.
```
a1denvalu3: Implementors don't care

```suggestion
```

00.md
```diff
@@ -10,7 +10,149 @@ This document details the notation and models used throughout the specification
...
+
+4. Group items by distinct mint key `K_k` and check:
+
+```
+e( Σ_i w_i·C_i , G2 ) == Π_k e( Σ_{i:K_i=K_k} w_i·Y_i , K_k )
+```
+
+Implementations **SHOULD** evaluate the equation with a single call to a multi-pairing primitive; this performs one final exponentiation for the whole equation instead of one per pairing.
```
a1denvalu3:
```suggestion
Implementations **SHOULD** evaluate `Π_k e( Σ_{i:K_i=K_k} w_i·Y_i , K_k )` with a single call to a multi-miller loop primitive, at the end of which they **SHOULD** perform the final exponentiation over the accumulated miller loop.
```

00.md
```diff
@@ -10,7 +10,149 @@ This document details the notation and models used throughout the specification
...
+
+Typical batch scopes:
+
+- A wallet receiving a token batches all v3 proofs across that token.
+- A mint batches all v3 input proofs of a single `/v1/swap` or `/v1/melt` request.
+
+A batch of one proof reduces to the single-proof pairing check above.
+
+The weights are public and deterministic; security does not rely on their secrecy. Per-proof weights are required because, without them, an attacker holding one aggregated signature `C' = a·(Y_1+Y_2)` could split it into two forgeries that both verify under a sum check.
```
a1denvalu3: Details not relevant:

```suggestion
The weights are public and deterministic; security does not rely on their secrecy.
```

00.md
```diff
@@ -10,7 +10,149 @@ This document details the notation and models used throughout the specification
...
+Implementations **MUST** deserialise all v3 `B_`, `C_`, `C`, and `K` values using the canonical compressed BLS12-381 encoding, and **MUST** reject:
+
+- non-canonical encodings,
+- the identity (point at infinity),
+- points not on the curve,
+- points not in the prime-order subgroup of `G1` (for `B_`, `C_`, `C`) or `G2` (for `K`).
+
+> [!IMPORTANT]
+> To defend against the [small-subgroup attack][lim-lee], mints **MUST** perform full validation on `B_` before computing `C_ = a·B_`; wallets and receivers **MUST** perform it on mint public keys `K`, on `C_` before unblinding, and on signatures `C` before any pairing check.
+> Use the library's prime-order subgroup predicate, such as `KeyValidate` / `in_g1` / `in_g2` (blst) or `isTorsionFree()` (`@noble/curves`).
```
a1denvalu3: Redundant, because you already specify the steps to validation:



```suggestion
```

00.md
a1denvalu3: You say it right above: "[...] of prime order `BLS_FR_ORDER`"

00.md
```diff
@@ -10,7 +10,149 @@ This document details the notation and models used throughout the specification
...
+The weights are public and deterministic; security does not rely on their secrecy. Per-proof weights are required because, without them, an attacker holding one aggregated signature `C' = a·(Y_1+Y_2)` could split it into two forgeries that both verify under a sum check.
+
+### Point Validation
+
+Implementations **MUST** deserialise all v3 `B_`, `C_`, `C`, and `K` values using the canonical compressed BLS12-381 encoding, and **MUST** reject:
+
+- non-canonical encodings,
+- the identity (point at infinity),
+- points not on the curve,
+- points not in the prime-order subgroup of `G1` (for `B_`, `C_`, `C`) or `G2` (for `K`).
```
a1denvalu3, line 139: If I am not mistaken `G1` is already the prime-order subgroup. What you're referring to is points on the BLS-13-381 curve that are not fully in `G1`. i.e. they have a cofactor component.

00.md
robwoodgate: Good spot. Redefined the generators to `g1/g2` per BLS conventions

00.md
robwoodgate: I removed the number and reordered the section

00.md
robwoodgate: A blinded message - I changed it to be clearer

00.md
robwoodgate: They should. The bias is significant.

00.md
robwoodgate: It's both - no mixing groups (ie B_ in G2 is invalid) - and no cofactors.

00.md
robwoodgate: If I add a validation line for no cofactors, then yes, we can remove this. But it's such a significant bear trap for BLS, i think we should highlight it somehow

00.md
a1denvalu3: They just implement it how it's written in the pseudo-code.

00.md
a1denvalu3: I am just suggesting you remove `of`. "points not in the prime-order subgroup `G1`".
Or "valid points on the curve, not in the prime-order subgroup `G1`".

00.md
robwoodgate: I shortened the message, but think we should keep the warning because implementors coming from SECP may not appreciate the bias in BLS

00.md
robwoodgate: I refactored and shortened the message

00.md
robwoodgate: I added cofactors to the validation list and shorted the lim lee warning

00.md
robwoodgate: removed

a1denvalu3 (approved):

TheMhv:
Why change the protocol cryptographic algorithm? What the benefits of this change?

02.md
```diff
@@ -56,15 +56,50 @@ Notice that since transactions can spend inputs from different keysets, the sum
...
+V3 keyset IDs are 33 byte hex strings with version byte `02`. V3 identifies keysets that use the BLS12-381 blinding protocol defined in [Pairing-based BDHKE (BLS12-381)](00.md#pairing-based-bdhke-bls12-381).
+
+Keyset IDs are derived from public data. To derive the keyset ID of a keyset, execute the following steps:
+
+```
+1 - sort public keys by their amount in ascending numerical order
+2 - concatenate each amount and its corresponding lowercase compressed BLS12-381 G2 public key hex string (as "amount:publickey_hex") to a single byte array, separating each pair with a comma (",")
+3 - add the lowercase UTF8-encoded unit string prefixed with "|unit:" to the byte array (e.g. "|unit:sat")
+4 - If input_fee_ppk is specified and non-zero, add the UTF8-encoded string prefixed with "|input_fee_ppk:" (e.g. "|input_fee_ppk:100"). If input_fee_ppk is omitted, null, or 0, it MUST be omitted from the preimage.
+5 - If a final expiration is specified, add the UTF8-encoded string prefixed with "|final_expiry:" (e.g. "|final_expiry:1896187313")
```
orangeshyguy21, line 72: While v3's derivation is still open it could be worth baking a deactivation timestamp, such as `active_expiry`, into the ID alongside `final_expiry`

The motivation for this is the PoL work in [#388](https://github.com/cashubtc/nuts/pull/388). Regular keyset rotations are what make auditing a mint's stated balances effective at the client level. If the deactivation time could be added to the keyset derivation the spec for PoL could take advantage of this by reducing or eliminating mechanisms for publishing timestamps and proving rotations have been honored.

It should be optional and omitted-when-null like `final_expiry`.

Raising this now because the change alone would bump the keyset version, so it's inexpensive to add while v3 is in the works.

robwoodgate: So just for clarification, you're proposing:

`active_expiry` = date after which no new proofs will be minted, the keyset becomes inactive

vs

`final_expiry` = date after which redemption of outstanding proofs may end, rendering them worthless.

callebtc: noting also that we're using a different separator | for keyset IDs than for any other messages in the protocol

robwoodgate: > noting also that we're using a different separator | for keyset IDs than for any other messages in the protocol

Good point!

robwoodgate: In offline discussions, it was noted that baking `final_expiry` into the keyset derivation is unenforceable, as it is just metadata the mint can ignore. It also means the expiry date MUST be known before creation and CANNOT be added afterwards. The suggestion was we remove it while keyset v3 is still in flux. 

@orangeshyguy21 - `active_expiry` is also just advisory metadata a mint can ignore. Does it carry any weight for PoL if a mint does not in fact deactivate a keyset from the date promised? Ie does baking `active_expiry` into the keyset ID still act as a "proof the mint did not do what it promised at keyset init"?

orangeshyguy21: It would act as proof yes, since the date of rotation is locked to the keyset and cannot be altered without changing the keyset wholesale. Wallets cannot stop a mint from lying, but they can prove when it has: that's the whole spirit of a PoL.

However, if the direction is to reduce metadata used in derivation that would make this addition moot. The idea here is to add onto an established standard. But if that standard is being walked back / eliminated, this idea should die with it as well.

Happy to let the fate of `final_expiry` guide the decision.

robwoodgate:
# BREAKING CHANGE

Length framed keyset ID derivation added in 68840ccd6bd71fa2c0a104353d741c78505fd7d1

This also removes `final_expiry` from the derivation as agreed

02.md
robwoodgate: Keyset derivation updated - removed final_expiry and length framed

tests/13-tests.md
```diff
@@ -92,6 +92,26 @@ The corresponding blinding factors `r` are:
...
+
+V3 secret derivation is identical to V2. V3 blinding factor derivation uses rejection sampling against `BLS_FR_ORDER` instead of modular reduction; the V3 HMAC input also includes a 4-byte big-endian `attempt` counter (see [NUT-13 V3 Blinding Factor](../13.md#v3-blinding-factor)).
+
+The vector below is chosen so that attempts `0-2` produce `x >= BLS_FR_ORDER` and are rejected; `attempt=3` is accepted. Implementations that omit the rejection loop will compute a different `blinding_factor` and fail this vector.
+
+Using [NUT-13](../13.md) derivation procedure for V3:
+
+```json
+{
+  "seed_utf8": "nut13 v3 test seed",
```
robwoodgate, line 105: Should the v3 vectors derive from a 64-byte seed (with the rejection-sampling property re-searched)?

PR #372, opened 2026-05-21
NUT-17: Simplify NUT-17 subscription kinds

callebtc:
# Implementations

- Cashu-TS: https://github.com/cashubtc/cashu-ts/pull/668
- CDK: https://github.com/cashubtc/cdk/pull/1988
- Nutshell: https://github.com/cashubtc/nutshell/pull/1013

## Summary
- replace method-specific `SubscriptionKind` values with generic `mint_quote` and `melt_quote`
- keep `proof_state` unchanged
- update NUT-17 command examples to advertise quote subscriptions per method-unit pair

## Rationale
The subscription kind is now independent of the payment method, which makes the websocket API extensible to custom payment methods without needing method-specific quote kinds in the enum.


Egge21M:
I like it. However we should make sure that mints are fine with signaling support only on that high level instead of on a method level

robwoodgate (approved):
ACK. LGTM

a1denvalu3 (approved):

robwoodgate:
If we decide not to go this route for any reason, then we need to add `online_mint_quote` / `online_melt_quote` to the list, and remember to update it when new first class methods are added.

But with the companion PR below, this should allow mints and wallets to track existing and custom method quotes via websockets in a sane way:

- https://github.com/cashubtc/nuts/pull/387

PR #373, opened 2026-05-21
NUT-20: add deterministic NUT-20 quote locking derivation

Egge21M:
## Implementations

- Coco: https://github.com/cashubtc/coco/pull/184
- Cashu-TS: https://github.com/cashubtc/cashu-ts/pull/697
- CDK: TBC
- Nutshell: TBC

## Summary

Adds deterministic wallet seed derivation guidance for NUT-20 quote locking keys using the Cashu SLIP-0044 coin type and a NUT-20-specific BIP32 path.

Updates the NUT-20 test vectors with the first five derived compressed secp256k1 public keys for the documented mnemonic.

## Validation

- Compared `Egge21M:nut-20-locking` against `cashubtc/nuts:main` with the GitHub connector: 1 commit ahead, 0 behind
- Reviewed the local diff for `20.md` and `tests/20-test.md`
- No local test runner or workflow config is present in this checkout

### DIFF
--- a/20.md
+++ b/20.md
@@ -39,6 +39,26 @@ with the requested `amount`,`unit`, and `description` according to [NUT-04][04].
 >
 > **Privacy:** To prevent the mint from being able to link multiple mint quotes, wallets **SHOULD** generate a unique public key for each mint quote request.
 
+### Deterministic quote locking key derivation
+
+Wallets can deterministically generate private keys for NUT-20 quote locking from their wallet seed.
+
+The following BIP32 derivation path is used to derive the private key:
+
+```
+m/129373'/20'/0'/0'/{counter}
+```
+
+Where:
+
+- `129373'` is the registered SLIP-0044 coin type for Cashu.
+- `20'` is the wallet derivation index for NUT-20 quote locking keys.
+- `{counter}` is an incrementing, non-hardened BIP32 child index.
+
+The wallet **SHOULD** use a new counter value for each mint quote request. The NUT-20 quote locking counter is independent from NUT-13 keyset counters and other wallet derivation counters.
+
+The wallet uses the compressed secp256k1 public key derived from the private key as the quote `pubkey`. This allows a wallet to restore the private key needed to sign for unpaid or paid NUT-20 locked mint quotes during a recovery process.
+
 The mint `Bob` then responds with a `PostMintQuoteBolt11Response`:
 
 ```json
--- a/20.md
+++ b/20.md
@@ -51,7 +51,7 @@ m/129373'/20'/0'/0'/{counter}
 
 Where:
 
-- `129373'` is the registered SLIP-0044 coin type for Cashu.
+- `129373'` is the Cashu wallet derivation namespace.
 - `20'` is the wallet derivation index for NUT-20 quote locking keys.
 - `{counter}` is an incrementing, non-hardened BIP32 child index.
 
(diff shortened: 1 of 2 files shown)


20.md
```diff
@@ -39,6 +39,26 @@ with the requested `amount`,`unit`, and `description` according to [NUT-04][04].
...
+
+The following BIP32 derivation path is used to derive the private key:
+
+```
+m/129373'/20'/0'/0'/{counter}
+```
+
+Where:
+
+- `129373'` is the registered SLIP-0044 coin type for Cashu.
```
callebtc: wut

Egge21M: Fixed :)

robwoodgate:
As [per my comment](https://github.com/cashubtc/nuts/pull/331#pullrequestreview-4380936838) on the related deterministic P2PK key spec ( #331  ), unless there is an overwhelmingly strong reason to use BIP32, then a HMAC-SHA256 KDF is much faster, unless BIP-32 is carefully handled.

```
HMAC-SHA256                                       ~165 µs    1.0×
```

robwoodgate:
I have proposed a [HMAC-SHA256 KDF derived alternative](https://github.com/cashubtc/nuts/pull/384), which would close this PR

Egge21M:
> I have proposed a [HMAC-SHA256 KDF derived alternative](https://github.com/cashubtc/nuts/pull/384), which would close this PR

Iirc this has been discussed in the past in Leitos original PR for P2PK locking keys. Using BIP32 might be a tad slower, but it leaves the possibility to build something with the extended public keys. Imagine a watch-only wallet for mint quotes

robwoodgate:
> > I have proposed a [HMAC-SHA256 KDF derived alternative](https://github.com/cashubtc/nuts/pull/384), which would close this PR
> 
> 
> 
> Iirc this has been discussed in the past in Leitos original PR for P2PK locking keys. Using BIP32 might be a tad slower, but it leaves the possibility to build something with the extended public keys. Imagine a watch-only wallet for mint quotes

Yeah, am not convinced it would not be a huge linkability issue compromising privacy. I mentioned it in my replacement pr

a1denvalu3 (approved):

Egge21M:
> Yeah, am not convinced it would not be a huge linkability issue compromising privacy. I mentioned it in my replacement pr

This is both a bug and a feature. I think we have general consensus about a quote-lookup-by-pubkey API, so this would literally allow watch-only / notification / aggregation services.
Putting that aside: The performance gain is marginal. Alignment with the spec is given, as P2PK keys are already BIP32. CDK and Coco both already ship BIP32 derivation for P2PK locks.

robwoodgate:
> > Yeah, am not convinced it would not be a huge linkability issue compromising privacy. I mentioned it in my replacement pr
> 
> 
> 
> This is both a bug and a feature. I think we have general consensus about a quote-lookup-by-pubkey API, so this would literally allow watch-only / notification / aggregation services.
> 
> Putting that aside: The performance gain is marginal. Alignment with the spec is given, as P2PK keys are already BIP32. CDK and Coco both already ship BIP32 derivation for P2PK locks. 

Yeah, unfortunately that was kind of a premature shipping before spec sign off.

We moved away from bip32 for keysets v2+. Baking that dep into future stuff means apps will have to carry it a long time more.

The performance gain is only marginal if you are batch deriving. If you have to derive sporadically for any reason performance drops off a cliff

robwoodgate (approved):
Following offline discussion, and the fact this is already live, I've closed #384.

robwoodgate:
Cashu-TS support now added:
- https://github.com/cashubtc/cashu-ts/pull/697

PR #374, opened 2026-05-22
feat(nut04/05): add method_name to [Mint|Melt]MethodSetting

robwoodgate:
- [x] Cashu-TS - https://github.com/cashubtc/cashu-ts/pull/672 and https://github.com/cashubtc/cashu-ts/pull/707
- [x] CDK - https://github.com/cashubtc/cdk/pull/2120
- [ ] Nutshell - https://github.com/cashubtc/nutshell/pull/1063

This PR adds an optional `method_name` param to the NUT-04 `MintMethodSetting`, and NUT-05 `MeltMethodSetting`.

The motivation is to make it easy for wallets to support custom payment methods in their UI.

##

### DIFF
--- a/04.md
+++ b/04.md
@@ -140,7 +140,7 @@ The settings for this NUT indicate the supported method-unit pairs for minting. 
 }
 ```
 
-`min_amount` and `max_amount` indicate the minimum and maximum amount for an operation of this method-unit pair. `options` are method-specific and can be defined in method-specific NUTs. `method_name` is a human-readable name for the method.
+`min_amount` and `max_amount` indicate the minimum and maximum amount for an operation of this method-unit pair. `options` are method-specific and can be defined in method-specific NUTs. `method_name` is a human-readable name for the payment method.
 
 ## Unblinding Signatures
 
--- a/04.md
+++ b/04.md
@@ -133,13 +133,14 @@ The settings for this NUT indicate the supported method-unit pairs for minting. 
 {
   "method": <str>,
   "unit": <str>,
+  "method_name": <str|null>,
   "min_amount": <int|null>,
   "max_amount": <int|null>,
   "options": <Object|null>
 }
 ```
 
-`min_amount` and `max_amount` indicate the minimum and maximum amount for an operation of this method-unit pair. `options` are method-specific and can be defined in method-specific NUTs.
+`min_amount` and `max_amount` indicate the minimum and maximum amount for an operation of this method-unit pair. `options` are method-specific and can be defined in method-specific NUTs. `method_name` is a human-readable name for the method.
 
 ## Unblinding Signatures
 
--- a/04.md
+++ b/04.md
@@ -140,7 +140,9 @@ The settings for this NUT indicate the supported method-unit pairs for minting. 
 }
 ```
 
-`min_amount` and `max_amount` indicate the minimum and maximum amount for an operation of this method-unit pair. `options` are method-specific and can be defined in method-specific NUTs. `method_name` is a human-readable name for the payment method.
+`min_amount` and `max_amount` indicate the minimum and maximum amount for an operation of this method-unit pair. `options` are method-specific and can be defined in method-specific NUTs.
+
+`method_name` is a human-readable name for the payment method. If `null` or omitted, it is derived from `method` by replacing `_` and `-` with spaces and title-casing each word (e.g. `bolt11` -> `Bolt11`, `apple-pay` -> `Apple Pay`).
 
 ## Unblinding Signatures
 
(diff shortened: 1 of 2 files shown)


04.md
```diff
@@ -133,13 +133,14 @@ The settings for this NUT indicate the supported method-unit pairs for minting.
...
   "unit": <str>,
+  "method_name": <str|null>,
   "min_amount": <int|null>,
   "max_amount": <int|null>,
   "options": <Object|null>
 }
 ```
 
-`min_amount` and `max_amount` indicate the minimum and maximum amount for an operation of this method-unit pair. `options` are method-specific and can be defined in method-specific NUTs.
+`min_amount` and `max_amount` indicate the minimum and maximum amount for an operation of this method-unit pair. `options` are method-specific and can be defined in method-specific NUTs. `method_name` is a human-readable name for the method.
```
callebtc:
```suggestion
`min_amount` and `max_amount` indicate the minimum and maximum amount for an operation of this method-unit pair. `options` are method-specific and can be defined in method-specific NUTs. `method_name` is a human-readable name for the payment method.
```

05.md
```diff
@@ -236,13 +236,14 @@ The mint's settings for this NUT indicate the supported method-unit pairs for me
...
   "unit": <str>,
+  "method_name": <str|null>,
   "min_amount": <int|null>,
   "max_amount": <int|null>,
   "options": <Object|null>
 }
 ```
 
-`min_amount` and `max_amount` indicate the minimum and maximum amount for an operation of this method-unit pair. `options` are method-specific and can be defined in method-specific NUTs.
+`min_amount` and `max_amount` indicate the minimum and maximum amount for an operation of this method-unit pair. `options` are method-specific and can be defined in method-specific NUTs. `method_name` is a human-readable name for the method.
```
callebtc:
```suggestion
`min_amount` and `max_amount` indicate the minimum and maximum amount for an operation of this method-unit pair. `options` are method-specific and can be defined in method-specific NUTs. `method_name` is a human-readable name for the payment method.
```

callebtc (approved):

Kelbie:
Does it make sense to add localization here or is that too much? If wallets have language configuration we could pick the relevant one for the user.

robwoodgate:
> Does it make sense to add localization here or is that too much? If wallets have language configuration we could pick the relevant one for the user.

I think too much for an info setting - could run to hundreds of entries. Wallet should make a policy decision to translate / ignore / decide their own names.

thesimplekid:
I do not think we should localization for the reason Rob stated. However, should we give each method specified in the nut a name (bolt11, bolt12, onchain) so that mints all use the same one and then wallets can display this however they want. I think it would be confusing if cdk names bolt11 "lighting" for example and nutshell calls it "bolt11", wallets would then have to map these two to the same thing.

KvngMikey (approved):

KvngMikey:
> I do not think we should localization for the reason Rob stated. However, should we give each method specified in the nut a name (bolt11, bolt12, onchain) so that mints all use the same one and then wallets can display this however they want. I think it would be confusing if cdk names bolt11 "lighting" for example and nutshell calls it "bolt11", wallets would then have to map these two to the same thing.

i'm thinking we can do:

Bolt 11 -> Lightning 
Bolt 12 -> Offers
onchain -> OnChain

so cdk and nutshell are compliant to these.

thesimplekid:
> > I do not think we should localization for the reason Rob stated. However, should we give each method specified in the nut a name (bolt11, bolt12, onchain) so that mints all use the same one and then wallets can display this however they want. I think it would be confusing if cdk names bolt11 "lighting" for example and nutshell calls it "bolt11", wallets would then have to map these two to the same thing.
> 
> i'm thinking we can do:
> 
> Bolt 11 -> Lightning Bolt 12 -> Offers onchain -> OnChain
> 
> so cdk and nutshell are compliant to these.

Why not just bolt11, bolt12, and onchain, wallets can decide to call it something else on the user facing side if they want but don't think we need to add other naming to the spec?

KvngMikey:
> Why not just bolt11, bolt12, and onchain, wallets can decide to call it something else on the user facing side if they want but don't think we need to add other naming to the spec?

that works also, just thought we wanted something extra.

robwoodgate:
> should we give each method specified in the nut a name (bolt11, bolt12, onchain) so that mints all use the same one and then wallets can display this however they want. I think it would be confusing if cdk names bolt11 "lighting" for example and nutshell calls it "bolt11", wallets would then have to map these two to the same thing.

Yes - wallets need predictable strings for translation, but rather than update all the first class nut methods (23/25/30), I added a one liner default rule.

robwoodgate:
While we are here, anticipating custom methods, we should specify the allowed characters for `method`.

Cashu-TS only allows "a-z0-9-_" (base64url safe), so I used that as the template.

thesimplekid (approved):

PR #377, opened 2026-05-24
Align NUT-04 mint quote accounting with amount_paid /   amount_issued

Egge21M:
# Implementations

- [x] cdk: https://github.com/cashubtc/cdk/pull/2119
- [x] cashu-ts: https://github.com/cashubtc/cashu-ts/pull/698
- [x] nutshell:
    - https://github.com/cashubtc/nutshell/pull/1044
    - https://github.com/cashubtc/nutshell/pull/1070

## Summary

This updates NUT-04 to make amount_paid and amount_issued the
  canonical way to track mint quote accounting, matching the reusable
  quote model already used by NUT-25 and NUT-30.

  The deprecated state field is kept as an optional compatibility
  field for single-use quotes, with derivation rules documented from
  amount_paid and amount_issued. This also adds updated_at to mint
  quote responses so clients can avoid applying stale quote updates.

  Changes:

  - Define common amount_paid, amount_issued, and updated_at fields in
    NUT-04 mint quote responses
  - Document mintable amount as amount_paid - amount_issued
  - Deprecate state while preserving compatibility guidance
  - Add stale-update handling guidance using updated_at
  - Update NUT-20, NUT-23, NUT-25, NUT-29, and NUT-30 examples/schemas
    to match


robwoodgate:
In offline discussion, it was proposed to add a `version` param to mint quotes.

This would allow us to make the changes (eg remove `state`, evolve NUT-20 signing) cleanly, as wallets can adjust their handling appropriately to the version.

I'd recommend we start at `version: 02`, so that any quote with unspecified `version` param is treated as an impleict`version: 01` quote with `state` param.

robwoodgate:
The `updated_at` param is a good idea, both for monotonic state tracking as noted, as well as being potentially useful for tracking stalled or failed payments.

a1denvalu3:
Agreed.

robwoodgate (approved):
ACK

a1denvalu3 (approved):

Egge21M:
Chatted with some folks offline. They raised the point that the combination of `amount_issued` and `amount_paid` is already enough to allow monotonic state updates (e.g. `new_amount_issued + new_amount_paid > old_amount_issued + old_amount_paid`). However this is less explicit and mints should not have any issue adding `updated_at`.

Should I kick it, or keep it?

robwoodgate:
> Chatted with some folks offline. They raised the point that the combination of `amount_issued` and `amount_paid` is already enough to allow monotonic state updates (e.g. `new_amount_issued + new_amount_paid > old_amount_issued + old_amount_paid`). However this is less explicit and mints should not have any issue adding `updated_at`.
> 
> Should I kick it, or keep it?

I think it has utility - whilst the combination does tell you if the state has changed, it doesn't tell you WHEN.

So unless payload size is a concern, it might be more useful than not, esp for consumer apps.

thesimplekid (approved):

asmogo (approved):

04.md
```diff
@@ -52,19 +52,47 @@ The mint `Bob` responds with a quote that includes some common fields for all me
...
+
+For single-use quotes, the deprecated finite `state` can be derived from `amount_paid` and `amount_issued`:
+
+- `"UNPAID"` if `amount_paid == 0` and `amount_issued == 0`
+- `"PAID"` if `amount_paid > amount_issued`
+- `"ISSUED"` if `amount_paid == amount_issued` and `amount_issued > 0`
+
+For reusable quotes, no finite `state` can fully represent whether the quote may receive future payments. Wallets **MUST** use `amount_paid` and `amount_issued` to determine the currently mintable amount.
+
+Mints **MUST** update `updated_at` whenever `amount_paid` or `amount_issued` changes. Mints **SHOULD** ensure that `updated_at` monotonically increases for each quote, even if multiple updates occur within the same timestamp resolution. Wallets that receive multiple responses for the same quote **MUST NOT** replace locally stored quote data with a response whose `updated_at` is lower than the latest processed value for that quote. Wallets **MUST NOT** decrease locally stored `amount_paid` or `amount_issued` values based on stale responses.
```
robwoodgate: Have implemented this in cashu-ts (cashubtc/cashu-ts#693). The derivation rules round-trip cleanly in both directions, so transition handling is straightforward.

The only wrinkle is how to handle the case when `updated_at` is absent (any pre-accounting mint). Suggest we add:

```suggestion
Mints **MUST** update `updated_at` whenever `amount_paid` or `amount_issued` changes. Mints **SHOULD** ensure that `updated_at` monotonically increases for each quote, even if multiple updates occur within the same timestamp resolution. Wallets that receive multiple responses for the same quote **MUST NOT** replace locally stored quote data with a response whose `updated_at` is lower than the latest processed value for that quote. Wallets **MUST NOT** decrease locally stored `amount_paid` or `amount_issued` values based on stale responses, or when the response has no `updated_at` value.
```

robwoodgate (approved):

04.md
Egge21M: TBH I am not sure how to handle this properly. This document is now written in a way that makes `updated_at` mandatory (e.g. `Mints **MUST** include`). Including this addition would mean covering off-spec mint behavior. Maybe the new fields should be added as optional an with `SHOULD` language instead?

robwoodgate: On reflection you are right. Handling out of spec older mints is not the protocol's concern.

Ignore my suggestion - leave it as it was.

04.md
```diff
@@ -52,19 +52,47 @@ The mint `Bob` responds with a quote that includes some common fields for all me
   "quote": <str>, // UUID v7
   "request": <str>,
   "unit":  <str_enum[UNIT]>,
+  "amount_paid": <int>,
+  "amount_issued": <int>,
+  "updated_at": <int>,
+  "state": <str_enum[STATE]>, // Deprecated, optional
```
thesimplekid: State should not be in nut04 that is in nut 23 as it is bolt11 specefic.

04.md
```diff
@@ -52,19 +52,47 @@ The mint `Bob` responds with a quote that includes some common fields for all me
...
-Where `quote` is the quote ID string in UUID v7 format, `request` is the payment request for the quote, and `unit` corresponds to the value provided in the request.
+Where:
+
+- `quote` is the quote ID in UUID v7 format
+- `request` is the payment request for the quote
+- `unit` corresponds to the value provided in the request
+- `amount_paid` is the total amount that has been paid to the mint for this quote and is eligible for minting, denominated in `unit`
+- `amount_issued` is the total amount of ecash that has been issued for this quote, denominated in `unit`
+- `updated_at` is a Unix timestamp integer indicating when the quote was last updated
+- `state` is a deprecated compatibility field
```
thesimplekid:
```suggestion
```

04.md
```diff
@@ -52,19 +52,47 @@ The mint `Bob` responds with a quote that includes some common fields for all me
...
+
+`amount_paid` and `amount_issued` are the canonical way to track the state of a mint quote. The deprecated `state` field **SHOULD** be included for single-use quotes for compatibility with older clients. Wallets **SHOULD** use `amount_paid` and `amount_issued` instead of `state` whenever these fields are present.
+
+For single-use quotes, the deprecated finite `state` can be derived from `amount_paid` and `amount_issued`:
+
+- `"UNPAID"` if `amount_paid == 0` and `amount_issued == 0`
+- `"PAID"` if `amount_paid > amount_issued`
+- `"ISSUED"` if `amount_paid == amount_issued` and `amount_issued > 0`
+
+For reusable quotes, no finite `state` can fully represent whether the quote may receive future payments. Wallets **MUST** use `amount_paid` and `amount_issued` to determine the currently mintable amount.
```
thesimplekid: Nut23 specefic

20.md
```diff
@@ -45,7 +45,10 @@ The mint `Bob` then responds with a `PostMintQuoteBolt11Response`:
 {
   "quote": <str>,
   "request": <str>,
-  "state": <str_enum[STATE]>,
+  "amount_paid": <int>,
+  "amount_issued": <int>,
+  "updated_at": <int>,
+  "state": <str_enum[STATE]>, // Deprecated, optional
```
thesimplekid, line 51:
```suggestion
```

Think state should be removed here as well as this should just extend the NUT04 type

thesimplekid (changes requested):

robwoodgate:
Cashu-TS implements in:

- https://github.com/cashubtc/cashu-ts/pull/693

20.md
robwoodgate: This is a bolt11 example, so state would exist here I think?

thesimplekid: Ah okay.

robwoodgate:
> Cashu-TS implements in:
> 
> * [feat(nut04/05)!: quote accounting and custom payment method base structs cashu-ts#693](https://github.com/cashubtc/cashu-ts/pull/693)

Sorry, now moved to:

- https://github.com/cashubtc/cashu-ts/pull/698

KvngMikey:
Implementation for Nutshell:
Merged:
https://github.com/cashubtc/nutshell/pull/1044

Open:
https://github.com/cashubtc/nutshell/pull/1070

04.md
```diff
@@ -52,19 +52,37 @@ The mint `Bob` responds with a quote that includes some common fields for all me
...
+
+The amount currently mintable for a quote is `amount_paid - amount_issued`. Mints **MUST NOT** issue ecash whose total output amount exceeds `amount_paid - amount_issued`. If a wallet mints less than the currently mintable amount, `amount_issued` only increases by the amount that was issued.
+
+`amount_paid` and `amount_issued` are the canonical way to track mint quote accounting. For reusable quotes, no finite status can fully represent whether the quote may receive future payments. Wallets **MUST** use `amount_paid` and `amount_issued` to determine the currently mintable amount.
+
+Mints **MUST** update `updated_at` whenever `amount_paid` or `amount_issued` changes. Mints **SHOULD** ensure that `updated_at` monotonically increases for each quote, even if multiple updates occur within the same timestamp resolution. Wallets that receive multiple responses for the same quote **MUST NOT** replace locally stored quote data with a response whose `updated_at` is lower than the latest processed value for that quote. Wallets **MUST NOT** decrease locally stored `amount_paid` or `amount_issued` values based on stale responses.
```
thesimplekid: I think there is a conflict here between "Mints **SHOULD** ensure that `updated_at` monotonically increases for each quote" and "Wallets that receive multiple responses for the same quote **MUST NOT** replace locally stored quote data with a response whose `updated_at` is lower than the latest processed value for that quote."

Since the mint only **SHOULD** ensure this but a want is relying on it being correct a mint not doing this correctly is not out of spec as its a **SHOULD** not a **MUST** but the wallet is stuck.

04.md
```diff
@@ -52,19 +52,37 @@ The mint `Bob` responds with a quote that includes some common fields for all me
...
 }
 ```
 
-Where `quote` is the quote ID string in UUID v7 format, `request` is the payment request for the quote, and `unit` corresponds to the value provided in the request.
+Where:
+
+- `quote` is the quote ID in UUID v7 format
+- `request` is the payment request for the quote
+- `unit` corresponds to the value provided in the request
+- `amount_paid` is the total amount that has been paid to the mint for this quote and is eligible for minting, denominated in `unit`
```
thesimplekid:
```suggestion
- `amount_paid` is the total amount that has been paid to the mint for this quote, denominated in `unit`
```

I think eligible for minting is the difference between the two.

04.md
```diff
@@ -52,19 +52,37 @@ The mint `Bob` responds with a quote that includes some common fields for all me
...
+- `amount_issued` is the total amount of ecash that has been issued for this quote, denominated in `unit`
+- `updated_at` is a Unix timestamp integer indicating when the quote was last updated
+
+Mints **MUST** include `amount_paid`, `amount_issued`, and `updated_at` in all mint quote responses. `amount_paid` and `amount_issued` **MUST** be non-negative integers, and `amount_issued` **MUST NOT** exceed `amount_paid`.
+
+The amount currently mintable for a quote is `amount_paid - amount_issued`. Mints **MUST NOT** issue ecash whose total output amount exceeds `amount_paid - amount_issued`. If a wallet mints less than the currently mintable amount, `amount_issued` only increases by the amount that was issued.
+
+`amount_paid` and `amount_issued` are the canonical way to track mint quote accounting. For reusable quotes, no finite status can fully represent whether the quote may receive future payments. Wallets **MUST** use `amount_paid` and `amount_issued` to determine the currently mintable amount.
```
thesimplekid: I think this could be removed. It makes sense in the context of we used to have a state. But I think we should write the nut as that never happened.

23.md
```diff
@@ -28,17 +28,30 @@ The mint responds with a `PostMintQuoteBolt11Response`:
...
 - `"ISSUED"` means that the quote has been paid and the ecash has been issued.
 
+Wallets **SHOULD** use `amount_paid` and `amount_issued` instead of `state` whenever these fields are present.
+
+For `bolt11` single-use quotes, `state` can be derived from `amount_paid` and `amount_issued`:
+
+- `"UNPAID"` if `amount_paid == 0` and `amount_issued == 0`
+- `"PAID"` if `amount_paid > amount_issued`
+- `"ISSUED"` if `amount_paid == amount_issued` and `amount_issued > 0`
+
```
thesimplekid: I don't think we need this. If the mint provides the amount paid and issued they should use that so there is no need to derive it.

04.md
Egge21M: I don't think it is a conflict. The mint one is about avoiding sending updates with the same timestamp (not lower) and the wallet one is about not accepting state updates with lower timestamps. Wallets are allowed to make decisions on how to handle updates that are the same, even though it SHOULD be avoided

robwoodgate: > Mints **SHOULD** ensure that `updated_at` monotonically increases for each quote, even if multiple updates occur within the same timestamp resolution.

I think @thesimplekid may be right. If a mint quote somehow emits (eg via WS) two atomic updates in same timestamp - eg:

initial state: `..., custom_field='', amount_paid = 0, updated_at = 100`
update 1: `..., custom_field='foo', amount_paid = 0, updated_at = 123`
update 2: ..., custom_field='foo',`amount_paid = 1000, updated_at = 123`

but update 1 arrives after, the wallet has no way to know which state to choose as neither is "lower". 

if mints **MUST** ensure that `updated_at` monotonically increases for each quote update, there is no ambiguity.

Egge21M: Oh I see what you mean. In my head wallets would have had a fallback logic for that kind of thing, but I agree rewording to MUST is much easier

robwoodgate: Though as noted [in a comment](https://github.com/cashubtc/nuts/pull/377#issuecomment-4646573747), a wallet CAN refuse to decrement amount_paid / amount_issued and infer the order.

It's just easier if mints ensure a single update within a timestamp resolution.

thesimplekid (approved):

callebtc (approved):

PR #375, opened 2026-05-22
NUT-29: Fix batch minting signature message separator

a1denvalu3:
## Summary
- Update `msg_to_sign` for NUT-29 batch minting to include proper colon separators between the quote ID and the blinded messages (e.g. `quote_id:B_0:B_1`).
- Clarify that the outputs are UTF-8 encoded hex strings.
- Update test vectors in `tests/29-tests.md` to reflect the new msg_to_sign and signature validation rule.

- [x] Cashu-TS
    - v4: https://github.com/cashubtc/cashu-ts/pull/674
    - v5: https://github.com/cashubtc/cashu-ts/pull/675
- [x] CDK - https://github.com/cashubtc/cdk/pull/2176
- [x]  Nutshell - https://github.com/cashubtc/nutshell/pull/1057

PR #379, opened 2026-05-26
NUT-20: harden mint quote signature (breaking) + NUT-04 UUIDv7 quote ids

robwoodgate:
## Summary

- **NUT-20** — replaces the `quote || B_0 || ... || B_(n-1)` signature message with a domain-separated, length-framed, amount-committing `msg_to_sign`. **Breaking change**: mints no longer accept the legacy message. The message does not commit the keyset `id` (so a wallet can re-target a rotated keyset, mirroring `SIG_ALL`).
- **NUT-04** — `quote` ids **MUST** be a UUIDv7.
- **NUT-29** — each locked quote is signed independently per NUT-20 over the consolidated `outputs`; mixed-method batches are rejected.
- **tests/20-test.md** — hardened-message test vector.

## Motivation

The legacy message doesn't commit output amounts and has ambiguous field boundaries across keyset curves. We chose a clean breaking change over a versioned migration because locked mint quotes have ~no production use and NUT-29 batch minting isn't fully released yet; wallets handle the transition as needed.

## Relationship to other work

Stacked on #375 (NUT-29 batch minting) — the diff includes its commits until it merges.


robwoodgate:
Re-targeted onto the NUT-29 branch (#375) so it can be merged directly into that PR: a1denvalu3/nuts#3. Closing this in favour of that.

PR #380, opened 2026-05-26
feat: preferred mints

d4rp4t:
Add ms (mint strict) field to NUT-18 Payment Requests

Adds a new optional ms boolean field to the Payment Request that controls whether the mint list m is treated as a strict requirement or a suggestion for the payer.

When ms is absent or true (default), the recipient MUST reject payments from mints not in the list. When ms is false, the mint list serves only as a hint to the payer.

This allows receivers to either enforce specific mints or merely express a preference while still accepting payments from any mint.

Implementations
- [x] cashu-ts: https://github.com/cashubtc/cashu-ts/pull/683
- [x] cdk: https://github.com/cashubtc/cdk/pull/2007
- [x] dotnut:  https://github.com/Kukks/DotNut/pull/37

### DIFF
--- a/18.md
+++ b/18.md
@@ -24,6 +24,7 @@ A Payment Request is defined as follows
   "u": str <optional>,
   "s": bool <optional>,
   "m": Array[str] <optional>,
+  "ms": bool <optional>,
   "d": str <optional>,
   "t": Array[Transport] <optional>,
   "nut10": NUT10Option <optional>,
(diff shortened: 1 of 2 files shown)


robwoodgate:
I have cherry-picked the commits from this PR into https://github.com/cashubtc/nuts/pull/381, which also adds an "extra fee" field for non preferred mint payments.

d4rp4t:
replaced by #381

PR #381, opened 2026-05-27
NUT-18: Payment request consolidation (and extra fees for non-preferred mints)

robwoodgate:
# Consolidates and Supersedes: #355, #380

- [x] Cashu-TS: https://github.com/cashubtc/cashu-ts/pull/683
- [x] CDK: https://github.com/cashubtc/cdk/pull/2007
- [x] Nutshell: https://github.com/cashubtc/nutshell/pull/1059
- [ ] Dotnut https://github.com/Kukks/DotNut/pull/37

## Summary

Consolidates the two open NUT-18 payment request proposals around mint preferences, strict mint lists, fee signalling, and required mint payment methods.

This PR keeps the existing `m` mint list as the single mint-list field, and adds small companion fields to clarify how that list should be interpreted and how the payer compensates the receiver for fees.

## Changes

- Retains `sm` (`supported_method`) from #355 to require that the payer's mint supports at least one listed payment method (e.g. `bolt11`, `bolt12`, `onchain`), now as a list of `{mn, mf?}` objects. The optional per-method fee `mf` compensates the receiver for melting out via that method:
  - applies only to payments from a mint outside `m`, or from any mint if `m` is not set; payments from a listed mint carry no per-method fee
  - the payer owes the lowest `mf` among the listed methods their mint supports
- Adds `mp` (`mint_preferred`) to signal whether the `m` mint list is strict or preferred (this is a flip from `ms` in #380):
  - absent / `false`: receiver only accepts proofs from listed mints
  - `true`: listed mints are preferred, but other mints may be accepted
- Specifies that the requested amount is net of input fees: the payer MUST select proofs such that `sum(proofs) - input_fee(proofs) >= amount owed`, with the input fee computed from the sending mint's `input_fee_ppk` (NUT-02). This protects the receiver from pathological proof sets (e.g. 1000 x 1 sat proofs at 250 ppk costs the receiver 250 sats to swap).
- Updates NUT-26 TLV tags for the new fields
- Adds test vectors covering the combined behaviour

Note: earlier revisions had a flat `fr` (`fee_required`) surcharge for non-preferred mints, then an `nf` (`net_fees`) opt-in flag. Review discussion concluded `fr` was redundant once `mf` is scoped to non-preferred mints, and `nf` had no sensible `false` case (no receiver wants to be underpaid by dust), so net-of-input-fees is now the unconditional rule. Both flags have been dropped and the `supported_method` sub-TLV uses tag 0x0a.

## Rationale

There were overlapping proposals for preferred mints and strict mint lists. Instead of introducing a separate preferred mint list, this keeps the request format minimal:

- `m` remains the mint list
- `mp` defines whether that list is strict or advisory
- `sm` lets receivers express the mint capabilities required for the payment flow, with per-method fees that compensate for melting out of a non-preferred mint
- the requested amount is net of input fees, so receivers get the amount they asked for regardless of the sender's proof selection

This avoids two competing mint-list fields while covering the desired behaviours.


a1denvalu3 (approved):

d4rp4t (approved):

18.md
```diff
@@ -36,11 +39,26 @@ Here, the fields are
...
+- `sm`: A list of supported methods (e.g. `"bolt11"`, `"bolt12"`, `"onchain"`) that the mint from which the payer sends ecash MUST support.
 - `d`: A human readable description that the sending wallet will display after scanning the request
 - `t`: The method of `Transport` chosen to transmit the payment (can be multiple, sorted by preference)
 - `nut10`: The required [NUT-10][10] locking condition
 
+## Mint list
+
+A Payment Request containing a mint list `m` describes the mints the receiver accepts payments from or prefers.
+
+If `m` is set and `ms` is not present, or is `true`, the mint list is strict: the sender MUST only send proofs from these mints, and the receiver MAY ignore payments sent from other mints.
```
thesimplekid:
```suggestion
If `m` is set and `ms` is not present, or is `true`, the mint list is strict: the sender MUST only send proofs from these mints, and the receiver SHOULD ignore payments sent from other mints.
```

I think its best wallet just ignore payments from mints they are not accepting so I think SHOULD is better, or course this can't be enforced and its up to the sender to send from accepted mints.

robwoodgate:
@callebtc , @thesimplekid - we have three implementations, can we consider final review / merge please.

cc @ye0man

KvngMikey:
4 implementations:
https://github.com/cashubtc/nutshell/pull/1059

KvngMikey (approved):

robwoodgate:
The only thing I still don't love about this is the `mint_strict` boolean being true if omitted.

It opens up all kinds of edge case issues as `undefined` is true but other falsey values are false.

I'd rather default to `false` on omitted, or if strict is the desirable default, flip the boolean semantic to `mint_prefer`, where true means the list of mints is a preference and false or omitted means it is strict

What are the thoughts on this?

thesimplekid:
> The only thing I still don't love about this is the `mint_strict` boolean being true if omitted.
> 
> It opens up all kinds of edge case issues as `undefined` is true but other falsey values are false.
> 
> I'd rather default to `false` on omitted, or if strict is the desirable default, flip the boolean semantic to `mint_prefer`, where true means the list of mints is a preference and false or omitted means it is strict
> 
> What are the thoughts on this?

I agree, I think default true is an anti pattern and we should avoid it.

robwoodgate:
> I agree, I think default true is an anti pattern and we should avoid it.

Agreed. Would you prefer to keep `mint_strict` and make the default false, or flip the boolean semantics to `mint_prefer` / `mint_suggest` or similar?

thesimplekid:
> > I agree, I think default true is an anti pattern and we should avoid it.
> 
> Agreed. Would you prefer to keep `mint_strict` and make the default false, or flip the boolean semantics to `mint_prefer` / `mint_suggest` or similar?

I would flip the semantics. `mint_preferred` / `mp`: absent or false means the existing strict `m` behavior, true means the `m` list is advisory/preferred.

robwoodgate:
> I would flip the semantics. `mint_preferred` / `mp`: absent or false means the existing strict `m` behavior, true means the `m` list is advisory/preferred.

yeah, that's my leaning too. @d4rp4t @a1denvalu3, do you agree?

KvngMikey:
> I would flip the semantics. `mint_preferred` / `mp`: absent or false means the existing strict `m` behavior, true means the `m` list is advisory/preferred.

i agree with this ...

robwoodgate:
`ms` (mint_strict) now replaced with `mp` (mint_preferred). 

This gives the same semantic, but flipped so an omitted value means `false` (strict).

robwoodgate:
Cashu-TS updated to use new `mp` semantic (replacing `ms`)

thesimplekid:
cdk updated https://github.com/cashubtc/cdk/pull/2007/commits/7009b0d8c5f45bcb3571b5e707406f90ad2b5967

KvngMikey:
Nutshell updated https://github.com/cashubtc/nutshell/pull/1059/changes/11612fe28436f9fd17f4d6cb3225009a5875c54c

thesimplekid (approved):

18.md
```diff
@@ -36,11 +39,26 @@ Here, the fields are
 - `a`: The amount of the requested payment
 - `u`: The unit of the requested payment (MUST be set if `a` is set)
 - `s`: Whether the payment request is for single use
-- `m`: A set of mints from which the payment is requested
+- `m`: The mint list from which payment will be accepted
+- `mp`: Whether the mint list is advisory (`true`) or strict (`false` or omitted)
+- `fr`: Additional fee reserve, in the request unit, that the payer MUST add when paying from a mint outside a non-strict mint list
```
callebtc: #### naming
small nits:  fee "reserve" implies we had some way of returning the overpaid fees. 

#### fee per method
together with the list of supported methods, it may be worth considering having a fee value per method

26.md
```diff
@@ -33,16 +33,19 @@ The payment request is encoded as a sequence of TLV fields. Each TLV entry consi
...
+| 0x04 | single_use        | u8        | Single-use flag: 0=false, 1=true (corresponds to `s` in JSON)                                     |
+| 0x05 | mint              | string    | Mint URL (repeatable for multiple mints, corresponds to `m` in JSON)                              |
+| 0x06 | description       | string    | Human-readable description (corresponds to `d` in JSON)                                           |
+| 0x07 | transport         | sub-TLV   | Transport configuration (repeatable, corresponds to `t` in JSON)                                  |
+| 0x08 | nut10             | sub-TLV   | NUT-10 spending conditions (corresponds to `nut10` in JSON)                                       |
+| 0x09 | mint_preferred    | u8        | Mint list strictness flag: 0=false, 1=true; defaults to 0 if absent (corresponds to `mp` in JSON) |
+| 0x0a | fee_reserve       | u64       | Additional fee reserve for payments from non-preferred mints (corresponds to `fr` in JSON)        |
+| 0x0b | supported_methods | string    | Supported payment method (repeatable for multiple methods, corresponds to `sm` in JSON)           |
```
callebtc:
```suggestion
| 0x0b | payment_method | string    | Supported payment method (repeatable for multiple methods, corresponds to `sm` in JSON)           |
```

26.md
```diff
@@ -33,16 +33,19 @@ The payment request is encoded as a sequence of TLV fields. Each TLV entry consi
...
+| 0x03 | unit              | u8/string | Currency unit (corresponds to `u` in JSON)                                                        |
+| 0x04 | single_use        | u8        | Single-use flag: 0=false, 1=true (corresponds to `s` in JSON)                                     |
+| 0x05 | mint              | string    | Mint URL (repeatable for multiple mints, corresponds to `m` in JSON)                              |
+| 0x06 | description       | string    | Human-readable description (corresponds to `d` in JSON)                                           |
+| 0x07 | transport         | sub-TLV   | Transport configuration (repeatable, corresponds to `t` in JSON)                                  |
+| 0x08 | nut10             | sub-TLV   | NUT-10 spending conditions (corresponds to `nut10` in JSON)                                       |
+| 0x09 | mint_preferred    | u8        | Mint list strictness flag: 0=false, 1=true; defaults to 0 if absent (corresponds to `mp` in JSON) |
+| 0x0a | fee_reserve       | u64       | Additional fee reserve for payments from non-preferred mints (corresponds to `fr` in JSON)        |
```
callebtc:
```suggestion
| 0x0a | method_fee       | u64       | Additional fee for each payment method (corresponds to `fr` in JSON)        |
```
maybe something like this

18.md
robwoodgate: I've changed the wording of `fr` to `fee_required`. No change to implementations required.

26.md
robwoodgate: I think you are mixing two semantics here. `fr` is the additional "fee required" if using a non-preferred mint.

A per-method fee would be a different thing to consider. For example, if there are three supported methods (bolt11, bolt12, onchain) how should we represent a fee for each method?

Edit: Idea here: https://github.com/cashubtc/nuts/pull/381#discussion_r3501713194

26.md
robwoodgate: The `supported_methods` label matches the `sm` key. It should be singluar (as repeatable in nut26), so I changed the label to `supported_method`. No changes to implementations required.

18.md
robwoodgate: We could support **fee per method** by redefining `sm` to an array of objects (like `transport`).

Something like:

```ts
"sm": [
  { "mn": "bolt11" },  // omitted mf = 0
  { "mn": "bolt12",  "mf": 0 },
  { "mn": "onchain", "mf": 50 }
]
```

method_fee (`mf`) omitted = 0. Fees are additive with top level `fr` (fee_required for non-preferred mint)

NUT-26 TLV:  0x0b goes from string to sub-TLV (mirrors the transport sub-TLV exactly):

```markdown
┌─────────┬─────────────┬────────┬─────────────────────────────────────────────────┐
│ Sub-Tag │ Field       │  Type  │                   Description                   │
├─────────┼─────────────┼────────┼─────────────────────────────────────────────────┤
│ 0x01    │ method name │ string │ e.g. "bolt11" (or u8 enum, like transport kind) │
├─────────┼─────────────┼────────┼─────────────────────────────────────────────────┤
│ 0x02    │ method fee  │ u64    │ optional; absent = 0                            │
└─────────┴─────────────┴────────┴─────────────────────────────────────────────────┘
```
What do you think? And do you want to bolt it into this PR or do that separately?

thesimplekid: There is probably also an argument to keeping individual payment requests simple and chaining them if you have more complex things you want to express creqA...:creqB for example. This would make sense if we expect these more complex fees to not be used much. 

However, in this case if its just this one thing we want to add and it mirrors what we already have for transport I think the above suggestion is good.

robwoodgate:
**fee per method** implemented as per https://github.com/cashubtc/nuts/pull/381#discussion_r3501713194

## Commit: https://github.com/cashubtc/nuts/pull/381/commits/46f6f6e29a4ebd72b46da54abcf8d696404753b6

- sm entries become {mn, mf?} objects; mf per-method fee stacks additively with top-level fr
- NUT-26 tag 0x0b promoted to sub-TLV (0x01 method, 0x02 fee)
- regenerate + verify test vectors; fix pre-existing basic vector blob/JSON mismatch

## Note: This is a change of spec, implementations will need to be updated.

## EDIT: Discussions ongoing. Will change again

cc @callebtc , @thesimplekid , @d4rp4t , @KvngMikey

robwoodgate:
CTS updated to use new fee per method:
- https://github.com/cashubtc/cashu-ts/pull/683/changes/7ecf99ddbb035fc5a4d3e9e5445d26cab02f6b60

18.md
```diff
@@ -36,11 +39,38 @@ Here, the fields are
...
+{
+  "mn": str,
+  "mf": int <optional> // omitted = 0
+}
+```
+
+- `mn`: the method name (e.g. `"bolt11"`, `"bolt12"`, `"onchain"`)
+- `mf`: additional fee, in the request unit, the payer MUST add when using this method
+
+The per-method fee (`mf`) is independent of the top-level fee required for non-preferred mints (`fr`): if the payer sends from a non-preferred mint (outside `m`, `mp` = true) **and** relies on an `sm` method that carries a fee, the payer MUST add both the `fr` and `mf` fees.
```
callebtc: what is the point of "fr" if there are "mf" fees as well?

robwoodgate: `fr` is the fee for using a non-preferred mint (mint not in the `m` list, `mp=true`)
`mf` is the fee for using a specific payment method.

This allows fee adds for unknown mints, as well as for using expensive methods.

Example:

```typescript
// preferred list (mp=true), fr=5, and onchain carries mf=10
pr.amountToSend('https://in-list.mint'); // Amount, base only    → 100
pr.amountToSend('https://in-list.mint', 'onchain'); // + mf      → 110
pr.amountToSend('https://other.mint', 'bolt11'); // + fr         → 105
pr.amountToSend('https://other.mint', 'onchain'); // + fr + mf   → 115
```

thesimplekid: I think this is redundant? As presumably the fee for using a non default mint is because they are going to melt out and that fee will be dependent on method so can just include it in the mf? And mf should only apply to mints outside of preferred?

robwoodgate: The thing I'm confused about with the whole per-method fee idea is how the sender knows which method the receiver will use to cash out?

callebtc: > I think this is redundant?

I think so too. 

If the mint is in the list, you should accept the ecash (no method fee applies). If it's a foreign mint, you required a method and a certain fee buffer for that method.

robwoodgate: So what shape would you want to see? If the sender cannot know which method the receiver will use, it seems that **fee per method** itself seems redundant.

`sm` can just be a single method string (your mint must support x) and `fr` the fee as now.

```json
"sm": "bolt11",
"fr": "2",
```

Otherwise, given a menu, the sender is likely to just choose the "cheapest" because they cannot know whether the receiver will offload via bolt11, onchain etc

Alternatively, we can go back to the original multi-list `sm` (your mint must support at least one of these methods) and a single `fr` for using the non-preferred mint.

```json
"sm": [ "bolt11", "onchain" ],
"fr": "2",
```

robwoodgate: Some thoughts on this, and the wider payment request needs / issues we've discussed over time:

### Fee per method

I suppose a sender is rightfully able to send cheapest per method fee if their chosen mint support 2 or more methods.

So if payment request says: 
```
"sm": [
  { "mn": "bolt11" },  // omitted mf = 0
  { "mn": "onchain", "mf": 50 }
]
```

And the sender mint supports both, they add no extra `mf` fee because their mint was compliant with request.

But if a sender's mint is `onchain` only, they have to pay the lowest per method fee their mint supports (50 in example).

Is that how you envisaged it @callebtc ?

And should the `mf` per mint fee apply in all cases, or just when using a non-preferred mint?

### Fee Required (`fr`)

I'm not sure the `fr` is totally redundant. We discussed a case where a sender can use a mint they control with high lightning fees. The `fr` was an extra fee for using a mint the sender doesn't explicitly trust, regardless of the methods they support.

If the `mf` fee applies only to non-preferred mint, then I agree, `fr` is redundant.

### Include input fees

Nothing we have discussed so far handles the case where a sender sends a pathological set of proofs - example, 1000 x 1 sat proofs from a mint with 250ppk fee = 250 sats. Which is kind of what I imagined the `fr` fee would help to mitigate. "If you choose to send from your own mint, add this fee".

Maybe the input fee issue could disappear along with `fr` by replacing it with an "include fees" boolean, or mandating the `amount` is net of input fees?

robwoodgate:
Following discussion in the TS Devs meeting, I've updated the spec as follows:

## Commit: https://github.com/cashubtc/nuts/pull/381/commits/7ff78640bac280488223a89c99368ca0f86fa8a8

- `fr` (fee required) dropped
- `mf` (method fee) semantics clarified to ONLY apply for non-preferred mints (or when `m` is unset)
- `amount` specified as being net of per-proof input fees

## Note: Change of spec, implementations will need to be updated if this new direction is agreeable

Related discussion: https://github.com/cashubtc/nuts/pull/381#discussion_r3528893628

cc @callebtc , @thesimplekid , @d4rp4t , @KvngMikey

18.md
```diff
@@ -33,14 +35,52 @@ A Payment Request is defined as follows
...
+
+If `m` is set and `mp` is not present, or is `false`, the mint list is strict: the sender MUST only send proofs from these mints, and the receiver SHOULD ignore payments sent from other mints.
+
+If `m` is set and `mp` is `true`, the mint list is preferred: the receiver accepts payments from other mints, but the payer SHOULD use a mint in `m` when possible.
+
+If `m` is not set, `mp` SHOULD be ignored.
+
+## Supported payment methods
+
+If `sm` is set, the payer MUST send ecash from a mint that supports at least one of the listed payment methods. Each `SupportedMethod` is defined as:
```
thesimplekid: nit: Should we specify supports melting? As it's possible for a mint to support different methods for minting and melting.

thesimplekid (approved):
Updated cdk to current spec in https://github.com/cashubtc/cdk/pull/2007/commits/b5a5eb45590df4b214ac42d25b111d4d51e5146e

18.md
robwoodgate: good point!

robwoodgate: changed in 51bb42bb5b47c1a7551cb30f1f7aa6cc1ea764b9

KvngMikey (approved):
Nutshell updated to current spec https://github.com/cashubtc/nutshell/commit/91195fdec95b35c4fcbf494181298711e92b6575

a1denvalu3 (approved):

a1denvalu3:
Updated CDK PR [here](https://github.com/cashubtc/cdk/pull/1925)

callebtc (approved):

Issue #378, opened 2026-05-24
NUT-29: add method field for mint and melt quotes, web socket notifications


ye0man:

thesimplekid:
I think this issue is not Nut29, but is related to the to the the NUT04/NUT05 request and response structs not having the method within the struct. This means that the type can only be know by the the shape or the the endpoint that is being used. The shape fails to be a good way to differentiate in the case of the shape being the same for multiple methods, for example onchain, bolt12 and if we add custom methods the shapes are all the same. 

This is specifically an issue for websocket notifications where there is not even the endpoint, so the only option is the shape, so since this is not an option wallets must track what methods map to what subscription ids.

So the suggestion is to add the method name to the struct of the Mint/Melt Quote request and response types.

robwoodgate:
Ok, I understand the issue in context of WS notifications. Though the wallet would have requested a quote ID using a specific method, it then needs to track that method <--> ID mapping so it can figure out how to complete an operation  if the update comes via WS.

thesimplekid:
> Though the wallet would have requested a quote ID using a specific method, it then needs to track that method <--> ID mapping so it can figure out how to complete an operation if the update comes via WS.

Yes it is possible to track this, and is what cdk does now. However, I think it is better to make it more explicit what method the struct is for within the struct itself.

robwoodgate:
These two companion PRs should allow mints and wallets to handle existing and custom method quotes via websockets in a sane way:

- https://github.com/cashubtc/nuts/pull/372
- https://github.com/cashubtc/nuts/pull/387

PR #382, opened 2026-05-27
NUT-04/05: custom payment methods

asmogo:
# Implementations

- Cashu-TS: https://github.com/cashubtc/cashu-ts/pull/698
- CDK: https://github.com/cashubtc/cdk/pull/2146
- Nutshell: TBC

support custom payment methods not defined in a dedicated NUT.


04.md
```diff
@@ -101,6 +101,35 @@ The mint then responds with:
...
+
+### Method-Specific Fields
+
+Custom payment methods **MAY** include additional fields in requests and responses. Implementations **MUST** ignore unrecognized fields to preserve forward compatibility. When a custom method gains widespread adoption, its fields **MAY** be formalized in a dedicated NUT.
+
+Custom payment methods **MAY** include extra fields that the mint forwards to the payment processor without validation.
+
+### Websocket Notifications
+
+Custom methods **MAY** support websocket subscriptions via [NUT-17][17] with kind `"{method}_mint_quote"`.
```
robwoodgate: Related: https://github.com/cashubtc/nuts/pull/372

05.md
```diff
@@ -203,6 +203,24 @@ GET https://mint.host:3338/v1/melt/quote/bolt11/TRmjduhIsPxd...
...
+
+### Melt Quote
+
+For a custom `{method}`, the wallet sends a request following the common melt quote request format (see [General Flow](#general-flow)). The `request` field is the method-specific payment target (e.g., a bank account identifier, an on-chain address, a payment processor reference). `unit` is the unit the wallet would like to pay with.
+
+The mint responds with the common melt quote response format. Method-specific fields are defined by the method-specific NUT.
+
+### Websocket Notifications
+
+Custom methods **MAY** support websocket subscriptions via [NUT-17][17] with kind `"{method}_melt_quote"`.
```
robwoodgate: Related: https://github.com/cashubtc/nuts/pull/372

04.md
robwoodgate, line 48: To better support custom, I think we should expand the base request something like:

```suggestion
{
  "unit": <str_enum[UNIT]>,
  "amount": <int>, // Optional
  "description": <str>, // Optional
  "pubkey": <str> // Optional (NUT-20)
}
```

Then NUT specific methods can enforce / ignore as needed - eg: NUT-23 makes `amount` required., NUT-25 makes `pubkey` required.

robwoodgate, line 67: Custom would be better supported if the base reply is more like:

```suggestion
{
  "quote": <str>,
  "request": <str>,
  "unit":  <str_enum[UNIT]>,
  "expiry": <int|null>,
  "pubkey": <str>, // Optional (NUT-20)
  "amount_paid": <int>,
  "amount_issued": <int>
  // Additional method-specific fields will be included
}
```

05.md
robwoodgate, line 52: Melt base suggestion for better custom handling:

```suggestion
{
"request": <str>,
"unit": <str_enum[UNIT]>,
"amount": <int>, // Optional
// Additional method-specific fields will be required
}
```

robwoodgate: Melt response base suggestion for better custom method handling:

```suggestion
{
"quote": <str>,
"request": <str>,
"amount": <int>,
"unit": <str_enum[UNIT]>,
"fee_reserve": <int>, // Optional
"state": <str_enum[STATE]>,
"expiry": <int>
// Additional method-specific fields will be included
}

Where `quote` is the quote ID, `amount` and `unit` the amount and unit that need to be provided, and `expiry` is the Unix timestamp until which the melt quote is valid.

`request` contains the method-specific payment routing instructions (eg bolt11 Lightning invoice, Bitcoin onchain address, bank account details, paypal email etc).

`fee_reserve` is the additional fee for using the method. The mint expects the wallet to include `Proofs` of _at least_ `total_amount = amount + fee_reserve + fee` where `fee` is calculated from the keyset's `input_fee_ppk` as described in [NUT-02][02].
```

robwoodgate (commented):
I think this is a fantastic step forward for wallets to be able to better support unknown methods - I've added some suggested tweaks to the base structs to make life even better.

robwoodgate:
Sorry @asmogo - I questioned an earlier draft that included `method` in the quote structs (thinking mint and wallet know by virtue of the endpoint used to request it), but [this related issue](https://github.com/cashubtc/nuts/issues/378) highlights that for WS, and especially with the suggestion on harmonizing the base structs for custom methods, the distinction is easy to blur/lose.

So maybe we DO need to include `method: <str>,` in the quote structs for both mint and melt.

robwoodgate:
Related https://github.com/cashubtc/nuts/pull/377

which adds the `amount_oaid`, `amount_issued`, `updated_at` parms to default quote structs

04.md
```diff
@@ -101,6 +101,35 @@ The mint then responds with:
...
 
+## Custom Payment Methods
+
+Payment methods not specified in a dedicated NUT can be supported as custom payment methods. A custom payment method is identified by a lowercase string `{method}` (e.g., `paypal`, `stripe`, `onchain-payment-processor`). The `{method}` string **MUST** contain only ASCII alphanumeric characters, hyphens (`-`), and underscores (`_`), and **MUST** be non-empty.
+
+### Mint Quote
+
+For a custom `{method}`, the wallet sends a request following the common mint quote request format (see [General Flow](#general-flow)). Method-specific fields (e.g., an `amount` of tokens to mint, a `description`, or a `pubkey` for [NUT-20][20] locks) are defined by the method-specific NUT.
+
+The mint responds with the common mint quote response format. The `request` field contains the method-specific payment request (e.g., a payment URL, an on-chain address, an account identifier). Additional method-specific fields (such as `amount` or `expiry`) are defined by the method-specific NUT.
```
robwoodgate: Wallets can derive the accounting fields from state for methods they know, but for an unknown method there's no fallback, so these need to be explicit. Suggest:

```suggestion
The mint responds with the common mint quote response format, and **MUST** include the `amount_paid`, `amount_issued` and `updated_at` fields. The `request` field contains the method-specific payment request (e.g., a payment URL, an on-chain address, an account identifier). Additional method-specific fields (such as `amount` or `expiry`) are defined by the method-specific NUT.
```

05.md
```diff
@@ -203,6 +203,24 @@ GET https://mint.host:3338/v1/melt/quote/bolt11/TRmjduhIsPxd...
...
 
+## Custom Payment Methods
+
+Payment methods not specified in a dedicated NUT can be supported as custom payment methods. See [NUT-04][04] for the custom method identifier definition and general conventions.
+
+### Melt Quote
+
+For a custom `{method}`, the wallet sends a request following the common melt quote request format (see [General Flow](#general-flow)). The `request` field is the method-specific payment target (e.g., a bank account identifier, an on-chain address, a payment processor reference). `unit` is the unit the wallet would like to pay with.
+
+The mint responds with the common melt quote response format. Method-specific fields are defined by the method-specific NUT.
```
robwoodgate: Wallets can't interpret a custom state machine for a method they don't understand, so custom melts need the standard states. Suggest:

```suggestion
The mint responds with the common melt quote response format, using the standard `state` values (`"UNPAID"`, `"PENDING"`, `"PAID"`). Method-specific fields are defined by the method-specific NUT.
```

robwoodgate (commented):
Have implemented custom method support in cashu-ts on top of this + #377 + #387  (cashubtc/cashu-ts#698) and it works nicely. Wallets can quote, mint and melt methods they don't fully understand using just the base structs.

A few tweaks below that came out of implementing. Note the suggestions assume #377 merges first.

Related to my base struct suggestions above: `expiry` really wants to be in the common mint quote response - every method has one, and a wallet showing a quote for an unknown method needs to know when to stop polling. With #377 in, it's the only field from the suggested base response still left as method-specific.

(method in the quote structs is now covered by #387, which also resolves the WS routing question from #378.)

Can also confirm the optional `fee_reserve` in the melt base works in practice - onchain can't make it required (it uses `fee_options`), and bolt11/12 tighten it to required in their own NUTs.

robwoodgate:
Added https://github.com/asmogo/nuts/pull/1 with my suggestions ready to merge into this PR if acceptable.

robwoodgate (approved):

04.md
```diff
@@ -64,6 +71,7 @@ Where:
 - `quote` is the quote ID in UUID v7 format
 - `request` is the payment request for the quote
 - `unit` corresponds to the value provided in the request
+- `expiry` is the Unix timestamp until which the quote is valid (`null` if it does not expire).
```
robwoodgate: Consistency with other bullets

```suggestion
- `expiry` is the Unix timestamp until which the quote is valid (`null` if it does not expire)
```

robwoodgate (approved):
LGTM

robwoodgate:
@callebtc - I've fixed the merge conflicts in this one.

04.md
```diff
@@ -40,18 +40,25 @@ Depending on the payment method, the request structure may vary, but all methods
 
 ```json
 {
-  "unit": <str_enum[UNIT]>
+  "unit": <str_enum[UNIT]>,
+  "amount": <int>,        // Optional
+  "description": <str>,   // Optional
```
thesimplekid, line 45: I don't think that description should be in this base on for example for onchain it does not make sense. That base should be minimal that all have. We already say that custom can extend this so adding description to custom is valid if we remove it here. 

Same applies to amount.

04.md
```diff
@@ -119,6 +127,34 @@ The mint then responds with:
...
+Custom payment methods **MAY** include extra fields that the mint forwards to the payment processor without validation.
+
+### Websocket Notifications
+
+Custom methods **MAY** support websocket subscriptions via [NUT-17][17] with kind `"{method}_mint_quote"`.
+
+### Cached Responses
+
+Custom method endpoints **MAY** support caching via [NUT-19][19] (e.g., `POST /v1/mint/{method}`).
+
```
thesimplekid: Nut19 is already generic to method so I don't think we need to call it out here.

04.md
```diff
@@ -119,6 +127,34 @@ The mint then responds with:
...
+### Method-Specific Fields
+
+Custom payment methods **MAY** include additional fields in requests and responses. Implementations **MUST** ignore unrecognized fields to preserve forward compatibility. When a custom method gains widespread adoption, its fields **MAY** be formalized in a dedicated NUT.
+
+Custom payment methods **MAY** include extra fields that the mint forwards to the payment processor without validation.
+
+### Websocket Notifications
+
+Custom methods **MAY** support websocket subscriptions via [NUT-17][17] with kind `"{method}_mint_quote"`.
```
thesimplekid: This references an optional nut in a mandatory one. Does it make more sense to remove this from here and add something to nut17 on the naming if that is unclear?

04.md
```diff
@@ -119,6 +127,34 @@ The mint then responds with:
...
+
+### Cached Responses
+
+Custom method endpoints **MAY** support caching via [NUT-19][19] (e.g., `POST /v1/mint/{method}`).
+
+### Amount Precision
+
+If an `amount` is provided, fiat-denominated units (e.g., `usd`, `eur`) **SHOULD** represent the value in the smallest currency unit (e.g., cents). For example, `"amount": 100` with `"unit": "usd"` represents $1.00 USD.
+
```
thesimplekid: Isn't this covered here https://github.com/cashubtc/nuts/blob/main/01.md#supported-currency-units so we can remove it from nut04?

04.md
```diff
@@ -119,6 +127,34 @@ The mint then responds with:
...
+For a custom `{method}`, the wallet sends a request following the common mint quote request format (see [General Flow](#general-flow)). Method-specific fields (e.g., an `amount` of tokens to mint, a `description`, or a `pubkey` for [NUT-20][20] locks) are defined by the method-specific NUT.
+
+The mint responds with the common mint quote response format and **MUST** include the `amount_paid`, `amount_issued` and `updated_at` accounting fields. The `request` field contains the method-specific payment request (e.g., a payment URL, an on-chain address, an account identifier). Additional method-specific fields are defined by the method-specific NUT.
+
+### Method-Specific Fields
+
+Custom payment methods **MAY** include additional fields in requests and responses. Implementations **MUST** ignore unrecognized fields to preserve forward compatibility. When a custom method gains widespread adoption, its fields **MAY** be formalized in a dedicated NUT.
+
+Custom payment methods **MAY** include extra fields that the mint forwards to the payment processor without validation.
```
thesimplekid: Worth noting that I think payment processor is something we use in cdk but I don't think it is referenced or defined else where in the spec.

05.md
```diff
@@ -205,6 +217,24 @@ GET https://mint.host:3338/v1/melt/quote/bolt11/019e6d5a-2347-7000-89e2-35fe79f9
...
+The mint responds with the common melt quote response format, using the standard `state` values (`"UNPAID"`, `"PENDING"`, `"PAID"`). Method-specific fields are defined by the method-specific NUT.
+
+### Websocket Notifications
+
+Custom methods **MAY** support websocket subscriptions via [NUT-17][17] with kind `"{method}_melt_quote"`.
+
+### Cached Responses
+
+Custom method endpoints **MAY** support caching via [NUT-19][19] (e.g., `POST /v1/melt/{method}`).
+
```
thesimplekid: Same comments as the related sections in nut04.

thesimplekid (changes requested):

04.md
robwoodgate: I think the idea here is to signal that wallets may look for an amount or description if provided and that custom methods should not (for example) use 'desc', 'amt' etc if they wish to communicate those ideas

04.md
robwoodgate: Maybe just replace "the" with "a" eg: 'a third party payment processor'

04.md
asmogo: yes. NUT-01 is more strict. I have updated this in https://github.com/cashubtc/nuts/pull/382/commits/7551066d21b813768e08be34305ba32f9569e8e7

04.md
asmogo: I have updated this in https://github.com/cashubtc/nuts/pull/382/commits/7551066d21b813768e08be34305ba32f9569e8e7

05.md
asmogo: I have updated this in https://github.com/cashubtc/nuts/pull/382/commits/7551066d21b813768e08be34305ba32f9569e8e7

thesimplekid (approved):

robwoodgate (approved):

a1denvalu3 (approved):

callebtc (approved):

PR #388, opened 2026-06-11
NUT-XX: Proof of Liabilities (PoL)

a1denvalu3:
## Summary

This draft proposes an epoch-based Proof of Liabilities protocol for Cashu mints. It lets wallets and auditors verify append-only issued and spent liability histories, authenticate individual liability events with transactional receipts, and detect signed or committed contradictions.

The proposal builds on the original [ecash PoL design](https://gist.github.com/callebtc/ed5228d1d8cbaade0104db5d1cf63939) by @callebtc and the subsequent [Merkle-sum tree, receipt, and OpenTimestamps work](https://gist.github.com/a1denvalu3/9cb8437714e0f5c0e9517611e811ca08).

## Base PoL protocol

The base specification now defines:

- synchronized append-only sum-MMRs for issued blinded messages and spent proof secrets
- deterministic right-to-left peak bagging, inclusion proofs, and epoch-transition consistency proofs
- per-keyset issued, spent, and outstanding-balance commitments
- a unit-aware Merkle commitment over every unexpired keyset in an epoch
- NUT-06 BIP-340 signatures over canonical epoch manifests
- OpenTimestamps anchoring of the global epoch digest
- signed transactional PoL receipts for issued outputs and spent inputs
- mandatory, immutable keyset `unit` and `deactivation_epoch` declarations
- fraud evidence for receipt omission, manifest equivocation, append-only violations, and keyset lifecycle violations
- executable test vectors for sum-MMR operations, Merkle commitments, manifests, receipts, consistency proofs, and lifecycle violations

## Experimental Bonded PoL extension

The branch also includes a separate experimental extension that explores trustless Bitcoin-native slashing under prospective covenant functionality. A mint locks bitcoin in a recursive covenant carrying the authenticated PoL state; the first confirmed challenger proving a supported violation can receive the complete bond.

The extension specifies:

- `ACTIVE`, `PENDING`, `CHALLENGED`, `WITHDRAWAL_DELAY`, and `UNINITIALIZED` covenant states
- consensus-validated initialization, epoch publication, finalization, closing, withdrawal, challenge, refutation, and slashing transitions
- one immutable Cashu unit per bond while authenticating the complete all-unit PoL epoch
- same-unit bonded keyset filtering and exact outstanding-liability aggregation
- immutable keyset fee, expiry, lifecycle, receipt-scheme, and denomination-key commitments
- Merkle-authenticated BIP-340 receipt keys, excluding unsupported BLS receipt schemes in version 1
- concrete keyset, epoch, historical-MMR, amount-key, and liability opening proofs
- mandatory issued/spent MMR consistency verification at every bonded epoch transition
- epoch-based redemption expiry and a full withdrawal challenge delay
- strict successor-state mutation allowlists and complete bond-value preservation
- canonical bounded witness schemas and compiler-neutral script pseudocode with every helper defined
- a deployment resource gate binding opcode semantics, compiled verifier templates, tree layout, and resource limits into `program_hash`

## Bitcoin assumptions

Bonded PoL is not deployable on Bitcoin today. It is explicitly experimental and assumes semantics equivalent to:

- BIP-443 `OP_CHECKCONTRACTVERIFY` for recursive state and value-preserving covenant outputs
- BIP-348 `OP_CHECKSIGFROMSTACK` for BIP-340 manifest and receipt verification
- BIP-347 `OP_CAT` for bounded canonical message construction
- unsigned 64-bit arithmetic, division/modulo, bit operations, and limited transaction introspection
- bounded scripts and proofs that fit the resource limits of whatever consensus profile is ultimately activated

No final Tapscript bytecode or deployability claim is made. The document defines the required logical semantics and requires a future compiler to demonstrate that a chosen parameter set fits activated consensus limits.

## Draft identifier

The documents currently use `NUT-388` and `NUT-388-B` because 388 is this pull request number. These identifiers must be replaced with the actual NUT number assigned if the proposal is merged.

## Additional material

The branch contains:

- `pol.md` — base Proof of Liabilities proposal
- `bonded-pol.md` — experimental bonded and slashable extension
- `bonded-pol-scripts.md` — concrete covenant pseudocode and witness definitions
- `tests/pol-tests.md` — cryptographic and lifecycle test vectors
- `presentation/` — a separate visual explanation of the proposal


pol.md
```diff
@@ -0,0 +1,291 @@
...
+- **Verification:** Wallets verify the BIP340 Schnorr signature on the SHA-256 hash of the formatted message against the keyset public key corresponding to the output's amount (`keyset.public_keys[amount]`).
+
+#### B. Spent Inputs (Ecash Proofs)
+For each spent proof in a swap or melt, the mint returns a list of `spent_receipts` in the response.
+- **Message format to sign:**
+  `{Y_hex}:{target_epoch}`
+  where `{Y_hex}` is the hex-encoded public key string of the spent proof's curve point `Y = hash_to_curve(secret)`, and `{target_epoch}` is the target upcoming epoch index string.
+- **Signing key:** The keyset private key corresponding to the spent input's specific amount (`keyset.private_keys[amount]`).
+- **Verification:** Wallets verify the BIP340 Schnorr signature on the SHA-256 hash of the formatted message against the keyset public key corresponding to the spent input's amount (`keyset.public_keys[amount]`).
+
```
robwoodgate: The per-amount public keys in BLS12-381 (02) keysets (#371) are in G2, so a BIP340 receipt can't verify against public_keys[amount] as written.

To preserve the per-amount binding, perhaps the receipt signature could track the keyset's curve (BLS signature for 02)?

Otherwise, binding a receipt to its amount would need a separate secp key per amount for BLS keysets.

```suggestion
### 2. Message Formats and Keys

Each receipt is signed under the keyset's per-amount key (`private_keys[amount]`), binding it to the note's denomination. The signature scheme follows the keyset's curve (NUT-02 version byte):

| Version | Curve | Signature |
|-----------|-----------|------------------------------------------------------------|
| `00`/`01` | secp256k1 | BIP340 Schnorr over `SHA-256(message)` |
| `02` | BLS12-381 | `σ = a · H_G1(message)`, verified `e(σ, G2) == e(H_G1(message), K)` |

**Message to sign:** the UTF-8 bytes of the point hex, a literal `:`, and the target epoch index as a decimal string.
- **Output:** `<B'_hex>:<target_epoch>`, e.g. `02b1a...d4f:12`
- **Spent input:** `<Y_hex>:<target_epoch>`, e.g. `02e9c...a07:12`, where `Y = hash_to_curve(secret)`.

**secp256k1 (`00`/`01`):** verify the BIP340 signature against `public_keys[amount]`.

**BLS12-381 (`02`):** `H_G1` is `hash_to_curve_G1` (RFC 9380) under DST `"Cashu_PoL_Receipt_BLS12381G1_XMD:SHA-256_SSWU_RO_"`, which MUST differ from the BDHKE secret DST. `σ` is the compressed G1 encoding; `K = public_keys[amount]` (G2). Receipts MAY be batch-verified per NUT-00.

```

robwoodgate:
Will there be a section on "proof of assets" in this spec?

At the moment it solidly covers proof of liabilities, but there is no guarantee the mint has the assets to back the outstanding liabilities. If just liabilities, we should probably tone down the "solvency" side of the introduction.

pol.md
```diff
@@ -0,0 +1,291 @@
...
+- `hash_P` is the SHA256 hash of the concatenated bytes:
+  `hash_L + hash_R + bytes_8(sum_L) + bytes_8(sum_R)`
+
+---
+
+## Epoch Manifest Creation and On-Chain Commitments
+
+To prevent the mint from rewriting historic trees or showing different root values to different users, roots are committed to epochs synchronized across the entire mint.
+
+Every epoch interval (e.g., 24 hours), the mint builds the Issued and Spent trees for all unexpired keysets, publishes on-chain commitments, and signs the resulting manifests:
```
robwoodgate: What would be the plan for expired keysets? These can still have liabilities attached.

pol.md
```diff
@@ -0,0 +1,291 @@
...
+1. **Upgrade Proof (Optional):** Submit the binary `ots_receipt` to a public calendar server upgrade endpoint (e.g., `https://alice.btc.calendar.opentimestamps.org/upgrade`). If a mined block contains the commitment, the calendar returns an upgraded receipt containing the on-chain Merkle path.
+2. **Scan for Block Attestation:** Scan the binary upgraded receipt for the Bitcoin Block Header attestation tag (`0x00` followed by `0x05` for `A_BLOCKHEADER`).
+3. **Parse Block Height:** Parse the Bitcoin block height immediately following the tag (serialized as a standard variable-length VarInt).
+4. **Verify Block Confirmation:** Query any independent public Bitcoin block explorer API (e.g. `https://mempool.space/api/blocks/tip/height`) to check that the parsed block height exists, matches the manifest timestamp, and has sufficient confirmations. If present, the attestation is fully confirmed. If the pending tag `0x00 0x06` is found instead, the receipt has been submitted to the calendar but is still awaiting blockchain anchoring.
+
+### Step 3: Validate Issued Path Walks
+For each active token held:
+1. Reconstruct `B'` (if only holding `C` and `r`, compute `B' = Y + rG`).
```
robwoodgate: Would need a BLS counterpart adding too (`B_ = r·Y`)

pol.md
```diff
@@ -0,0 +1,291 @@
...
+
+### Step 5: Verify Solvency Equation
+Check that the outstanding balance reported matches:
+`outstanding_balance == root_issued_sum - root_spent_sum`
+
+---
+
+## Cryptographic Fraud Challenge (JSON format)
+
+If any verification path walk fails (e.g., the leaf is not present or has a sum value different from the token's face value, or the computed root doesn't match the signed root), the wallet generates a **Fraud Challenge**. This is an aggregated, self-contained JSON document that can be shared publicly to mathematically prove the mint committed perjury:
```
robwoodgate: An honest mint might only get the note into an epoch after the target, for example a swap near the cutoff settles into N+1 instead of N. Since the trees are cumulative, checking the latest published epoch (≥ target_epoch) still finds it, whereas pinning the check to exactly `target_epoch` could falsely flag the mint. A note that never appears is still caught.

```suggestion
If any verification path walk fails in the latest published epoch `≥ target_epoch` (e.g., the leaf is not present or has a sum value different from the token's face value, or the computed root doesn't match the signed root), the wallet generates a **Fraud Challenge**. This is an aggregated, self-contained JSON document that can be shared publicly to mathematically prove the mint committed perjury:
```

pol.md
a1denvalu3: Once the keyset expires, the mint simply stops publishing manifests including the roots for that keyset. The fact that they can still have liabilities is inconsequential IMO

pol.md
a1denvalu3: > whereas pinning the check to exactly target_epoch could falsely flag the mint

It's the mint that decides which epoch to commit them in, and lets the client know with a `pol_receipt`. So -in other words- it's the Mint's responsibility to make sure the entries appear by the specified epoch in the `pol_receipt`.

a1denvalu3:
> Will there be a section on "proof of assets" in this spec?

The scope of this spec is proving liabilities. Proving the reserves might change depending on which backend(s) the Mint decides to use. For example on with Ark it's very easy to prove the reserves, by just spending its VTXOs to themselves.

pol.md
```diff
@@ -0,0 +1,275 @@
...
+- **Issued Leaf Index ($I_{\text{issued}}$):**
+  - Index: $\text{SHA256}(\text{bytes}(B'))$ (using the 33-byte compressed public key representation of $B'$) parsed as a big-endian integer.
+  - Value: Face amount of the signed promise.
+- **Spent Leaf Index ($I_{\text{spent}}$):**
+  - Index: $\text{SHA256}(\text{bytes}(Y))$ (using the 33-byte compressed public key representation of $Y$) parsed as a big-endian integer.
+  - Value: Face amount of the spent proof.
+
+### 2. Node Structure & Hashing
+
+Each node is represented as $(hash, sum\_value)$ where $hash$ is 32 bytes and $sum\_value$ is an 8-byte big-endian integer.
```
Egge21M: We should be explicit about overflow and duplication behavior. e.g.

```md
If `sum_L + sum_R >= 2^64`, tree construction MUST fail.

A mint MUST NOT insert the same issued item `B'` more than once in the same issued tree, and MUST NOT insert the same spent item `Y` more than once in the same spent tree. Duplicate insertion MUST be treated as an implementation error.
```

pol.md
```diff
@@ -0,0 +1,275 @@
...
+   `"{keyset_id}:{epoch_index}:{timestamp}:{root_issued_hash}:{root_issued_sum}:{root_spent_hash}:{root_spent_sum}:{outstanding_balance}:{ots_receipt}"`
+   where `ots_receipt` is the hex-encoded OTS receipt.
+6. **Signing:** Sign the message with a BIP-340 Schnorr signature using the mint's master NUT-06 private key signing the SHA256 digest of this serialized manifest string.
+7. **Publish:** Store and publish the signed manifest, signatures, and OTS receipts.
+
+---
+
+## Compact Bitmasked Sibling Proofs
+
+To minimize proof size, default empty sibling nodes are omitted. Instead, the mint returns a 256-bit bitmask where the $d$-th bit indicates if the sibling at level $d$ is non-empty ($1$) or empty ($0$, to be replaced by the precomputed default empty node).
```
Egge21M: For clarity

```suggestion
To minimize proof size, default empty sibling nodes are omitted. The mint returns a 256-bit `compact_mask`, serialized as a 32-byte big-endian integer and displayed as 64 lowercase hex characters.

Bit `d` of the integer, where `d = 0` is the leaf level and `d = 255` is the sibling below the root, indicates whether the sibling at level `d` is non-empty:

- `1`: the sibling is present in `siblings`.
- `0`: the verifier MUST use the precomputed default empty node for level `d`.
```

pol.md
```diff
@@ -0,0 +1,275 @@
...
+  "target_epoch": 12,
+  "signature": "<hex_encoded_signature>"
+}
+```
+
+### 2. Message Formats and Cryptography
+
+Each receipt is signed under the keyset's per-amount private key (`private_keys[amount]`) corresponding to the note's denomination.
+
+- **Message to Sign:** `<point_hex>:<target_epoch_decimal_string>`
```
Egge21M: Should this commit to more that just these two?

pol.md
```diff
@@ -0,0 +1,275 @@
...
+
+### Step 3: Validate Issued Path Walks
+
+For each held active token:
+
+1. Reconstruct $B'$ (BDHKE: $B' = Y + rG$; BLS: $B' = r \cdot Y$).
+2. Leaf index $I = \text{SHA256}(B')$ parsed as a big-endian integer.
+3. Walk up the 256 levels:
+   - For $d \in [0, 255]$:
+     - The $d$-th bit of the bitmask determines child position.
```
Egge21M: This is inconsistent with the earlier paragraph where the $d$-th bit of the bitmask determines supplied vs. default node, not child position.

pol.md
```diff
@@ -0,0 +1,275 @@
...
+
+- **Query Params:** `epoch_index` (optional, integer)
+- **Response:**
+
+```json
+{
+  "keyset_id": "009a6154b71113b7",
+  "epoch_index": 1,
+  "timestamp": "2026-06-11T12:00:00Z",
+  "signing_pubkey": "020...",
```
Egge21M: Looks like a compressed pubkey, not a 32-byte BIP340

pol.md
```diff
@@ -0,0 +1,275 @@
...
+| :-------- | :-------- | :----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `00`/`01` | secp256k1 | BIP340 Schnorr over $\text{SHA256}(\text{message})$. Verified against `public_keys[amount]`.                                                                                                                                                                                   |
+| `02`      | BLS12-381 | $\sigma = a \cdot H_{G1}(\text{message})$ (compressed G1). Verified via $e(\sigma, G2) == e(H_{G1}(\text{message}), K)$ where $K = \text{public\_keys}[amount]$ (G2). $H_{G1}$ is RFC 9380 `hash_to_curve_G1` under DST `"Cashu_PoL_Receipt_BLS12381G1_XMD:SHA-256_SSWU_RO_"`. |
+
+### 3. Response Alignment
+
+Returned receipts are fully order-preserving (1:1 index matching of request inputs/outputs):
+
+- **Mint (`POST /v1/mint/bolt11` & `/v1/mint/bolt11/batch`):** Nested in `pol_receipt` of each `BlindSignature` inside `signatures`.
```
Egge21M: should use the generic-method language instead `/v1/mint/{method}`

Egge21M (changes requested):
Love the design. A bit hard to grok at first, but I think that is unavoidable. I think some minor things are under specified. I think I found one or two inconsistencies

pol.md
a1denvalu3: > A mint MUST NOT insert the same issued item `B'` more than once in the same issued tree, and MUST NOT insert the same spent item `Y` more than once in the same spent tree. Duplicate insertion MUST be treated as an implementation error.

It's not physically possible to insert duplicates  because it's the same entry.

pol.md
a1denvalu3: I think no

thesimplekid (commented):
Not a technical comment but I am getting some rendering issues. Just me? 

Image

a1denvalu3:
@thesimplekid yes, github is really bad at rendering latex math. I'll change everything to code blocks.

DrVelvetFog:
Read `pol.md` in full at the current head (`2042052`). The transactional receipt design is the strongest part of this — binding `target_epoch` under a keyset signature is what turns "the mint omitted my token" from an unprovable complaint into signed perjury evidence, and the fraud-challenge document being self-contained is exactly right. Most of my comments are on the OTS layer, where I've been spending time recently (I have an open fix PR against the canonical calendar server, opentimestamps/opentimestamps-server#118, for a failure mode that's directly relevant here).

**1. Pending vs. anchored: the spec needs a rule for receipts that never confirm, and the mint should carry the upgrade burden.**

The manifest embeds the OTS receipt at epoch-construction time, which means it necessarily embeds a *pending* receipt — at submission the calendar has only committed to aggregate, and the `0x00 0x06` pending tag is all you have. Verification Step 2 then has every wallet contact a calendar's `/upgrade` endpoint. Two problems with that, one of them lived:

- **Calendar anchoring is not guaranteed to complete.** The failure mode behind opentimestamps/opentimestamps-server#114/#116 was the public calendars accepting commitments with HTTP 200 while the stamper was crashed — commitments sat unanchored for weeks. Under that condition every epoch here passes Step 1 (signature) forever while Step 2 can never complete, and the spec gives verifiers no rule beyond "still awaiting block confirmation." Worse, a mint that runs or colludes with its calendar can keep receipts pending indefinitely — at which point "committed to the Bitcoin blockchain" has quietly degraded to "committed to a calendar server's promise." Suggest: a verifier MUST treat a receipt still pending after some bound (calendar aggregation is typically hours; a day is generous) as an audit failure for that epoch, not a wait state.
- **Upgrade burden.** Rather than every wallet polling calendars, the mint SHOULD upgrade the receipt once anchored and republish the completed `.ots` alongside the manifest. The mint does the round-trip once; verifiers then check a self-contained proof offline against a block header. That also removes the availability dependency on calendars at verification time entirely.

**2. Sign over the digest, not over the receipt.**

Step 5/6 puts `ots_receipt` *inside* the signed manifest string. The receipt is the calendar's proof *about* the global digest, not a statement *by* the mint — folding it into the signature has two awkward consequences: the mint cannot swap in the upgraded (anchored) receipt without producing a second validly-signed manifest for the same epoch, and two distinct signed manifests for one epoch is precisely the shape an auditor should be treating as equivocation. If the signature covers everything up to and including `outstanding_balance` and the receipt travels beside it, any receipt that validly attests the same global digest is acceptable, upgrades are free, and multi-calendar aggregation (standard OTS client behavior, and the right hedge against a single calendar stalling per point 1) doesn't create signature churn.

**3. Epochs don't chain, so timestamping alone doesn't prevent equivocation.**

The commitment data is the concatenated per-keyset roots of one epoch. OTS proves existence, not uniqueness: a mint can maintain two divergent liability histories and timestamp both, showing each wallet the fork in which its own tokens are included. Adding the previous epoch's global digest (or manifest hash) to the commitment data makes a fork require *committed* divergence — any two parties comparing a single recent manifest hash then detect it, and the abstract's "immutable history" becomes literally a history rather than a set of independent snapshots. Cheap to add now, breaking to add later.

**4. Domain-separation parity between the two receipt signature versions.**

The BLS variant hashes with a proper DST (`Cashu_PoL_Receipt_BLS12381G1_...`); the secp256k1 variant signs `SHA256("<point_hex>:<epoch>")` raw — under the same per-amount key that performs BDHKE blind signing, over a point the wallet chose. I'm not claiming a concrete cross-protocol attack, but signing attacker-influenced messages under a multi-use key with no domain tag is the exact situation DSTs exist to preclude — and nothing currently distinguishes an issued-receipt message from a spent-receipt message (both are `<66-hex-point>:<decimal>`). A prefix pair like `Cashu_PoL_Receipt_Issued:` / `Cashu_PoL_Receipt_Spent:` costs nothing now and is painful to retrofit.

**Small one:** Step 2.4 says the anchored block should "match the manifest timestamp" — a Bitcoin block's `nTime` is median-time-past-bounded, roughly ±2h, and non-monotonic, so exact matching will fail on honest data. Specify a tolerance window (block time within N hours of the manifest timestamp) instead.

None of this touches the tree construction or the proof format, which read implementable as specified.

a1denvalu3:
@DrVelvetFog Thank you for the incredibly useful feedback. I've addressed your concerns.

DrVelvetFog:
Re-read the spec at `9e952ca` — all five points landed, and cleanly: the mint now carries the upgrade burden with pending-past-24h treated as an audit failure, the signature excludes the receipt (with the equivocation-false-positive rationale stated in-text, which future readers will thank you for), the commitment chains via `previous_global_digest` with a zero genesis, the receipt messages got proper Issued/Spent domain prefixes, and block-time matching became a tolerance window. Fast, thorough turnaround.

One gap the chaining change opens, worth closing while you're in the file: `previous_global_digest` is now committed and published, but the **Verification Protocol never checks it**. As written, a verifier still validates each epoch in isolation, so the chain doesn't yet bind anything — a mint maintaining two forked histories passes every existing step, since both forks chain internally. Suggest a step alongside Step 1: the verifier MUST compare this epoch's `previous_global_digest` against the previous epoch's recomputed global digest (or against the digest the wallet cached at its last audit), with a mismatch treated as an audit failure. It's also the one failure that's provable with nothing but two signed manifests, so it may deserve its own `challenge_type` — no tree walk required.

Tiny serialization note: the commitment data prepends `previous_global_digest` once globally, while the step-5 manifest message now carries it per keyset. Both are workable, but one sentence saying the asymmetry is intentional would keep an implementer from "fixing" one side to match the other.

pol.md
```diff
@@ -0,0 +1,300 @@
...
+2. Leaf index `I = SHA256(bytes(Y))` parsed as a big-endian integer.
+3. Walk up the tree and verify matching `root_spent`.
+
+### Step 5: Verify Liabilities Equation
+
+Ensure:
+
+```
+outstanding_balance == root_issued_sum - root_spent_sum
+```
```
orangeshyguy21: This step is a consistency check of the manifest rather than a proof. All three values here come from the mint itself. The difference isn't a trustworthy liability figure even when checked: the spent sum can be inflated with fabricated leaves that no wallet can dispute. `outstanding_balance` is effectively a floor set by auditing users but not a verified total. 

I'm not sure this can be avoided. If not, limitations such as this could be stated in a dedicated section.

Since the field is fully derivable from `root_issued_sum - root_spent_sum`, it could be dropped from the manifest (and the signed epoch string) entirely which also eliminates the need for this step.

a1denvalu3: > inflated with fabricated leaves that no wallet can dispute

Read the original pol scheme by @callebtc on how the mint is eventually caught doing this

orangeshyguy21: Thanks for the pointer. How the mint gets caught in the PoL scheme may be worth working into the spec. The catch is collective: "only when multiple users come together and prove that the aggregate of their balances is larger than the outstanding balance of the report, they can prove that the mint has added fake entries to this list." A single wallet only knows its own tokens, so alone it can only catch edge cases involving it's own balance out-sizing the reported outstanding balance:

Current spec hands the check to individual wallets ("Wallets periodically audit their held and spent tokens"). So Step 5 as written is a check that no single actor can meaningfully act on. The group version that functionally catches a cheating mint lacks a mechanism so far. Three pieces could be useful here:

1. **A format wallets can pool.** The Fraud Challenge is close, but it can't just be collected and summed: it carries one Merkle path where a pooled entry needs two ("counted as issued" and "not counted as spent"), it's built to say "this check failed" while a pooled entry says the opposite ("my token checks out fine, and yet the totals don't add up"). Perhaps a rule for deduping and summing entries against a manifest is appropriate too.

2. **Define what `/proofs/spent` returns when a token is *not* in the tree.** A pooled entry has to prove its token was never redeemed otherwise the mint can dismiss the whole sum with "those are already among the spends I counted." The natural proof is the mint showing the token's slot is empty: the same sibling walk as an inclusion proof, but starting from the default empty leaf and landing on the anchored `root_spent`. The ingredients already exist in the doc (the test vectors precompute the empty nodes, the Fraud Challenge carries `actual_value: 0`) only the endpoint's response for absent items is left undefined.

3. **Keyset lifecycle in the manifests.** The group check becomes increasingly effective if a keyset's outstanding balance is forced to shrink over time through regular rotations.

Question: is the intent that wallets run Step 5 individually (and if so, what should a single wallet do with the result?), or that auditors aggregate across wallets? If aggregation is the model, should this NUT define the format they'd collect?

a1denvalu3: > "only when multiple users come together and prove that the aggregate of their balances is larger than the outstanding balance of the report, they can prove that the mint has added fake entries to this list."

No, the users keep the root of the merkle-sum trees honest through random audits. Then, if the mint inflates (fill the spent book with fake ecash notes), the merkle-sum tree for the spent ecash will at some point be larger than the issued one, because when the keyset is rotated out it can no longer be minted from and therefore the issued book is pinned (only the spent book can change).

a1denvalu3: > Question: is the intent that wallets run Step 5 individually (and if so, what should a single wallet do with the result?), or that auditors aggregate across wallets? If aggregation is the model, should this NUT define the format they'd collect?

Yes the intent is very much that. I left undefined how one would go to formally challenge the Mint to provide a proof of inclusion, but the idea is that the wallet claiming fraud "publishes" this challenge out in the open somewhere and other wallets can query the same Mint for an inclusion proof in that epoch. If it fails to deliver then it proves fraud.

a1denvalu3:
image


https://gist.github.com/a1denvalu3/9cb8437714e0f5c0e9517611e811ca08

pol.md
orangeshyguy21: Agreed on the mechanism here for the independent wallet audit. It's the spent tree growing against the locked issued tree on an inactive keyset that leads to revealing the fraud to any single auditor.

Noting two undefined mechanisms for consideration that could increase wallet audit effectiveness:

1. **Committing to a keyset rotation.** Rotations aren't currently explicit. Is it sufficient that a keyset beyond a certain age becomes itself a red flag to the auditor? A keyset's age is provable from its first appearance in the anchored manifest history.

2. **The spent tree could be verified as append-only.** Two anchored roots don't show leaf *removals* between epochs, so a mint could delete fake spends (as real ones are added) on an inactive keyset undetected. Publishing each epoch's added leaves (old root + additions = new root) would close that. The cost ofc is an additional, potentially large, payload.

a1denvalu3: Point 2 is actually kind of critical. We need to rethink this. I propose we use a MMR (Merkle Mountain Range) instead of simple merkle-sum trees. The MMR is append-only and grows over time and we could capture a snapshot of it at every epoch, and commit it onchain with OTS. The proofs of inclusion the clients receive would be valid for that epoch, and further epochs would only require an "extension" thereof. The only downside is that with a MMR we cannot offer proofs of non-inclusion (but we don't really use them in the scheme).

DrVelvetFog:
Nice writeup — the wallet-side framing reads clean. Before I dig in, quick check on intent: is this gist meant to supersede the PR as the next revision, or is it a standalone/companion doc? Asking because a few things we hardened at `9e952ca` aren't in it, and I want to know whether they were dropped on purpose or just didn't survive the rewrite.

Three that stood out:

- **Manifest signs over `ots_receipt` again** (§3). We'd pulled the receipt out of the signed string specifically so the mint can upgrade a pending OTS proof to an anchored one without producing a *second* valid signature for the same epoch — which is exactly the equivocation shape an auditor should flag. Signing over it brings that false-positive back.
- **No `previous_global_digest` in `commitment_data`** (§3). Without epochs chaining, OTS only proves each epoch *existed*, not that it's *unique* — a mint can timestamp two divergent histories and both pass. The prev-digest link (zero genesis) is what makes "immutable history" literal rather than per-epoch.
- **Receipt message is back to bare `{B_hex}:{target_epoch}` / `{Y_hex}:{target_epoch}`** (§4.2) — the `Cashu_PoL_Receipt_Issued:` / `_Spent:` prefixes are gone. Issued and spent are signed under the same per-amount key, so those prefixes are what keep the two message spaces from overlapping.

Related, and still open from last round: even with chaining committed, the 5-step protocol (§6) never actually *checks* `previous_global_digest`. A forking mint passes all five steps unless step 1 also verifies this epoch's manifest links to the previous epoch's digest.

Happy to fold these into a small PR against whichever doc you're treating as canonical — just say which.

a1denvalu3:
@DrVelvetFog Are you referring to this spec document or the gist linked in the description?

DrVelvetFog:
Re-read both the current `pol.md` (`d0d9f4d`) and the auditor-service gist — all three land in both now, and the append-only sum-MMR switch is what did it, so no real disambiguation needed on my end:

- **Manifest signature excludes `ots_receipt`** — the signed string is manifest metadata only, so a pending→anchored receipt upgrade no longer yields a second valid signature for the epoch. The equivocation false-positive is gone. ✓
- **`previous_global_digest` chains epochs** in `commitment_data`, and the Append-Only Consistency Verification (distributed sibling-path prefix checks) is actually stronger than the digest-chain-at-step-1 I was asking for: any wallet holding a prior inclusion proof detects a historical edit the moment its old sibling path stops being a prefix of the new one. That also closes my last open point — the wallet protocol's Step 3 now *checks* append-only consistency, rather than the invariant being stated but never verified. ✓
- **Receipt messages carry the `Cashu_PoL_Receipt_Issued:` / `_Spent:` prefixes** again, so issued and spent stay in disjoint message spaces under the shared per-amount key. ✓

So against the current revision these are resolved — the MS-SMT→sum-MMR rewrite folded them in cleanly, and prefix-preservation is a nicer enforcement mechanism than epoch-digest chaining alone. Nothing to restore from my side; happy to keep reviewing as it firms up.

pol.md
```diff
@@ -0,0 +1,485 @@
...
+1. **Standard MMR Consistency Proofs:** Cryptographically proving that the peaks of the old MMR N are deterministically preserved and folded into the peaks of the new MMR M.
+2. **Distributed Sibling Path Prefix Checks:** Any client holding a valid inclusion proof (such as for active or spent ecash) can re-request a new proof in the following epoch and verify that the old sibling path is preserved as a direct prefix of the new sibling path. If the mint modified or removed any historical entries, the internal hashes would change, immediately violating prefix preservation.
+
+---
+
+## Epoch Manifests & On-Chain Commitments
+
+Every epoch interval (e.g., 24 hours), the mint constructs and signs an Epoch Manifest:
+
+1. **Sort Keysets:** Normalize all active unexpired `keyset_id` strings to lowercase hexadecimal representation, and then sort them alphabetically (lexicographically by their ASCII values).
```
orangeshyguy21: > all active unexpired `keyset_id` strings

"active" could be dropped here or the list item cleared up to include both active and inactive keysets, as active=false keysets would be included in the list.

pol.md
```diff
@@ -0,0 +1,485 @@
...
+
+### Note on Spent-Side Inflation & Keyset Rotations
+
+A naive threat model might suggest that a malicious mint could fabricate fake spent leaves (by appending random curve points `Y` to the Spent MMR) to artificially deflate its proven outstanding liabilities. However, because the mint owns the private signing keys for its keysets, a malicious mint can always generate valid ecash signatures (`C'`) for themselves, and then "validly" spend this self-signed fake ecash. Because the mint holds the preimages/secrets for these spends, a preimage/spend verification challenge is futile and cannot prevent this behavior.
+
+Instead, spent-side inflation is caught naturally by the **append-only nature of the MMR** and **keyset rotations/deactivations**:
+
+1. **Append-Only Immutability:** Once a leaf is appended to either MMR, it is permanent and cannot be deleted. If the mint fabricates a spend, that spent leaf is locked into history forever.
+2. **Keyset Rotation:** Eventually, keysets are deactivated/expired. Once deactivated, no more new ecash can be issued or spent under that keyset.
```
orangeshyguy21: > Once deactivated, no more new ecash can be issued or spent under that keyset.

This could be a typo. Should it be "Once expired,"

pol.md
```diff
@@ -0,0 +1,485 @@
...
+    "consistency_proof": {
+      "old_size": 125000,
+      "new_size": 126500,
+      "proof_hashes": [{ "hash": "8f3c...", "sum": 1000000, "height": 3 }]
+    }
+  }
+  ```
+- **Mint's Defense / Response & Verification:**
+  - **If the mint is NOT in the wrong:** The mint must respond to the challenge by publishing the **correct, valid MMR consistency proof** specified in the response schema (consisting of the list of sibling/peak proof hashes, sums, and heights) that mathematically merges the peaks of epoch `E_1` and the new elements to produce the peaks of epoch `E_2`.
+  - **Verification:** Third parties verify the mint's published consistency proof against the root hashes and sums in the two signed manifests. If the proof successfully validates the deterministic transition, the challenge is refuted and proven false. If the mint is silent or fails to provide a mathematically verifying consistency proof within a reasonable window, the challenge is sustained and the mint is proven fraudulent.
```
orangeshyguy21: Practically, the mint must be made aware of the existence of any fraud challenges for it be able to respond.

Since the coordination scheme between auditor, wallet, and mint is undefined (what auditors does the mint observe regularly for challenges? etc.) perhaps it is best to leave the definition of a proven fraud up to each 3rd party auditor.

For example: Auditor X might have a precondition for fraud claims that a mint has flagged observation of it's services such that it can responsibly enforce a time window for responses.

pol.md
```diff
@@ -0,0 +1,485 @@
...
+  - **If the mint is NOT in the wrong:** The mint must publish the above response containing the valid sum-MMR inclusion proof for the challenged `item` at `epoch_index` with the correct `value`.
+  - **Verification:** Third parties verify that the inclusion proof walk computes up to one of the mountain peaks, and that the peaks recursively peak-bagged from right-to-left match the manifest's `issued_mmr_root_hash` / `spent_mmr_root_hash` and sum. If it verifies, the challenge is refuted.
+
+---
+
+### 2. History Rewriting or Append-Only Violation (`append_only_violation`)
+
+- **Description:** The mint has modified, deleted, or reordered past leaves. This changes the internal hashes/sums of the tree, violating the strict append-only prefix preservation property.
+- **Challenge Schema:**
+  The challenger presents two signed epoch manifests along with two valid inclusion proofs for the exact same `leaf_index` showing a mismatch.
```
orangeshyguy21: Nothing in the verification walk uses `leaf_index`. A proof shows its leaf is *somewhere* in the tree, not *where*. So "two valid inclusion proofs for the exact same `leaf_index`" isn't something a third party can actually check.

That seems to allow **false convictions of an honest mint**:

1. Take a genuine proof of leaf **A** from epoch 11, and a genuine proof of a *different* leaf **B** from epoch 12. Both verify against their signed manifests.
2. Publish them as an `append_only_violation`, claiming both are index `45012`.
3. The items differ, so fraud is "proven" and the only permitted defense (showing a proof is mathematically invalid) fails, because both proofs *are* valid. The challenge can't be refuted as is.

A possible fix: the proof shape may already determine the position (`is_left` bits give the index within its mountain, peak heights give the mountain's offset), so verifiers could derive the index from the proof itself and reject any mismatches.

pol.md
```diff
@@ -0,0 +1,485 @@
...
+2. **Upgrade Proof (if pending):** If the receipt is still pending but within the timeout window, the verifier can attempt to fetch the Merkle path by posting the `ots_receipt` to a calendar upgrade endpoint (e.g., `https://alice.btc.calendar.opentimestamps.org/upgrade`). Note, however, that a properly functioning mint carries the upgrade burden and should publish already upgraded, offline-verifiable receipts.
+3. **Scan for Block:** Find block header attestation tag `0x00 0x05` (`A_BLOCKHEADER`).
+4. **Parse Height:** Decode the Bitcoin block height (serialized as a VarInt) immediately following the tag.
+5. **Confirm Block Timestamp:** Check via an independent explorer (e.g., `https://mempool.space/api/...`) that the block height exists and has sufficient confirmations. Furthermore, verify that the Bitcoin block's timestamp is within a reasonable tolerance window (e.g., within 4 hours) of the manifest `timestamp`, rather than requiring an exact match.
```
orangeshyguy21: As written, Step 2: Validate OpenTimestamps Attestation doesn't verify the receipt. It finds the attestation tag, reads the block height, and checks that block exists with a plausible timestamp, but it never executes the receipt's operations (the append/prepend/sha256 chain) to confirm they end at that block's merkle root, and never checks that the receipt's starting digest is the Global Digest of the manifest being audited. Those two checks honor the OTS receipt leading from a digest to a merkle root. Skipping them means a mint can hand-write a receipt pointing at any real block from the right day and pass Step 2 without anything ever being committed on chain.

There's also a practical blocker for the starting-digest check: the Global Digest preimage covers all keysets, but the epoch's keyset set isn't published anywhere (expired keysets aren't currently discoverable), so a verifier who recomputes and gets a mismatch can't tell fraud from a missing keyset. Publishing the epoch's keyset list (or `commitment_data` itself) would fix that, and would also give Step 3.1 a defined source for "the previous epoch's global digest" as currently the only way to obtain it is parsing it out of the previous `ots_receipt`, which the spec does not currently mention.

One more: the pending tag bytes look wrong. 
Pending is `0x83dfe30d2ef90c8e` (`0x06` doesn't appear in OTS), Bitcoin's is `0x0588960d73d71901`.

tests/pol-tests.md
```diff
@@ -0,0 +1,247 @@
...
+
+```
+009a6154b71113b7:1:2026-06-11T12:00:00Z:0000000000000000000000000000000000000000000000000000000000000000:3:2518b42edfff24ecc53c8897d1860783d1d26c41d61c378fe612cddeed877040:850:0:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855:0:850
+```
+
+### Signature Computation
+
+- **Message SHA-256 Digest:** `faaafafdc99bf27b8ba4d9b52d7ed5cd29d61c4a19ff65f8d4aaf49f6b964480`
+- **Auxiliary Random Data (`aux_rand`):** `b777e0270e6f6bd9302268a253ffda221ce9257a6e13349e198169745c45d72e`
+- **BIP-340 Schnorr Signature (`mint_signature`):** `f2720ae3c1eb13600acee4ee5694df89f12a51c32c171b52d71543f75534c197b2e98246828c41d85c31ea89fbed842df4b9622a6557899fb357aa4057920322`
```
orangeshyguy21: Tried reproducing these vectors and this signature doesn't verify against the pubkey and digest given here. Everything else in the file checks out: leaf hashes, both bagged roots, the manifest digest, and both receipt sigs in section 5 reproduce byte-for-byte. Just this one fails. 

Looks like the manifest string format changed after this was signed (the MMR sizes and previous_global_digest got added) and the signature never got regenerated.

orangeshyguy21 (commented):
Went through the MMR revision and ran the test vectors. I like the append-only redesign, it's a great match for both ledgers that conform to the same shape. I also like the challenge/response format.

One identified shortcoming in the OTS receipts
One issue with the test vectors
A couple of smaller wording things scattered about

One gap deliberately not commented inline: A wallets fighting chance to detect balance manipulation still depends on keysets regularly rotating, and nothing yet commits the mint to a schedule. We're evaluating ways the scheme could strengthen that commitment and will follow up separately.

DrVelvetFog:
`manifest_equivocation` is a clean addition, and marking it the one challenge with no possible defense is right — a mint signature over two conflicting states for the same epoch is unforgeable evidence, full stop.

One gap on making that *verifiable*, though: the "no defense" property rests on a third party being able to confirm both signatures are the mint's over the same `(keyset_id, epoch_index)`. But the payload embeds only `{timestamp, issued_mmr_root_hash, mint_signature}` per manifest, while the signature is over the full serialization — `keyset_id:epoch_index:timestamp:previous_global_digest:issued_mmr_size:issued_mmr_root_hash:issued_mmr_root_sum:spent_mmr_size:spent_mmr_root_hash:spent_mmr_root_sum:outstanding_balance`. With `previous_global_digest`, the sizes/sums, the spent-tree fields, and `outstanding_balance` absent, a verifier can't reconstruct either signed preimage — so neither BIP-340 signature can actually be checked, and the "definitive proof" isn't independently verifiable as written. Fix: carry each manifest's **complete signed serialization** (every field that goes into the digest), so anyone can recompute both SHA256 digests, verify both signatures against the mint's NUT-06 key, and confirm they share `(keyset_id, epoch_index)` but differ somewhere. (Same note technically applies to `append_only_violation`, but there the proofs verify against each epoch's *published* manifest — equivocation is the one case where both manifests are off-record by definition, so it's the acute one.)

Small thing on the same schema: I'd drop the "(or timestamp)". Equivocation is two different signed states for the same *logical epoch*, so key it strictly on `(keyset_id, epoch_index)` with any signed-content difference — `timestamp` is itself a signed field, not an alternate identity, and treating it as one invites false negatives (same epoch, different timestamp → not flagged).

pol.md
a1denvalu3: Addressed by the append-only sum-MMR redesign and the dedicated spent-side inflation/keyset rotation section. Fabricated spent leaves are now permanently committed and become detectable as inactive keysets wind down.

pol.md
a1denvalu3: Fixed in 9af6c6d: the global commitment now explicitly includes all unexpired keysets, both active and inactive.

pol.md
a1denvalu3: Fixed in 9af6c6d: deactivation now stops new issuance while allowing existing ecash to be spent until expiry; expiry stops both.

pol.md
a1denvalu3: Fixed in 9af6c6d: coordination and deadlines are now explicitly auditor policy, and silence is not cryptographic proof unless the mint verifiably received the challenge and the auditor applies its published policy.

pol.md
a1denvalu3: Fixed in 9af6c6d: verifiers must now derive the leaf index from the MMR peak heights and is_left path bits, and append-only challenges are rejected unless both derived positions equal the claimed index.

pol.md
a1denvalu3: Fixed in 9af6c6d: manifests publish global_digest plus the ordered epoch_keysets preimage data; verification binds the receipt starting digest, executes the complete OTS operation path to the Bitcoin Merkle root, and uses the correct full pending and Bitcoin attestation tags.

tests/pol-tests.md
a1denvalu3: Fixed in 12bcb58. Rechecked after the latest changes: the documented SHA256 digest reproduces and the BIP-340 signature verifies against the documented x-only public key. The new changes do not alter the signed serialization.

DrVelvetFog:
Re-read at the two 07-14 commits (`make equivocation proofs verifiable` + `address remaining audit review`) — both points from my last note land, cleanly:

- **Equivocation is now independently verifiable.** §3 has the challenger present two complete manifests, each carrying every field of the serialized message, and the verification step has third parties recompute each SHA256 digest and check both BIP-340 signatures against the mint's NUT-06 key. That closes it — a verifier can now reconstruct both signed preimages, so "no valid defense when both signatures verify" is actually checkable rather than asserted. ✓
- **Keyed strictly on `(keyset_id, epoch_index)` with any signed-content difference.** The "(or timestamp)" is gone and `timestamp` is back to being just one signed field, so same-epoch / different-timestamp can't slip through as a false negative. ✓

Reads implementable as written. Nothing open on my side — happy to keep reviewing as it firms up.

pol.md
```diff
@@ -0,0 +1,515 @@
...
+2. **Distributed Sibling Path Prefix Checks:** Any client holding a valid inclusion proof (such as for active or spent ecash) can re-request a new proof in the following epoch and verify that the old sibling path is preserved as a direct prefix of the new sibling path. If the mint modified or removed any historical entries, the internal hashes would change, immediately violating prefix preservation.
+
+---
+
+## Epoch Manifests & On-Chain Commitments
+
+Every epoch interval (e.g., 24 hours), the mint constructs and signs an Epoch Manifest:
+
+1. **Sort Keysets:** Normalize all unexpired `keyset_id` strings (both active and inactive) to lowercase hexadecimal representation, and then sort them alphabetically (lexicographically by their ASCII values).
+2. **Commitment Data:** Prepend the 32-byte binary `previous_global_digest` of the previous epoch (for the first epoch, this MUST be 32 bytes of zeros `0x00...00`). Then concatenate the UTF-8 lowercase `keyset_id`, 8-byte big-endian `issued_mmr_size`, 32-byte binary `issued_mmr_root_hash`, 8-byte big-endian `spent_mmr_size`, and 32-byte binary `spent_mmr_root_hash` for each keyset sequentially.
```
orangeshyguy21: The commitment data includes each keyset's tree sizes and root hashes but not the root sums. For a size-1 tree, the root hash doesn't contain its sum since sums only enter hashing when nodes combine. For those epochs the anchor doesn't commit the sum at all; it exists only in the signed string, which isn't what gets timestamped.

This case could occur on a low-activity spent or issued tree exacerbated by short epoch intervals.

Possible fix: include `issued_mmr_root_sum` and `spent_mmr_root_sum` (8-byte big-endian) in each keyset's `commitment_data` entry, and carry them in the `epoch_keysets` entries so the preimage stays reconstructable.

PR #390, opened 2026-06-17
NUT-XX: Non-Interactive Mint Delegation on Liveness Failure (Nimdolf)

KvngMikey:
## Summary

Based on this proposal: https://gist.github.com/lukechilds/302e7fa04b09d6e4568bdcd64cb43356

This PR adds a new optional spec, **NUT-XX: Non-Interactive Mint Delegation on Liveness Failure** (aka "Nimdolf"), which improves liveness guarantees for custodial mints that hold on-chain reserves. This NUT lets a mint nominate ordered **delegate mints** that can take over redemptions if it goes dark, **non-interactively**: the delegate never holds the original mint's signing key and the two mints never interact at failover time.

If a set of mints delegate to each other, they collectively gain a **1-of-n liveness** property, as long as one mint in the set stays online, users of every mint in the set can redeem.

## Files

- `xx.md`
- `README.md`
- `error_codes.md`


xx.md
```diff
@@ -0,0 +1,304 @@
...
+- `seq` is a strictly increasing, gap-free integer sequence number assigned by the original mint. It MUST start at `0` and increment by exactly `1` per event. It MUST NOT be reordered or reused.
+- `Y` is the spent proof's identifier, the hex-encoded compressed point `Y = hash_to_curve(secret)`.
+- `id` is the keyset ID ([NUT-02][02]) the proof was signed under.
+- `amount` is the proof's amount.
+- `witness` is the serialized witness data if the proof carried a [NUT-10][10] spending condition (as in [NUT-07][07]), otherwise `null`.
+
+> [!IMPORTANT]
+> The original mint MUST publish exactly one `SpentEvent` for every proof it transitions to `SPENT` ([NUT-07][07]), and MUST do so durably (an event, once published, is never removed). The feed reflects only `SPENT` proofs; `PENDING` proofs are not published.
+
+### Real-time stream (WebSocket)
```
Egge21M: I strongly believe SSEs are the better choice for a uni-directional firehose.

- Better reconnect behaviour
- Plain HTTP
- Less app layer complexity

xx.md
```diff
@@ -0,0 +1,304 @@
...
+  "params": {
+    "subId": <str>,
+    "payload": <SpentEvent>
+  }
+}
+```
+
+If a subscriber detects a gap (a received `seq` is greater than the last seen `seq` plus one), it MUST backfill the missing range via the catch-up endpoint below before treating its mirror as complete.
+
+### Catch-up (HTTP)
```
Egge21M, line 95: I think we should consider a bucket-like API where clients can requests multiple batches or single events from different indexes.

The current API shape is built for blocks of consecutive IDs

Egge21M (changes requested):

xx.md
KvngMikey: good call, i was torn between both options.

xx.md
KvngMikey: reworked this into a POST `/firehose/batch` taking `ranges` + `indexes`.

PR #391, opened 2026-06-18
NUT-XX: replay-protect pubkey quote lookup signatures

robwoodgate:
# Intended to be merged into #341 as an amendment.

## Problem

As written, the lookup signature is over the bare public key bytes. That message is static and over a value that is itself public, so the signature is a permanent bearer token: anyone who observes it once (mint logs, a TLS-terminating proxy, a MITM) can replay it forever to read that pubkey's quote history.

## Change

The signature now commits to a fresh, mint-scoped message:

```
"Cashu_MintQuoteLookup_v1" || mint_pubkey || pubkey
```

- `mint_pubkey` (from the NUT-06 info response) binds the signature to one mint, so it cannot be replayed against another mint. The mint pubkey is used rather than the mint URL to avoid URL canonicalization ambiguity. Mints supporting this NUT MUST advertise a `pubkey`.

### DIFF
--- a/xx.md
+++ b/xx.md
@@ -1,4 +1,4 @@
-# NUT-29: Mint Quote Lookup by Public Key
+# NUT-XX: Mint Quote Lookup by Public Key
 
 `optional`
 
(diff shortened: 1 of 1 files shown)


TheMhv (approved):
LGTM!

ACK e799ab41a0943d6906e12cde7d6c4fd04dcb4535

a1denvalu3:
> nonce (random 16 bytes) dedups within that window, so the mint only retains nonces for max_age. No challenge round trip.

I think it's not worth it to add `nonce`. it would mean the Mint has to keep track of them and run background tasks to clear them etc... no. A timestamp, with a wide +/-10 minutes tolerance, is enough.

robwoodgate:
> > nonce (random 16 bytes) dedups within that window, so the mint only retains nonces for max_age. No challenge round trip.
> 
> I think it's not worth it to add `nonce`. it would mean the Mint has to keep track of them and run background tasks to clear them etc... no. A timestamp, with a wide +/-10 minutes tolerance, is enough.

nonce is time bound by timestamp - it covers the replay protection concern within the time allocation. 

The mint only needs to track nonces for `max_age` seconds. After that, the signature can't be replayed because of age. It removes the need for an extra call to the mint to get a nonce challenge (which the mint would ALSO have to track). 

Removing it means we accept replay in the `max_age` window.

robwoodgate:
PR updated following offline discussions. Nonce and timestamp now removed as agreed.

@callebtc - this should be ready to merge into #341, making it in ready for final review and NUT numbering.

a1denvalu3 (approved):

TheMhv:
> PR updated following offline discussions. Nonce and timestamp now removed as agreed.

How this will work? Removing timestamp will be able replay attack again

PR #402, opened 2026-07-14
tests: add NUT-28 HTLC example proof

robwoodgate:
Adds an HTLC example to the NUT-28 test vectors, using the same fixed keypairs as the P2BK P2PK example.

Tests pinned in:

- [ ] Cashu-TS: https://github.com/cashubtc/cashu-ts/pull/786
- [ ] CDK: wallet side
- [ ] Nutshell: wallet side

It pins the blinding slot order for HTLC secrets: the hash lock occupies slot 0 unblinded, the `pubkeys` key blinds at slot 1 and the `refund` key at slot 2. The `locktime` tag sits between them in the JSON as a check that slots are numbered by the canonical `[data, ...pubkeys, ...refund]` order, not by tag position. This mirrors the fix in cashubtc/cashu-ts#763, where keys were previously blinded from slot 0.

The blinded keys are the existing slot 1 and 2 table entries, so the example is cross-checkable against the published vectors. The hash lock reuses the documented NUT-14 preimage pair. C and dleq are internally consistent (DLEQ verifies via reblind).

## Rationale

Cashu-TS had a bug where `slot 0` was ignored in HTLC. This test prevents other implementations doing the same.
Vectors confirmed and pinned in https://github.com/cashubtc/cashu-ts/pull/786


robwoodgate:
@ye0man - this should be a shoo-in. Be good to get the vector into CDK / Nutshell for completeness

PR #405, opened 2026-07-17
NUT-29: use valid curve points in the batch mint example

robwoodgate:
Two output `B_` values in the batch mint request example are not valid secp256k1 points, so implementations that validate points on parse (e.g. cdk) would reject the example verbatim. Replaces them with the compressed points for the well-known test keys `02` and `03`.

Found while pinning the vectors for https://github.com/cashubtc/nuts/pull/404.

### DIFF
--- a/29.md
+++ b/29.md
@@ -161,12 +161,12 @@ Content-Type: application/json
     {
       "amount": 4,
       "id": "009a1f293253e41e",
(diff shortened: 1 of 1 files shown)


robwoodgate:
@ye0man - this doesn't need assigning, just merging, and that needs an approving review

ye0man (approved):
ACK

callebtc (approved):

PR #406, opened 2026-07-17
NUT-03/05: state unit consistency normatively

robwoodgate:
Error codes `11009`/`11010` (inputs/outputs of multiple units, inputs and outputs not of same unit) already imply that a transaction cannot mix currency units, and both nutshell and cdk enforce it, but neither NUT-03 nor NUT-05 states the rule in prose. This adds one normative sentence to each.

Surfaced by https://github.com/cashubtc/nuts/pull/404, where the reasoning for not committing to the output keyset `id` relies on this invariant.

### DIFF
--- a/03.md
+++ b/03.md
@@ -37,6 +37,8 @@ With the data being of the form `PostSwapRequest`:
 }
 ```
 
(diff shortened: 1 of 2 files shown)


robwoodgate:
@ye0man - just needs an approving review

ye0man (approved):
ACK

callebtc (approved):

Issue #344, opened 2026-02-21
GH Pages are not updating


robwoodgate:
The repo's pages seem to be out of date: https://cashubtc.github.io/nuts/

Only lists NUTS up to NUT-22.

prusnak:
> The repo's pages seem to be out of date: https://cashubtc.github.io/nuts/

I see nuts up to 28 on this URL

prusnak:
• It’s configuration drift.

  The GitHub Pages workflow:

  1. Checks out gh-pages-builder, which contains mkdocs.yml.
  2. Copies main into docs/.
  3. Builds every Markdown file.

  But the manually maintained nav: in mkdocs.yml (https://github.com/cashubtc/nuts/blob/gh-pages-builder/mkdocs.yml) explicitly stops at NUT-22 and was last updated in January
  2025. MkDocs still builds NUTs 23–30, so [NUT-30 is directly accessible](https://cashubtc.github.io/nuts/30/) (https://cashubtc.github.io/nuts/30/), but pages omitted from nav: don’t appear in the sidebar.

  So: nothing is wrong with deployment; the separate builder branch was simply never updated. Add NUTs 23–30 to gh-pages-builder/mkdocs.yml—or, preferably, move/generate the
  navigation from main so it can’t drift again.

prusnak:
fix in https://github.com/cashubtc/nuts/pull/408

PR #408, opened 2026-07-18
fix missing nuts in navigation

prusnak:
fix for https://github.com/cashubtc/nuts/issues/344

PS: ideally the builder would live in the main branch too, so we won't forget updating this

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -68,7 +68,15 @@ nav:
           - NUT-19 - Cached Responses: 19.md
           - NUT-20 - Signature on Mint Quote: 20.md
           - NUT-21 - Clear authentication: 21.md
-          - NUT-22 - Blind authentication: 22.md          
+          - NUT-22 - Blind authentication: 22.md
+          - "NUT-23 - Payment Method: BOLT11": 23.md
+          - NUT-24 - HTTP 402 Payment Required: 24.md
+          - "NUT-25 - Payment Method: BOLT12": 25.md
+          - NUT-26 - Payment Request Bech32m Encoding: 26.md
+          - NUT-27 - Nostr Mint Backup: 27.md
+          - NUT-28 - Pay to Blinded Key (P2BK): 28.md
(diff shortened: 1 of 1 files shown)


robwoodgate (approved):

thesimplekid (approved):

mkdocs.yml
```diff
@@ -68,7 +68,15 @@ nav:
...
-          - NUT-22 - Blind authentication: 22.md          
+          - NUT-22 - Blind authentication: 22.md
+          - "NUT-23 - Payment Method: BOLT11": 23.md
+          - NUT-24 - HTTP 402 Payment Required: 24.md
+          - "NUT-25 - Payment Method: BOLT12": 25.md
+          - NUT-26 - Payment Request Bech32m Encoding: 26.md
+          - NUT-27 - Nostr Mint Backup: 27.md
+          - NUT-28 - Pay to Blinded Key (P2BK): 28.md
+          - NUT-29 - Batched Mint: 29.md
+          - "NUT-30 - Payment Method: Onchain": 30.md
```
thesimplekid, line 79: Does it matter some of them are in quotes?

thesimplekid (commented):
This looks fine I would merge but it shows as a locked branch so I'm not sure if merging is correct or if there is another way it is supposed to be updated, maybe @callebtc you know?

Image

mkdocs.yml
prusnak: if the text contains `:` it needs to be quoted

robwoodgate:
Probably best ultimately to script the update of these, but might need branch unprotecting temporarily to allow this update. 

Will be locked normally because updates are usually built automatically I suspect.

StephenChi-hi (approved):

thesimplekid:
> Probably best ultimately to script the update of these, but might need branch unprotecting temporarily to allow this update.
> 
> Will be locked normally because updates are usually built automatically I suspect.

I can override the lock, I'm just not the one that set this up so I'm not sure if that's correct or will break things.

PR #410, opened 2026-07-23
NUT-Exchange: Atomic Multi-Asset Exchange

joemphilips:
Adds an optional NUT for an atomic exchange of two existing Cashu asset classes at one mint. Participants lock bearer proofs to exact blinded receive outputs via a new NUT-10 `PAY_TO_UNLOCK` condition; the mint validates the conditions, conserves each asset class independently, and commits all input spends and output signatures in one database transaction — or nothing.

**Context.** This branched out of ongoing work on prediction markets with Cashu (see #337). The key observation: we do not need HTLC-based atomic swaps — genuinely troublesome, as anyone who has dealt with Lightning knows — like we usually reach for in Bitcoin, when both assets are managed by a single mint, because Cashu has a different trust model. A Cashu mint already validates proofs, tracks spentness, and issues blind signatures, so it can serve directly as the atomic settlement layer. This removes the interactive claim sequence, the both-wallets-online requirement, and the free-option locktime race of a client-to-client swap.

Custody is enforced by blind signatures plus the condition binding (any submitter/relay cannot steal or redirect); recovery via NUT-09; post-expiry refund authorized by a signed `refund` key. Trust floor matches NUT-11 P2PK plus one atomic commit; no transparency layer.

Draft for discussion — final NUT number/name, endpoint, and fee policy are open.

### DIFF
--- a/error_codes.md
+++ b/error_codes.md
@@ -1,52 +1,52 @@
 # NUT Errors
 
-| Code  | Description                                     | Relevant nuts                                          |
-| ----- | ----------------------------------------------- | ------------------------------------------------------ |
-| 10001 | Proof verification failed                       | [NUT-03][03], [NUT-05][05]                             |
-| 11001 | Proofs already spent                            | [NUT-03][03], [NUT-05][05]                             |
-| 11002 | Proofs are pending                              | [NUT-03][03], [NUT-05][05]                             |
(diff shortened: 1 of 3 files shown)


joemphilips:
Manual self-review done. Open for discussion

PR #415, opened 2026-07-28
Wallet request expiry for mint quotes

thesimplekid:
- CDK: 
- Nutshell: 
- cashu-ts:

### DIFF
--- a/23.md
+++ b/23.md
@@ -16,9 +16,14 @@ For the `bolt11` method, the wallet includes the following specific `PostMintQuo
 {
   "amount": <int>,
   "unit": <str_enum[UNIT]>,
-  "description": <str> // Optional
+  "description": <str>, // Optional
+  "expiry_seconds": <int> // Optional
 }
 ```
+
+`expiry_seconds` is the requested quote lifetime in seconds from quote creation. If omitted, the mint uses its default lifetime.
+
+Mints **MUST** return error `11018` if `expiry_seconds` is not positive, exceeds their maximum, or cannot be honored exactly. Mints **MUST NOT** silently clamp it.
 
 The mint responds with a `PostMintQuoteBolt11Response`:
 
@@ -33,7 +38,7 @@ The mint responds with a `PostMintQuoteBolt11Response`:
   "amount_issued": <int>,
   "updated_at": <int>,
   "state": <str_enum[STATE]>, // Deprecated, optional
-  "expiry": <int|null>
+  "expiry": <int>
 }
 ```
 
@@ -47,14 +52,14 @@ The mint responds with a `PostMintQuoteBolt11Response`:
 
 Wallets **SHOULD** use `amount_paid` and `amount_issued` instead of `state` whenever these fields are present.
 
-`expiry` is the Unix timestamp until which the `request` can be paid (i.e. the bolt11 invoice expiry).
+`expiry` is the invoice timestamp plus its BOLT11 expiry duration, expressed as a Unix timestamp. If `expiry_seconds` was set, it **MUST** be used as the invoice expiry duration.
 
 ### Example
 
 Request with curl:
 
 ```bash
-curl -X POST http://localhost:3338/v1/mint/quote/bolt11 -d '{"amount": 10, "unit": "sat"}' -H "Content-Type: application/json"
+curl -X POST http://localhost:3338/v1/mint/quote/bolt11 -d '{"amount": 10, "unit": "sat", "expiry_seconds": 100}' -H "Content-Type: application/json"
 ```
 
 Response:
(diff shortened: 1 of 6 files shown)


a1denvalu3 (approved):

robwoodgate:
Not sure I understand the rationale for this. Quote expiry binds the mint to a monitoring period - why would a wallet voluntarily reduce that and risk a payment being missed?

I kind of understand it in cases where the mint doesn't have an expiry and the wallet wants to enforce one. But mints committing to monitor quotes perpetually are just storing up pain for themselves in the future.

23.md
```diff
@@ -47,14 +52,14 @@ The mint responds with a `PostMintQuoteBolt11Response`:
 
 Wallets **SHOULD** use `amount_paid` and `amount_issued` instead of `state` whenever these fields are present.
 
-`expiry` is the Unix timestamp until which the `request` can be paid (i.e. the bolt11 invoice expiry).
+`expiry` is the invoice timestamp plus its BOLT11 expiry duration, expressed as a Unix timestamp. If `expiry_seconds` was set, it **MUST** be used as the invoice expiry duration.
```
robwoodgate, line 55: Should the default include a buffer for the HTLC timeouts too? If a payment is made close to LN invoice expiry, it may arrive after the expiry time and not be credited.

Egge21M: Mints should honor payments even expiry is reached. Wallets should do their best to check quotes that are technically expired but might have received a payment.

thesimplekid:
> Not sure I understand the rationale for this. Quote expiry binds the mint to a monitoring period - why would a wallet voluntarily reduce that and risk a payment being missed?
> 
> I kind of understand it in cases where the mint doesn't have an expiry and the wallet wants to enforce one. But mints committing to monitor quotes perpetually are just storing up pain for themselves in the future.

A wallet would want to reduce the number of quotes it must monitor. I think the best example for this is bolt12 right now. Currently, bolt12 never expires and this is useful for long running repeated payments however in the event bolt12 replaces bolt11 as the default invoice a wallet like numo would want to create short lived ones. Only the wallet knows if it wants a short one or the long one, so it must request and then the mint can decide if it supports this or not.

crodas:
I agree with @robwoodgate; I see this would generate friction for very little benefit. 

> A wallet would want to reduce the number of quotes it must monitor

Why? It is roughly ~7,000 txs every 10 minutes since the mint's inception.

thesimplekid:
> I agree with @robwoodgate; I see this would generate friction for very little benefit.

I don't see how it adds friction it is intentionally backwards compatible if wallets do not update or do not want to use this it works exactly how it does today where the mint just chooses the expiry.  

> > A wallet would want to reduce the number of quotes it must monitor
> 
> Why? It is roughly ~7,000 txs every 10 minutes since the mint's inception.

I gave the bolt12 example of why a wallet might want to do this. I don't know how this 7000 tx is relevant we don't know how many txs a wallet makes or will make. This could lead to a wallet having many quotes to monitor; this could be made more efficient by using batch checking, however this come with a privacy trade off and still has a cost. So allowing wallets to set a short expiry when they want it; reduces cost for both mint and wallet.  We have already had issues with wallets having many quotes to check on start up, and this becomes more of a pain point with long running repaybale quotes like bolt12 and older more active wallets.

robwoodgate:
@thesimplekid - the bolt12 example is reason enough to do it. Wallets should have a knob to reduce their tracking liability.

My point about mints not having a long backstop expiry policy stands. Tracking quotes forever - even bolt12 - seems like an eventual resource issue.

thesimplekid:
> My point about mints not having a long backstop expiry policy stands. Tracking quotes forever - even bolt12 - seems like an eventual resource issue.

I agree this is a problem, but don't think it is changed by this PR as its a preexisting.

robwoodgate (approved):

30.md
```diff
@@ -41,21 +46,23 @@ Where:
...
 - `request` is the Bitcoin address to send funds to
-- `expiry` is the Unix timestamp until which the mint quote is valid
+- `expiry` is the Unix timestamp until which the mint quote is valid, or `null` if `expiry_seconds` was omitted
 - `pubkey` is the public key from the request
 - `amount_paid` is the total confirmed amount paid to the request in UTXOs that are eligible for minting
 - `amount_issued` is the amount of ecash that has been issued for the given mint quote
 - `updated_at` is defined in [NUT-04][04]
 
-If `expiry` is not `null`, the wallet **SHOULD NOT** send payments to the request after `expiry`. Mints **MUST** keep monitoring transactions they detected before `expiry` until the transaction reaches the required number of confirmations or is evicted or replaced. Payments first detected by the mint after `expiry` **MUST NOT** increase `amount_paid`.
+The Bitcoin address does not expire; `expiry` only limits when payments are credited to the quote. If `expiry_seconds` was set, `expiry` **MUST** equal the quote creation time plus `expiry_seconds`.
```
Egge21M, line 55: Didn't we settle on that the mint SHOULD still watch expired quotes?

PR #416, opened 2026-07-29
NUT-06: require signed mint info

a1denvalu3:
# Implementations

- [ ] Nutshell: https://github.com/cashubtc/nutshell/pull/1125
- [x] Cashu-TS: https://github.com/cashubtc/cashu-ts/pull/990
- [ ] CDK: TBC: https://github.com/cashubtc/cdk/pull/2424

## Summary

- standardize derivation of the mint identity key from the mint seed
- make the mint identity public key mandatory
- require a BIP-340 Schnorr signature over the RFC 8785 canonicalized info response, excluding `time` and `signature`
- require wallets to reject missing, malformed, or invalid signatures
- add a reproducible, cryptographically verified example

## Validation

- `npx prettier@3.9.1 --check 06.md`
- verified that the example public key derives from the documented seed and that its signature validates against the specified payload

### DIFF
--- a/06.md
+++ b/06.md
@@ -19,7 +19,8 @@ With the mint's response being of the form `GetInfoResponse`:
 ```json
 {
   "name": "Bob's Cashu mint",
-  "pubkey": "0283bf290884eed3a7ca2663fc0260de2e2064d6b355ea13f98dec004b7a7ead99",
+  "pubkey": "02d53b2972b5fcfda2861d937a40bd4dd4170718276e35ac82a84fa60c3a365046",
+  "signature": "efed7369effde640baee2f089681a57fd559e45ebf5bbf9a1f2200bb4137291453b617d7955f0e47426be39f21c4253e92bac8ada04f730021a3e313465334d9",
   "version": "Nutshell/0.15.0",
   "description": "The short mint description",
   "description_long": "A description that can be a long piece of text.",
@@ -87,8 +88,11 @@ With the mint's response being of the form `GetInfoResponse`:
 }
 ```
 
+For reproducibility, the example uses the mint seed `NUT-06 example mint seed` and zero-filled auxiliary randomness for BIP-340 signing.
+
 - (optional) `name` is the name of the mint and should be recognizable.
-- (optional) `pubkey` is the hex pubkey of the mint.
+- (mandatory) `pubkey` is the mint's identity public key as a lowercase hex-encoded, 33-byte compressed secp256k1 public key.
+- (mandatory) `signature` is the lowercase hex encoding of the mint's 64-byte BIP-340 Schnorr signature on the response payload, constructed as specified below.
 - (optional) `version` is the implementation name and the version of the software running on this mint separated with a slash "/".
 - (optional) `description` is a short description of the mint that can be shown in the wallet next to the mint's name.
 - (optional) `description_long` is a long description that can be shown in an additional field.
@@ -99,6 +103,30 @@ With the mint's response being of the form `GetInfoResponse`:
 - (optional) `time` is the current time set on the server. The value is passed as a Unix timestamp integer.
 - (optional) `tos_url` is the URL pointing to the Terms of Service of the mint.
 - (optional) `nuts` indicates each NUT specification that the mint supports and its settings. The settings are defined in each NUT separately.
+
+## Mint identity key
+
+The mint identity key is deterministically derived from the mint seed. Given the exact `seed` string used by the mint:
+
+1. Encode `seed` as UTF-8.
+2. Compute `secret_key = SHA256(seed)`.
+3. Interpret `secret_key` as a 32-byte big-endian integer. It **MUST** be a valid secp256k1 secret key in the range `1..n-1`, where `n` is the order of the secp256k1 curve.
+4. Compute `pubkey = secret_key * G` and serialize it in 33-byte compressed SEC1 format.
+
+The mint **MUST** return this public key in `GetInfoResponse.pubkey`.
+
+## Signature
+
+The mint **MUST** sign every `GetInfoResponse` with the identity key corresponding to `pubkey`. To construct and sign the message:
+
+1. Start with the complete `GetInfoResponse` JSON object.
+2. Remove the top-level `signature` and `time` members. The `time` member is excluded so that an otherwise unchanged response has a stable signed payload.
+3. Canonicalize the resulting JSON object using the JSON Canonicalization Scheme (JCS) defined in [RFC 8785]. Encode the canonical JSON as UTF-8.
+4. Compute the SHA-256 hash of those bytes.
+5. Sign the 32-byte hash using [BIP-340] Schnorr over secp256k1.
+6. Encode the resulting 64-byte signature as lowercase hexadecimal and return it in `GetInfoResponse.signature`.
+
+Wallets **MUST** reject the response if `pubkey` or `signature` is missing or malformed, or if BIP-340 verification fails. For BIP-340 verification, wallets use the 32-byte x-coordinate of the compressed `pubkey`.
 
 With curl:
 
@@ -119,3 +147,5 @@ curl -X GET https://mint.host:3338/v1/info
 [10]: 10.md
 [11]: 11.md
 [12]: 12.md
+[RFC 8785]: https://www.rfc-editor.org/rfc/rfc8785
+[BIP-340]: https://github.com/bitcoin/bips/blob/master/bip-0340.mediawiki
--- a/06.md
+++ b/06.md
@@ -20,7 +20,7 @@ With the mint's response being of the form `GetInfoResponse`:
 {
   "name": "Bob's Cashu mint",
   "pubkey": "0338596797cef0627f653cd6568387361b00314add55d9f1ea9c94f46ae421e3da",
-  "signature": "d1ae4c576e3ffa667df795b06655a5400e590653ae1f60f2a256c01e794a70328fee6fc8f4139c86458d12c0d4f04781581a5890a9ee83b9f7c72c78a641add0",
+  "signature": "256c3fc7773edf908d8d86bf0b1efe5cecbca0e152dfcc4e3d9cb9e5d3e643daa67f62aeb23c640783d46538a601d7907afed9b7db19f7dbdaccfc6dc8d8a227",
   "version": "Nutshell/0.15.0",
   "description": "The short mint description",
   "description_long": "A description that can be a long piece of text.",
@@ -100,7 +100,7 @@ For reproducibility, the example uses the UTF-8 encoding of `NUT-06 example mint
 - (optional) `motd` is the message of the day that the wallet must display to the user. It should only be used to display important announcements to users, such as scheduled maintenances.
 - (optional) `icon_url` is the URL pointing to an image to be used as an icon for the mint. Recommended to be squared in shape.
 - (optional) `urls` is the list of endpoint URLs where the mint is reachable from.
-- (optional) `time` is the current time set on the server. The value is passed as a Unix timestamp integer.
+- (mandatory) `time` is the mint's current time as a Unix timestamp integer. It is included in the signed payload and **MUST** be within 3600 seconds of the verifier's current time.
 - (optional) `tos_url` is the URL pointing to the Terms of Service of the mint.
 - (optional) `nuts` indicates each NUT specification that the mint supports and its settings. The settings are defined in each NUT separately.
 
@@ -122,13 +122,13 @@ The mint **MUST** return this public key in `GetInfoResponse.pubkey`.
 The mint **MUST** sign every `GetInfoResponse` with the identity key corresponding to `pubkey`. To construct and sign the message:
 
 1. Start with the complete `GetInfoResponse` JSON object.
-2. Remove the top-level `signature` and `time` members. The `time` member is excluded so that an otherwise unchanged response has a stable signed payload.
+2. Remove the top-level `signature` member. The `time` member remains in the object and is therefore included in the signed payload.
 3. Canonicalize the resulting JSON object using the JSON Canonicalization Scheme (JCS) defined in [RFC 8785]. Encode the canonical JSON as UTF-8.
 4. Compute the SHA-256 hash of those bytes.
 5. Sign the 32-byte hash using [BIP-340] Schnorr over secp256k1.
 6. Encode the resulting 64-byte signature as lowercase hexadecimal and return it in `GetInfoResponse.signature`.
 
-Wallets **MUST** reject the response if `pubkey` or `signature` is missing or malformed, or if BIP-340 verification fails. For BIP-340 verification, wallets use the 32-byte x-coordinate of the compressed `pubkey`.
+Wallets **MUST** reject the response if `pubkey`, `signature`, or `time` is missing or malformed, if BIP-340 verification fails, or if `time` differs from the wallet's current Unix time by more than 3600 seconds. A difference of exactly 3600 seconds is valid. For BIP-340 verification, wallets use the 32-byte x-coordinate of the compressed `pubkey`.
 
 With curl:
 
--- a/06.md
+++ b/06.md
@@ -19,8 +19,8 @@ With the mint's response being of the form `GetInfoResponse`:
 ```json
 {
   "name": "Bob's Cashu mint",
-  "pubkey": "02d53b2972b5fcfda2861d937a40bd4dd4170718276e35ac82a84fa60c3a365046",
-  "signature": "efed7369effde640baee2f089681a57fd559e45ebf5bbf9a1f2200bb4137291453b617d7955f0e47426be39f21c4253e92bac8ada04f730021a3e313465334d9",
+  "pubkey": "0338596797cef0627f653cd6568387361b00314add55d9f1ea9c94f46ae421e3da",
+  "signature": "d1ae4c576e3ffa667df795b06655a5400e590653ae1f60f2a256c01e794a70328fee6fc8f4139c86458d12c0d4f04781581a5890a9ee83b9f7c72c78a641add0",
   "version": "Nutshell/0.15.0",
   "description": "The short mint description",
   "description_long": "A description that can be a long piece of text.",
@@ -88,7 +88,7 @@ With the mint's response being of the form `GetInfoResponse`:
 }
 ```
 
-For reproducibility, the example uses the mint seed `NUT-06 example mint seed` and zero-filled auxiliary randomness for BIP-340 signing.
+For reproducibility, the example uses the UTF-8 encoding of `NUT-06 example mint seed` as the mint seed and zero-filled auxiliary randomness for BIP-340 signing.
 
 - (optional) `name` is the name of the mint and should be recognizable.
 - (mandatory) `pubkey` is the mint's identity public key as a lowercase hex-encoded, 33-byte compressed secp256k1 public key.
@@ -106,12 +106,14 @@ For reproducibility, the example uses the mint seed `NUT-06 example mint seed` a
 
 ## Mint identity key
 
-The mint identity key is deterministically derived from the mint seed. Given the exact `seed` string used by the mint:
+The mint identity key is deterministically derived from the mint seed. Given the exact `seed` bytes used by the mint:
 
-1. Encode `seed` as UTF-8.
-2. Compute `secret_key = SHA256(seed)`.
-3. Interpret `secret_key` as a 32-byte big-endian integer. It **MUST** be a valid secp256k1 secret key in the range `1..n-1`, where `n` is the order of the secp256k1 curve.
+1. Set the one-byte counter `ctr` to `0x00`.
+2. Compute `secret_key = HMAC-SHA256(key=seed, data=MINT_IDENTITY_DOMAIN_SEPARATOR || ctr)`, where `MINT_IDENTITY_DOMAIN_SEPARATOR = b"Cashu_Mint_Identity_v1"` and `||` denotes byte concatenation.
+3. Interpret `secret_key` as a 32-byte big-endian integer. If it is not in the range `1..n-1`, where `n` is the order of the secp256k1 curve, increment `ctr` and retry (up to 256 attempts).
 4. Compute `pubkey = secret_key * G` and serialize it in 33-byte compressed SEC1 format.
+
+Derivation **MUST** fail if no valid secret key is found after 256 attempts.
 
 The mint **MUST** return this public key in `GetInfoResponse.pubkey`.
 
--- a/06.md
+++ b/06.md
@@ -30,7 +30,7 @@ With the mint's response being of the form `GetInfoResponse`:
 {
   "name": "Bob's Cashu mint",
   "pubkey": "0338596797cef0627f653cd6568387361b00314add55d9f1ea9c94f46ae421e3da",
-  "signature": "6495ee5d693994f9a7e821e5baaf87791c50eb98bae9b2047bdb18cfbed4d92298816032c97aa6382ce6b4226efc8401d1ef7e2d8731bbb658bb6c44acc86df5",
+  "signature": "e91f667871ff922f7430aee729bc0252ff893ec2b05ef2b0f98376f7f5119b152029b363f89c35ffa2d8e3395c889c5caa2484c8b716cc18881dad7b6ef9a739",
   "challenge": "000102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1f",
   "version": "Nutshell/0.15.0",
   "description": "The short mint description",
@@ -135,12 +135,14 @@ The mint **MUST** sign every `GetInfoResponse` with the identity key correspondi
 
 1. Start with the complete `GetInfoResponse` JSON object.
 2. Remove the top-level `signature` member. The `time` member and, if present, the `challenge` member remain in the object and are therefore included in the signed payload.
-3. Canonicalize the resulting JSON object using the JSON Canonicalization Scheme (JCS) defined in [RFC 8785]. Encode the canonical JSON as UTF-8.
-4. Compute the SHA-256 hash of those bytes.
-5. Sign the 32-byte hash using [BIP-340] Schnorr over secp256k1.
+3. Canonicalize the resulting JSON object using the JSON Canonicalization Scheme (JCS) defined in [RFC 8785]. Encode the canonical JSON as UTF-8 to obtain `bytes`.
+4. Compute `message_hash = tagged_hash("Cashu_MintInfo_v1", bytes)`.
+5. Sign the 32-byte `message_hash` using [BIP-340] Schnorr over secp256k1.
 6. Encode the resulting 64-byte signature as lowercase hexadecimal and return it in `GetInfoResponse.signature`.
 
-Wallets **MUST** reject the response if `pubkey`, `signature`, or `time` is missing or malformed, if BIP-340 verification fails, or if `time` differs from the wallet's current Unix time by more than 3600 seconds. A difference of exactly 3600 seconds is valid. For BIP-340 verification, wallets use the 32-byte x-coordinate of the compressed `pubkey`.
+The `tagged_hash` function uses the [BIP-340] construction `SHA256(SHA256(tag) || SHA256(tag) || message)`, where `tag` is encoded as UTF-8, each inner SHA-256 result is its raw 32-byte digest, and `||` denotes byte concatenation. For this tag, `SHA256("Cashu_MintInfo_v1") = 916a34ebf6f2244d64490e8eb2b5e7af19cdc45aa6176160186fc6543aa20d9b`.
+
+Wallets **MUST** reject the response if `pubkey`, `signature`, or `time` is missing or malformed, if BIP-340 verification fails, or if `time` differs from the wallet's current Unix time by more than 3600 seconds. A difference of exactly 3600 seconds is valid. For BIP-340 verification, wallets recompute `message_hash` as specified above and use the 32-byte x-coordinate of the compressed `pubkey`.
 
 If the wallet sent a challenge, it **MUST** also reject the response if `challenge` is missing, malformed, or does not exactly match the challenge sent in the corresponding request. If the wallet did not send a challenge, no challenge is required for verification.
 
--- a/06.md
+++ b/06.md
@@ -8,20 +8,20 @@ This endpoint returns information about the mint that a wallet can show to the u
 
 ## Request
 
-Wallets **MAY** include a `challenge` query parameter in `GET /v1/info` requests. If provided, the challenge **MUST** be 32 bytes generated using a cryptographically secure random number generator, encoded as a 64-character lowercase hexadecimal string (`[0-9a-f]{64}`). Wallets using challenges **MUST** generate a fresh challenge for each request and retain it until the response has been verified.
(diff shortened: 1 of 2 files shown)


06.md
```diff
@@ -100,6 +104,30 @@ With the mint's response being of the form `GetInfoResponse`:
 - (optional) `tos_url` is the URL pointing to the Terms of Service of the mint.
 - (optional) `nuts` indicates each NUT specification that the mint supports and its settings. The settings are defined in each NUT separately.
 
+## Mint identity key
+
+The mint identity key is deterministically derived from the mint seed. Given the exact `seed` string used by the mint:
+
+1. Encode `seed` as UTF-8.
+2. Compute `secret_key = SHA256(seed)`.
```
prusnak, line 112: we should use HMAC-SHA256 construction like we do in other NUTS (e.g. 12 or 13)

06.md
```diff
@@ -100,6 +104,30 @@ With the mint's response being of the form `GetInfoResponse`:
...
+3. Interpret `secret_key` as a 32-byte big-endian integer. It **MUST** be a valid secp256k1 secret key in the range `1..n-1`, where `n` is the order of the secp256k1 curve.
```
prusnak, line 113: What happens if it does not fit? Maybe introduce a counter to the HMAC above and retry with ctr = 1, 2, 3,  etc.?

a1denvalu3: probability it doesn't fit should be roughly $\frac{1}{2^{128}}$

prusnak: I know, but we use the same construction in NUT-12, see

https://github.com/cashubtc/nuts/blob/e0a16f83d5b88a3e15418596071ac66a587e905a/12.md#L21

https://github.com/cashubtc/nuts/blob/e0a16f83d5b88a3e15418596071ac66a587e905a/12.md#L39

SatsAndSports: ... and the retry technique is also in NUT-28 on blinded keys. That specifies just one retry, for joint probability of failure of 2^(-256)

SatsAndSports:
~Concept ACK~ _PS: I've since changed my mind. As far as I know, any attacker will be able to modify/observe/forward all endpoints, and therefore I don't see what we gain with these changes to `/v1/info`_

although I'm not sure about specifying a particular derivation _"The mint identity key is deterministically derived from the mint seed."_. The wallet can't know that the derivation was correct, and therefore I don't see what's gained by forcing a particular derivation. This seems like an implementation detail, specific to each mint: 

Although I guess I might contradict myself and say: it's nice for the mint operator to be able to switch from Nutshell to CDK and have the same pubkey

Is there some existing derivation protocol, e.g. BIP-85, that we could use instead of `secret_key = SHA256(seed)`?

prusnak:
> Is there some existing derivation protocol, e.g. BIP-85, that we could use instead of `secret_key = SHA256(seed)`?

BIP-85 uses BIP-32 and it overkill since we do not need to share an xpub. HMAC-SHA256 is better and we use this construction already in other NUTs (NUTs that used BIP-32 migrated to HMAC-SHA256 too).

06.md
a1denvalu3: I think this would change the pubkey of mints that are running today, but since it's still not used for anything (that i know of) it should be OK.

06.md
```diff
@@ -100,6 +104,32 @@ With the mint's response being of the form `GetInfoResponse`:
...
+The mint **MUST** sign every `GetInfoResponse` with the identity key corresponding to `pubkey`. To construct and sign the message:
+
+1. Start with the complete `GetInfoResponse` JSON object.
+2. Remove the top-level `signature` and `time` members. The `time` member is excluded so that an otherwise unchanged response has a stable signed payload.
+3. Canonicalize the resulting JSON object using the JSON Canonicalization Scheme (JCS) defined in [RFC 8785]. Encode the canonical JSON as UTF-8.
+4. Compute the SHA-256 hash of those bytes.
+5. Sign the 32-byte hash using [BIP-340] Schnorr over secp256k1.
+6. Encode the resulting 64-byte signature as lowercase hexadecimal and return it in `GetInfoResponse.signature`.
+
+Wallets **MUST** reject the response if `pubkey` or `signature` is missing or malformed, or if BIP-340 verification fails. For BIP-340 verification, wallets use the 32-byte x-coordinate of the compressed `pubkey`.
```
robwoodgate: Wallets following this spec would reject every legacy mint info. What would rejection mean -  the wallet refuses to load the mint, warns the user, or something else?

A signature is only really meaningful if the pubkey is public knowledge. Maybe the directive should be something like:

```
If a Wallet holds an expected pubkey for the mint, it MUST reject
on verification failure or on a missing signature; otherwise it verifies
if present and records the key.
```

06.md
```diff
@@ -100,6 +104,32 @@ With the mint's response being of the form `GetInfoResponse`:
...
+Derivation **MUST** fail if no valid secret key is found after 256 attempts.
+
+The mint **MUST** return this public key in `GetInfoResponse.pubkey`.
+
+## Signature
+
+The mint **MUST** sign every `GetInfoResponse` with the identity key corresponding to `pubkey`. To construct and sign the message:
+
+1. Start with the complete `GetInfoResponse` JSON object.
+2. Remove the top-level `signature` and `time` members. The `time` member is excluded so that an otherwise unchanged response has a stable signed payload.
```
robwoodgate: Removing `time` makes mintinfo a replayable blob. A MITM or spoofer could use a month-old validly-signed info indefinitely. Signing `time` would add a freshness guarantee.

06.md
```diff
@@ -87,8 +88,11 @@ With the mint's response being of the form `GetInfoResponse`:
 }
 ```
 
+For reproducibility, the example uses the UTF-8 encoding of `NUT-06 example mint seed` as the mint seed and zero-filled auxiliary randomness for BIP-340 signing.
```
robwoodgate: Suggestion: have the worked example state the intermediate values (secret key, payload hash), not just the final signature.  Helps to isolate canonicalisation from signing issues.

06.md
robwoodgate: `time` could be redefined as the "last updated" or "signed at" time, allowing info payload to remain stable/cached for a short period. A wallet can then treat a mint info older than X time as suspicious. 

It is not needed for caching, as a consumer can simply mark the time fetched. And it's not particularly useful for consumers as "current mint time" due to network latency and caching

crodas: @robwoodgate If we drop the `time`, the only thing that could happen is that people could broadcast stale data, but I see your point.

Maybe we can include it in the payload, something like an epoch, where epochs are every minute, every hour, or something (`unixtime in seconds() / some time`), and make that number part of the payload

I have two implementations in CDK: the first [PR is for signing any payload](https://github.com/cashubtc/cdk/pull/2386) (including deriving the private key), and the [follow-up](https://github.com/cashubtc/cdk/pull/2424) would sign the mint info (the improvement we are discussing here).

a1denvalu3:
I think we should just go ahead and include `time` in the message for the signature, so that stale responses are not replayable forever. I don't think there is any real argument against it.

crodas:
# Mint public key authentication

Here is a little essay on what I believe is the issue this NUT tries to tackle and a few alternatives.

## Threats

The mint must prove its identity to a wallet. The wallet must detect an
intermediary and refuse to disclose secrets to an unauthenticated endpoint.

A mint public key can support this authentication, but a signature over mint
information does not establish the mint's identity by itself. The wallet must
also know which public key belongs to the mint it intended to contact.

## Current proposal

The mint publishes [mint
information](https://mint.minibits.cash/Bitcoin/v1/info), including its public
key:

```json
{
  "name": "Minibits mint",
  "pubkey": "03c21ef6f899f661356103552c1945579b0f36f96f9f7d2190ca0cbdab3add4af9",
  "version": "cdk-mintd/0.17.6",
  "description": "...",
  "description_long": "...",
  "contact": [
    "..."
  ],
  "nuts": {
    "...": "..."
  },
  "icon_url": "https://minibits.cash/icon-192.png",
  "motd": "migrated",
  "time": 1788293780,
  "tos_url": "https://minibits.cash/terms"
}
```

The [proposed change](https://github.com/cashubtc/nuts/pull/416) signs this
payload with the mint's private key. A wallet can then verify that the payload
was produced by whoever controls that key.

This proves control of the private key. It does not prove that the key belongs
to the mint URL selected by the wallet.

## Can a signed payload be relayed?

Yes. An attacker can copy a valid signed response from Mint A and serve it from
another endpoint. If the wallet has no authenticated association between Mint
A's URL and public key, the signature still verifies.

The problem has two parts:

1. Freshness: Is this a current response rather than an old signed response?
2. Identity binding: Does this public key belong to the mint the wallet intended
   to contact?

A timestamp or nonce addresses freshness. Neither one establishes identity
binding.

## Expiration

The current signed mint information has no explicit expiration time. One
proposal signs the mint's server time and lets each wallet decide when the
response is stale. This creates different behavior across wallet
implementations.

The signed payload should include `valid_until`. The mint then defines the
lifetime of its assertion, and wallets can apply one clear validation rule:

```text
issued_at <= wallet_time <= valid_until
```

Wallets should allow a small, specified clock skew. They should reject a
payload with an excessive validity period if the specification defines a
maximum.

Expiration limits replay duration. It does not prevent an attacker from
relaying a response while it remains valid.

## Bind the signature to the mint URL

[NUT-06](https://cashubtc.github.io/nuts/06/) defines an optional `urls` field:

```markdown
- (optional) `urls` is the list of endpoint URLs where the mint is reachable.
```

For authentication, the signed payload should include a required canonical
mint URL, or a required list of permitted URLs. The wallet must verify that the
URL it contacted matches one of the signed URLs after applying specified URL
normalization rules.

This prevents a valid response for one mint from being presented as a valid
response for a different mint URL. It does not solve initial key discovery if
the attacker can replace the entire response and substitute another key.

Signing IP addresses would add operational problems without fixing the trust
model. Mint addresses may change, use a CDN, resolve differently by region, or
support IPv4 and IPv6. DNS poisoning is normally handled by authenticated HTTPS
and certificate validation. DNSSEC or a separately authenticated discovery
record could add another binding mechanism.

## Interactive proof

The wallet can send a challenge containing a random nonce and the canonical
mint URL it intended to contact. The mint signs a domain-separated response:

```text
challenge = {
  context: "cashu-mint-auth-v1",
  mint_url: canonical_url,
  nonce: random_32_bytes,
  issued_at: unix_time,
  valid_until: unix_time
}
```

The wallet verifies all of the following:

1. The signature is valid for the expected mint public key.
2. `context` matches the authentication protocol and version.
3. `mint_url` matches the URL selected by the wallet.
4. `nonce` matches the wallet's challenge and has not been used before.
5. The response is within its validity interval.

The nonce prevents replay of an earlier challenge response. The URL prevents a
response for one mint endpoint from being accepted for another endpoint.

The interactive proof still depends on the phrase "expected mint public key."
If the wallet learns the public key from the same unauthenticated connection,
an active attacker can provide its own public key and sign its own challenge
response. The protocol must therefore define how the wallet obtains or confirms
the authentic key.

## Required trust anchor

At least one authenticated binding must exist between the mint identifier and
its public key. Possible trust anchors include:

* The public key is pinned when the user adds the mint.
* The mint URL contains the public key or its fingerprint.
* A trusted directory or authenticated out-of-band channel provides the key.


## Proposed protocol requirements

The specification should define:

* A canonical mint URL and exact URL normalization rules.
* A domain-separated signature format.
* `issued_at` and `valid_until`, plus clock-skew and maximum-lifetime rules.
* An optional interactive nonce challenge for proof of current key control.
* An authenticated method for initial public-key discovery.
* A rule that wallets send secrets encrypted with the mint's public key.

A signed mint information document is useful for integrity, caching, and
authenticated metadata once the public key is trusted. Preventing MITM attacks
requires the signature to be connected to a trust anchor for the mint the
wallet intended to use.

SatsAndSports:
Thanks @crodas ! I think one big problem isn't discussed in that essay:

Even with the full challenge response on `/v1/info`, and also assuming that the wallet knows that it has the right public key to verify against, there could be a MITM who is observing every connection and is modifying the non-/v1/info requests and responses

The wallet wants to know that all connections - not just `/v1/info` - are connections to the real mint, such that the requests and responses are unaltered and unobservable by any attacker

crodas:
@SatsAndSports, that would require a protocol change. A change in which the entire payload is encrypted and the decryption key is encrypted with the mint's public key. 

If not the entire payload, perhaps all the secrets.

robwoodgate:
Thanks @crodas. To break this down a little more, we are dealing with three concerns:
- **integrity:** "Did the holder of the mint pubkey produce this document, and is the info fresh/current?"
- **identity:** "How do I learn the correct mint pubkey".
- **transport:** "How do I know I'm not being MITM'd?"

This PR is only intended to fully satisfy the first (integrity).

That doesn't mean this PR is deficient - because the solution to each of these problems is different.

## Integrity (this PR)

For this PR, the only debatable (and related) variables are:
- CACHING vs ON-DEMAND signing
- how long is considered 'fresh'

Mints currently (afaik) do not cache the mintinfo, and signing the JSON is microseconds of work, so freshness can be "this instant" now that `time` is signed over.

Wallet caching of this info can then be either:
- left to wallet discretion (as now)
- mandated in the spec (`Wallets MUST refresh mintinfo where time is older than X seconds`)
- suggested by the mint (the `valid_until` field), which wallets can choose to ignore if they wish.

My recommendation would be to include `valid_until` in any case, as this allows a mint to signal a shorter validity time around pubkey or url changes (migration etc).

Whichever we pick, staleness should be a soft fail: a stale-but-valid signature downgrades the info to the same trust level as *unsigned* info (the state every legacy mint is in today). Only a bad signature, or a pinned-pubkey mismatch, is a hard fail. This means a wallet with a wrong clock degrades gracefully instead of rejecting every mint.

On-demand signing also allows a challenge-response option for wallets that do not trust their own clock (they simply ignore `time` and rely on their challenge instead).

Example:
```
GET /v1/info?challenge=<32 random bytes as hex>
Mint then confirms the challenge IS 64 hex safe characters
and includes the challenge in the response, signed over.
```

My recommendation would be to ALWAYS sign over `time` (stamps the signature) and OPTIONALLY include challenge response.

For the signature itself, we should use the BIP-340 tagged hash construction (as NUT-10 nutroot #421 already does, eg `Cashu_TransactionInput`) with tag `Cashu_MintInfo`. This key will end up signing several kinds of statement, and domain separation is the one thing we get right up front.

One scoping note: the identity key is HMAC-derived from the mint seed, so it shares the seed's fate - if the seed leaks, the mint is fully compromised anyway and no "rotation" statement signed by the old key means anything. So key rotation is deliberately out of scope: a mint that replaces its identity key is, to wallets, a new mint.

## Identity

This is the perennial key distribution problem, and this PR can **HELP** (partially) solve it, by using @crodas suggestion to mandate the urls field and ensure it includes the primary one, not just secondary urls.

We should perhaps reframe this along the lines CDK is already moving (in cdk#2297)... the pubkey is the IDENTITY, and the url(s) are the LOCATORS.

So signing the urls into mintinfo as mandatory HELPs - the holder of the pubkey signs the urls into the info, attesting that they are the correct locators. The freshness guarantee of this PR (binding `time` and optionally `valid_until`) lets wallets know how long that link is valid for.

The rest of the identity problem can be solved with separate measures, such as:
- TOFU (Trust On First Use): Wallet pins pubkey on first contact, stores it with the mint record, hard-warns on change. My review comment above captures the MUST/SHOULD split for this.
- Cross attestation: The mint info signature covers the `contact` field, which can include the Nostr key. The mint's NIP-87 announcement will also carry and be signed by the same pubkey. That gives a wallet two ways to check.
- Tokens could be modified to carry mint pubkey alongside the url, which may make it easier for a wallet to TOFU a mint it has never seen before.

## Transport

We should distinguish between a valid MITM (such as a relay/proxy) and a MITM attacker. Both can forward the mintinfo verbatim, and neither gains anything by doing that alone.

The issue is when the MITM starts substituting outputs in a swap, swapping invoices in melt quotes, or lying about state.

None of that can be avoided or defended by this PR, regardless of whether it is challenge-response or just time signed.

The solutions here include:
- transport channel binding: eg TLS or .onion (all mints use one of these??)
- protocol binding: mint signs request-response, eg:

```
TaggedHash("Cashu_MintResponse", sha256(request_bytes_the_mint_received) || sha256(response_bytes))
```

Note: nutroot transcripts in #421 already bind the *write* side for v3 keysets (inputs sign over outputs), so protocol binding is only needed for what the mint *says*: quotes, states, keys.

06.md
```diff
@@ -87,19 +88,48 @@ With the mint's response being of the form `GetInfoResponse`:
...
+The mint **MUST** return this public key in `GetInfoResponse.pubkey`.
+
+## Signature
+
+The mint **MUST** sign every `GetInfoResponse` with the identity key corresponding to `pubkey`. To construct and sign the message:
+
+1. Start with the complete `GetInfoResponse` JSON object.
+2. Remove the top-level `signature` member. The `time` member remains in the object and is therefore included in the signed payload.
+3. Canonicalize the resulting JSON object using the JSON Canonicalization Scheme (JCS) defined in [RFC 8785]. Encode the canonical JSON as UTF-8.
+4. Compute the SHA-256 hash of those bytes.
```
robwoodgate: For consistency with #421 and domain separation, I wonder if we should perhaps use something like:

```suggestion
4. Compute the message hash as the [BIP-340] tagged hash of those bytes with tag `Cashu_MintInfo`, i.e. SHA256(SHA256(tag) || SHA256(tag) || bytes).
```

Would change the vectors you calculated though (sorry)

SatsAndSports:
> @SatsAndSports, that would require a protocol change. 

For now, I'm trying to avoid saying any solution would look like. I'm trying to clarify the problem

SatsAndSports:
Forgetting about solutions, what problem(s) are we interested in?

What _tools_ are available to any attacker?

They have enough control over DNS, and/or over the the Certificate Authorities, that they can redirect traffic. _If they don't have this power, then there is no problem to be solved 😀._

The wallet attempts to connect to `https://cashumint.cash` but instead the wallet gets connected to the attacker's machine. The attacker can forward calls to the real mint, and can therefore monitor everything. The attacker can also modify the requests and responses to steal tokens

So this (the paragraph just above) is the problem we're trying to solve?

robwoodgate:
> The wallet attempts to connect to `https://cashumint.cash` but instead the wallet gets connected to the attacker's machine. The attacker can forward calls to the real mint, and can therefore monitor everything. The attacker can also modify the requests and responses to steal tokens
> 
> So this (the paragraph just above) is the problem we're trying to solve?

Not in this PR, no. The para above is the transport concern - modifying requests and responses.

A signature on mint info can never resolve this on its own. 

This PR is about nailing the Integrity aspect, which in turn helps with the Identity aspect, but does not attempt to fix transport or all aspects of identity because a single signature cannot do all three jobs.

SatsAndSports:
> Not in this PR, no.

I think I agree that this PR doesn't solve that problem

Maybe these comments in this PR aren't the right place to discuss the issue. By discussing them here we are (implicitly) discussing _"is this solution the solution to the (undefined?) problem?"_

I think that's the wrong question. I want to first clarify the problem

I think the real question is _"What problem(s) are of interest? i.e. what are the tools available to an attacker, and how might they cause problems to honest participants? How do we clarify all that without jumping into particular proposed solutions?"_

crodas:
> I think the real question is "What problem(s) are of interest? i.e. what are the tools available to an attacker, and how might they cause problems to honest participants? How do we clarify all that without jumping into particular proposed solutions?"

@SatsAndSports  We want to use the public keys to identify wallets instead of their `mint_url`, especially since the wallet <> mint communication could be done over other transports.

A good step toward meaningful public keys is to sign the mint info and include the'mint_url' in it. Then, if the mint changes its domain for whatever reason, it can advertise it, and wallets can trust it because of the signed mint info.

06.md
robwoodgate: the vectors for this change would be:

```
tag                "Cashu_MintInfo_v1" = 43617368755f4d696e74496e666f5f7631
sha256(tag)        916a34ebf6f2244d64490e8eb2b5e7af19cdc45aa6176160186fc6543aa20d9b
message = sha256(sha256(tag) || sha256(tag) || payload)
                   5680a1b8da9f3662ca42482ca17887c665a22dab08f33320d7d632d14f534a96
aux_rand           0000000000000000000000000000000000000000000000000000000000000000
signature          bdd3706485708eabbdece3abc6e50074fd4f09dd8751a2d596e3ce4436d46b14c4dd2f4cf4301c18a3fe0cbbee3964a2e883fd608b27bc346183178d998bee45
```

06.md
```diff
@@ -6,20 +6,28 @@
 
 This endpoint returns information about the mint that a wallet can show to the user and use to make decisions on how to interact with the mint.
 
+## Request
+
+Wallets **MUST** include exactly one `challenge` query parameter in each `GET /v1/info` request. The challenge **MUST** be 32 bytes generated using a cryptographically secure random number generator, encoded as a 64-character lowercase hexadecimal string (`[0-9a-f]{64}`). Wallets **MUST** generate a fresh challenge for each request and retain it until the response has been verified.
+
+The mint **MUST** reject requests with a missing, repeated, or malformed `challenge` parameter with HTTP status code `400`, using the error response format in [NUT-00][00]. For a valid request, the mint **MUST** copy the challenge unchanged into `GetInfoResponse.challenge` and include it in the signed payload. Wallets **MUST** verify that the returned challenge exactly matches the challenge sent in the corresponding request.
```
robwoodgate: As worded, it creates a hard break for old wallets who don't know to pass a challenge. `repeated` could also be read as the mint must store challenges and ensure they are unique, which is not the intention.
  
Perhaps:

```suggestion
If the request has no `challenge` parameter, the mint **MUST** respond without a `signature`. The mint **MUST** reject a request with a malformed `challenge` or more than one `challenge` parameter with HTTP status code `400`, using the error response format in [NUT-00][00]. For a valid request, the mint **MUST** copy the challenge unchanged into `GetInfoResponse.challenge` and include it in the signed payload. Wallets **MUST** verify that the returned challenge exactly matches the challenge sent in the corresponding request.
```

Also, no specific error code is defined - should we have one?

06.md
```diff
@@ -87,25 +95,60 @@ With the mint's response being of the form `GetInfoResponse`:
...
+The mint **MUST** sign every `GetInfoResponse` with the identity key corresponding to `pubkey`. To construct and sign the message:
+
+1. Start with the complete `GetInfoResponse` JSON object.
+2. Remove the top-level `signature` member. The `challenge` and `time` members remain in the object and are therefore included in the signed payload.
+3. Canonicalize the resulting JSON object using the JSON Canonicalization Scheme (JCS) defined in [RFC 8785]. Encode the canonical JSON as UTF-8.
+4. Compute the SHA-256 hash of those bytes.
+5. Sign the 32-byte hash using [BIP-340] Schnorr over secp256k1.
+6. Encode the resulting 64-byte signature as lowercase hexadecimal and return it in `GetInfoResponse.signature`.
+
+Wallets **MUST** reject the response if `pubkey`, `signature`, `challenge`, or `time` is missing or malformed, if `challenge` does not exactly match the challenge sent in the corresponding request, if BIP-340 verification fails, or if `time` differs from the wallet's current Unix time by more than 3600 seconds. A difference of exactly 3600 seconds is valid. For BIP-340 verification, wallets use the 32-byte x-coordinate of the compressed `pubkey`.
```
robwoodgate: As above: avoids a hard break for a compliant wallet talking to a mint that hasn't upgraded yet.

```suggestion
If the response contains `signature`, wallets **MUST** reject it if `pubkey`, `challenge`, or `time` is missing or malformed, if `challenge` does not exactly match the challenge sent in the corresponding request, if BIP-340 verification fails, or if `time` differs from the wallet's current Unix time by more than 3600 seconds. A difference of exactly 3600 seconds is valid. If the response contains no `signature`, the wallet **SHOULD** treat the mint info as unverified. For BIP-340 verification, wallets use the 32-byte x-coordinate of the compressed `pubkey`.
```

robwoodgate (commented):
Couple of wording nits, but otherwise ACK the challenge approach.

I'd still prefer the signature be over a tagged hash rather than plain SHA256. The mint identity key will sign other things (#437 already has it sign `tagged_hash("Cashu_MintResponse_v1", ...)`), and tagging keeps each use of the key in its own domain. It's one line per library now and a format change later.

06.md
```diff
@@ -87,25 +95,60 @@ With the mint's response being of the form `GetInfoResponse`:
...
+
+## Signature
+
+The mint **MUST** sign every `GetInfoResponse` with the identity key corresponding to `pubkey`. To construct and sign the message:
+
+1. Start with the complete `GetInfoResponse` JSON object.
+2. Remove the top-level `signature` member. The `challenge` and `time` members remain in the object and are therefore included in the signed payload.
+3. Canonicalize the resulting JSON object using the JSON Canonicalization Scheme (JCS) defined in [RFC 8785]. Encode the canonical JSON as UTF-8.
+4. Compute the SHA-256 hash of those bytes.
+5. Sign the 32-byte hash using [BIP-340] Schnorr over secp256k1.
```
robwoodgate: For consistency with #421 and domain separation, I wonder if we should perhaps use something like:

```suggestion
4. Compute `tagged_hash("Cashu_MintInfo_v1", bytes)`, where `tagged_hash` is the [BIP-340] construction `SHA256(SHA256(tag) || SHA256(tag) || message)` over the tag's UTF-8 bytes. `SHA256("Cashu_MintInfo_v1")` = `916a34ebf6f2244d64490e8eb2b5e7af19cdc45aa6176160186fc6543aa20d9b`.
5. Sign the resulting 32-byte hash as the message using [BIP-340] Schnorr over secp256k1.
```

Would change the vectors you calculated though (sorry)

robwoodgate: Vectors for this change, computed against the current head (75aced6b), same seed, key and zero `aux_rand` as the PR:

```
tag            "Cashu_MintInfo_v1" = 43617368755f4d696e74496e666f5f7631
sha256(tag)    916a34ebf6f2244d64490e8eb2b5e7af19cdc45aa6176160186fc6543aa20d9b
message        sha256(sha256(tag) || sha256(tag) || canonical_payload)

Full 06.md example (signature removed, challenge and time kept):
message        3c08960dfc7c5f62e489271ed44cf82240b3fe48a13968179683461444855db6
signature      e91f667871ff922f7430aee729bc0252ff893ec2b05ef2b0f98376f7f5119b152029b363f89c35ffa2d8e3395c889c5caa2484c8b716cc18881dad7b6ef9a739

Minimal signed response in tests/06-tests.md:
message        374353ff76b340d0010379009839f8b29152808ec393409cfb79abf4b19e65bc
signature      e22072aff2d70f9cb210ae5bb9a1d3fe44349cbde03fcd26a73e0c10e30bb3b801046db0b096d02f25c8cf76dd24ffd6c1906c3cbc98193c7d2d8d6978c95f0a
```

The canonical payloads and their plain SHA-256 (`5c318fae…`, `a586404c…`) are unchanged from the PR; only the signed message and signature move.

06.md
a1denvalu3: There is no break for old wallets. The mint simply responds with the challenge in the signature if the wallet requests it.

robwoodgate: Your wording says mints must reject a request with missing challenge parameter which is a hard break for legacy wallets

robwoodgate: You probably mean `challenge=` but the ambiguity in the wording reads as "no challenge at all"

a1denvalu3:
PR #437 expands the scope of this change to any payload returned by the mint. We should probably discuss it.

callebtc:
Could we rename `challenge` to `request_nonce`? "Challenge" is valid in the challenge-response sense, but this value is specifically a wallet-generated freshness nonce that binds the signed response to the corresponding request and prevents replay of earlier responses. `request_nonce` makes that role explicit and avoids confusion with either the hash-derived Schnorr challenge or the secret Schnorr signing nonce.

crodas:
I think the whole `request_nonce` is a very cheap attack vector. Can the nonces be deterministic? Like [TOTP](https://en.wikipedia.org/wiki/Time-based_one-time_password#Algorithm) values, where time is divided into 30-second blocks (or every minute) so that the signatory would sign at most once every minute. That, plus including the hostnames (and perhaps the mint IPs if we want to be extra paranoid), is more than enough.

The attack I want to prevent is multiple requests with different nonces at all, to exhaust the mint's CPU.

06.md
```diff
@@ -87,25 +99,70 @@ With the mint's response being of the form `GetInfoResponse`:
...
+3. Canonicalize the resulting JSON object using the JSON Canonicalization Scheme (JCS) defined in [RFC 8785]. Encode the canonical JSON as UTF-8 to obtain `bytes`.
+4. Compute `message_hash = tagged_hash("Cashu_MintInfo_v1", bytes)`.
+5. Sign the 32-byte `message_hash` using [BIP-340] Schnorr over secp256k1.
+6. Encode the resulting 64-byte signature as lowercase hexadecimal and return it in `GetInfoResponse.signature`.
+
+The `tagged_hash` function uses the [BIP-340] construction `SHA256(SHA256(tag) || SHA256(tag) || message)`, where `tag` is encoded as UTF-8, each inner SHA-256 result is its raw 32-byte digest, and `||` denotes byte concatenation. For this tag, `SHA256("Cashu_MintInfo_v1") = 916a34ebf6f2244d64490e8eb2b5e7af19cdc45aa6176160186fc6543aa20d9b`.
+
+Wallets **MUST** reject the response if `pubkey`, `signature`, or `time` is missing or malformed, if BIP-340 verification fails, or if `time` differs from the wallet's current Unix time by more than 3600 seconds. A difference of exactly 3600 seconds is valid. For BIP-340 verification, wallets recompute `message_hash` as specified above and use the 32-byte x-coordinate of the compressed `pubkey`.
```
robwoodgate, line 145: For now, at least, we must leave it optional, as a spec compliant wallet will otherwise reject all legacy mints responses (which will still have optional `pubkey` and `time`, and no `signature`).

```suggestion
Wallets **MAY** reject the response if `pubkey`, `signature`, or `time` is missing or malformed, if BIP-340 verification fails, or if `time` differs from the wallet's current Unix time by more than 3600 seconds. A difference of exactly 3600 seconds is valid. For BIP-340 verification, wallets recompute `message_hash` as specified above and use the 32-byte x-coordinate of the compressed `pubkey`.
```

robwoodgate (commented):
This is pretty close now, I think. One suggestion, and a couple of non-blocking observations  below:

- Multiple `request_nonce` query params is unspecified (eg: `request_nonce=A&request_nonce=B`). Should return `HTTP 400` error.
- No specific NUT-00 error code is assigned for a bad nonce, only "the error response format".
- JCS serialises numbers as doubles, so an amount above `2^53` in the payload would canonicalise differently across languages. Theoretical more than problematic considering the current payload, but still a potential bear trap.

Also, PR description body is stale.

robwoodgate:
> I think the whole `request_nonce` is a very cheap attack vector. Can the nonces be deterministic? Like [TOTP](https://en.wikipedia.org/wiki/Time-based_one-time_password#Algorithm) values, where time is divided into 30-second blocks (or every minute) so that the signatory would sign at most once every minute. That, plus including the hostnames (and perhaps the mint IPs if we want to be extra paranoid), is more than enough.
> 
> The attack I want to prevent is multiple requests with different nonces at all, to exhaust the mint's CPU.

Interesting idea! Though surely the mint rate limiter covers this?

In any case, the processing of a JCS and tagged hash is already pretty cheap, I believe, and time is supposed to be live mint time so would have to be exempt from any caching.

PR #419, opened 2026-08-07
Mint Quote Lookup by Public Key spec fixes

vnprc:
Two amendments to #341, one commit each — commit messages carry the full reasoning.
 
Drop `{method}` from the path. The endpoint answers "what does this key own," and a NUT-20 key locks quotes of any payment method — so the segment forces one request per method and buys nothing. The signature doesn't cover it, and every `MintQuoteResponse` already carries a method field (NUT-04), so a mixed-method response is unambiguous without it.
 
Bound the pubkey count. The mint must verify every signature before it can tell whether the caller is entitled to anything, so request length decides how much work an anonymous caller can demand. Adds a MUST to limit it, error `11017` on overflow, and an optional `max_pubkeys` in the NUT-06 settings — the same shape NUT-29 already uses for `max_batch_size`.

### DIFF
--- a/xx.md
+++ b/xx.md
@@ -38,6 +38,8 @@ For each `pubkey`, the corresponding `pubkey_signatures` entry signs the SHA-256
 
 The mint **MUST** reject the request unless every signature is valid.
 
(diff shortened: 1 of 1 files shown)


thesimplekid:
https://github.com/cashubtc/nuts/pull/341#discussion_r3644500473

Can you do it as review comments

PR #302, opened 2025-10-24
SIG_ALL: extent aggregated message to `secret` + `C` for inputs and `amount` + `B_` for outputs

callebtc:
The aggregation message now contains `secret` + `C` for each input and the keyset `id` + `amount` + `B_` of each output in the transaction.

todo:
- [x] fix test vectors

- [ ] nutshell: https://github.com/cashubtc/nutshell/pull/811
- [x] cdk: https://github.com/cashubtc/cdk/pull/1212
- [ ] nutmix:
- [ ] cashu-ts: https://github.com/cashubtc/cashu-ts/issues/386

PR #421, opened 2026-08-11
NUT-10: Nutroot secrets (v3 keysets)

robwoodgate:
# Moved to #443 

# Implementations

These are Proof of concept implementations.

- [ ] Cashu-TS https://github.com/cashubtc/cashu-ts/pull/950
- [ ] Nutshell https://github.com/cashubtc/nutshell/pull/999 (via cashubtc/nutshell/pull/1111)
- [ ] CDK https://github.com/cashubtc/cdk/issues/2433

## Summary

Formalizes nutroot secrets, the secret family for v3 (BLS) keysets, as version-scoped additions to the existing NUTs rather than a new NUT.

The core expands **NUT-10**: secrets are compressed secp256k1 points, bare or tweaked BIP341-style over a tree of declarative condition leaves (`threshold`, `after`, `hashlock`); a shared TLV grammar; the transaction transcript that every v3 input signs; key path and script path witnesses; spend info for transfer; and receive-time verification. Full design rationale lives in the [discussion draft](https://hackmd.io/jdOrJ1COQfOHOkguUF8YOg) and is deliberately kept out of the spec text.

## Per-NUT deltas:

- **NUT-00**: v3 secret bytes for `hash_to_curve_G1`; secret uniqueness stated on the `Proof` model; `si` shown in the V4 token format.
- **NUT-03/05**: every v3 input carries a witness over the transaction transcript; transactions mixing keyset versions verify per input.
- **NUT-04**: v3 mint quotes are always locked and enter the transaction as inputs, signing via the existing `signature` field. **NUT-29**: a batch is one transaction, one digest.
- **NUT-07**: a check-state entry returning a version `02` witness also returns the transaction digest it signed, as a sibling field.
- **NUT-11/14**: scoped to pre-v3 keysets.
- **NUT-13**: the v3 secret branch derives a private key (rejection sampling against `SECP256K1_N`); per-purpose derivation paths are superseded for v3.
- **NUT-18/26**: the `nutroot` payment request option (`k`, `l`, `b`), carried in creqB under new TLV tag `0x0b`.
- **NUT-20**: the pre-v3 quote signature scheme is version-scoped; v3 quote lock keys derive beside it (NUT-13 type `0x04` with an empty keyset id, so a quote key survives keyset rotation and seed-only recovery).
- **NUT-22**: version `02` BATs are point secrets whose required witness signs a request transcript (method, target, body hash: the `0x05` container) under the ordinary transaction digest rules. Spec only; implementation deferred.
- **NUT-28**: the blinding core applied one layer down: positional sender slots, receivers matching by value, verbatim NUMS bases.

## How this differs from BIP341

The scheme is called **nutroot**, not taproot, because what is borrowed is the commitment structure and little else: a public key that secretly commits to a merkle tree of alternative spending conditions, where a key-path spend is byte-identical to an unconditional one. Almost none of Bitcoin's validation rules come with it, so do not port assumptions:

- **The secret is a 33-byte compressed point, not an x-only output key.** Signatures are still BIP340, verified against the secret's x-coordinate.
- **The tweak reduces mod `n` and is never rejected.** BIP341 rejects a tweak digest at or above the curve order; here it reduces.
- **The tweak preimage takes the full 33-byte internal key.** BIP341 hashes a 32-byte x-only key there, so ported code derives a different tweak and an unspendable proof.
- **Leaves are declarative TLV records, not script.** Leaf version `0x00` selects a fixed vocabulary (`threshold`, `after`, `hashlock`). There is no interpreter, no opcodes and no stack. An unknown field of either parity rejects: odd type numbers are reserved for a possible future ignorable class, so neither Lightning's it's-OK-to-be-odd rule nor an annex-style free data channel applies here.
- **The tree shape is fixed by the leaf count, not chosen per leaf.** BIP341 lets the constructor set each leaf's depth, burying rare scripts to shorten the likely spending path. Here the fold decides the shape, so a payer can reproduce a payee's requested tree without being told it.
- **The merkle fold is normative and sorts.** Leaf hashes sort ascending, then pair per level with an unpaired hash promoted unchanged, so the root commits the leaf set rather than any transmitted order and no constructor-chosen shape travels in a control block. Depth is capped at 8.
- **The tagged-hash domains are `Cashu_NutrootLeaf`, `Cashu_NutrootBranch` and `Cashu_NutrootTweak`.** The hash construction is BIP341's, the tags are not, so no derived value is interchangeable with Bitcoin's.
- **The NUMS point is never used verbatim.** A script-only internal key is `H + u*G` with a fresh `u` per proof, disclosed so a holder can verify no key path exists.
- **There is no sighash, annex or input index.** The single digest a witness signs is the NUT-10 transaction transcript, which has no Bitcoin analogue: its inputs are proofs and paid mint quotes, its outputs blinded messages and melt quotes.

## Test vectors

New `tests/10-tests.md` (leaf forms, tree folds, worked derivations, transaction transcripts with signatures, V4 tokens with spend info) plus additions to the NUT-13/18/26/28 vectors. Every value is produced by one of two independent implementations (cashu-ts and nutshell) that pass a shared integration suite against each other, and the tree, tweak, and transcript values were re-derived independently during review.

## Stacking

Builds on #371 (BLS12-381 v3 keyset protocol). The PR is based on `bls-protocol` (#371's content with current main merged in), so the diff here is the nutroot change alone. Once #371 merges to main, this retargets to main and the stack base branch can be deleted.


10.md
```diff
@@ -2,19 +2,252 @@
...
+msg    = "Cashu_Transaction_v1" || TLV transcript
+digest = SHA256(msg)
+```
+
+The domain tag is the ASCII bytes, plainly prefixed. The transcript is one TLV container record per element:
+
+| Container              | Type   | Fields (nested TLV)                                |
+| ---------------------- | ------ | -------------------------------------------------- |
+| proof input            | `0x01` | `01` amount, `02` keyset id, `03` secret, `04` `C` |
+| mint quote input       | `0x02` | `01` amount, `02` quote id                         |
```
a1denvalu3, line 148: **Which `amount` does this bind for a partially-redeemed quote?** NUT-04 explicitly permits drawing `amount_paid − amount_issued` across multiple mint calls (04.md:83), and NUT-29 admits the bolt12 "MUST NOT exceed" case — so a quote is not a one-shot input, yet the transcript gives it a single amount field.

- If this is the quote's paid/face amount, a partial draw signs an input that overstates what the transaction consumes, and "inputs cover outputs" no longer describes the accounting.
- If it's the drawn amount, batched mints can't compute it (outputs are a single unpartitioned set per NUT-29).

Either choice changes the digest, so the semantics need pinning in the text — otherwise two implementations produce different transcripts for the same request and every signature fails cross-implementation. Not a theft vector (the mint enforces accounting anyway), but exactly the byte-level ambiguity the framed transcript exists to eliminate.

10.md
```diff
@@ -2,19 +2,252 @@
...
+- Wallets **SHOULD** send receiver-keyed (`E`, [NUT-28][28]) when they hold a static key for the receiver; the bearer `k` is the offline fallback. A derived private key **MUST NOT** be re-gifted as a bearer `k` (the original sender knows the blinding tweak and could recover the receiver's static key): sweep, then send.
+
+A locked proof's owner needs the tree even for key-path spending (the tweak needs the root), so spend info is fund-critical wallet data: storage, token serialization, backups. A proof with no spend info is the self-owned, seed-recoverable case ([NUT-09][09]/[NUT-13][13]).
+
+#### Receive-time verification
+
+On receiving a v3 proof, the wallet **MUST** run this cascade and reject on fall-through:
+
+1. `k` present, no tree: accept iff `k*G` equals the secret (bare), or the empty tweak of `k*G` equals it (aggregated form).
+2. A disclosed tree plus a key source (`k*G`, an explicit `K`, or the key derived from `E`): every leaf **MUST** parse (an unknown version, type, or constraint field rejects), and the recomputed `K + t*G` **MUST** equal the secret.
```
a1denvalu3: **The disjunctive "a key source" leaves precedence undefined, and the choice is observable.** Consider `si = {E, K, tree}` where the explicit `K` satisfies `K + t*G == secret` but is *not* the key the receiver derives from `E` (sender blinded a third party's key, or a random one). An implementation using explicit `K` as the rule-2 source **accepts**; one using the `E`-derived key **rejects**. Same bytes, opposite outcomes across wallets.

The accepting wallet is also silently harmed: its `p + r₀ + t` doesn't match the committed key path, so it can't key-path spend and only discovers that when a sweep signature fails.

Suggest: when `E` is present and the receiver can derive a key from it, the derived key is authoritative for this check and a mismatched explicit `K` MUST reject; explicit `K` serves third-party leaf signers, not the cascade, except in the `{K, tree}` shape.

10.md
```diff
@@ -2,19 +2,252 @@
...
+3. `K` alone, no tree: accept iff its empty tweak equals the secret (an aggregated key).
```
a1denvalu3: **Rule 3 is a trap as written.** Accepting `si = {K}` alone hands the receiver no key material, and "aggregated key" is unverifiable from the data: a sender can set `K = s·G` from a single scalar it holds and retain the key-path scalar `s + tagged_hash("Cashu_TapTweak", K)` indefinitely. A receiver that isn't a cosigner of `K` then holds a proof it can *never* spend and the sender can *always* spend — strictly worse than the bearer-`k` concealment case (§3.8's domination argument), which at least leaves the receiver a working key.

Since the cascade is presented as the receive-time acceptance test, this needs explicit text: rule 3 is only meaningful when the receiver holds a share of `K` (or otherwise knows it can produce the aggregate signature); a `si = {K}` proof from a party the wallet didn't cosign with MUST NOT be treated as received value.

10.md
```diff
@@ -2,19 +2,252 @@
...
+4. Anything else, including a tree without a key source: reject.
```
a1denvalu3: Read literally, a proof with **no spend info at all** falls through to this rule and is rejected — while the spend-info section above says that's "the self-owned, seed-recoverable case." The seed path should be stated as part of the cascade: if no spend info is present, accept iff the secret derives from the receiver's own seed (NUT-13); otherwise reject. Otherwise implementers will either reject si-less tokens outright or improvise the rule.

a1denvalu3 (commented):
**Independent audit.** I read the full diff, cross-checked the construction against BIP341, and re-derived every test vector with independent code (tree fold, tweaks, BIP-340 witnesses, NUT-13 derivation, `hash_to_curve_G1`, token encodings).

**All vectors check out byte-for-byte, and the core design holds up**: a bearer key hiding a tree gives the sender no power they didn't already have; the re-gift rule is correct; value-matching survives tree reordering; the x-coordinate rule closes the BIP-340 parity loophole; per-input transaction signatures make v3 requests non-replayable — a real upgrade over pre-v3.

Findings are inline. None needs a redesign. The four to fix before freeze:

1. **10.md:137** — mint-quote input `amount` is undefined for partially minted quotes.
2. **10.md:226** — rule 2 doesn't say which key wins when `E` and `K` disagree.
3. **10.md:227** — rule 3 lets a wallet accept a proof it can't spend and the sender still controls.
4. **18.md:132** — the tree check must be exact match, not "contains".

(AI-assisted review; all crypto claims machine-verified.)

18.md
```diff
@@ -102,6 +104,36 @@ The `nut10` field specifies the payee's requested locking condition as a `NUT10O
...
+
+The payment flow:
+
+1. The payer picks a fresh ephemeral keypair `(e, E)` per output and blinds `k` at slot 0 into the internal key `K` ([NUT-28][28]); a NUMS `k` is used verbatim instead.
+2. The payer reproduces the leaves of `l` **byte for byte**, replacing each key tagged in `b` with its slot-blinded form (slots `1..`, transmitted order). A payer that cannot reproduce a leaf exactly **MUST** refuse to pay: altered leaf bytes change the secret, and the payee cannot spend what arrives.
+3. The payer computes the tree root and the tweaked secret `P = K + t*G` ([NUT-10](10.md#taproot-secrets-v3-keysets)), swaps to an output with secret `P`, and sends the proof with spend info `E`, the transmitted tree, and `K`.
+4. The payee derives the internal key from `E` and her static key (or checks a NUMS `K` byte-for-byte), runs the [receive-time cascade](10.md#receive-time-verification), and **MUST** verify the disclosed tree honors `l` and `b`, matching blinded keys by value ([NUT-28][28]).
```
a1denvalu3: **"Honors `l` and `b`" should be pinned to exact set equality modulo `b`-substitutions.** If a payee implements this as "contains the requested leaves," a payer can append an extra leaf — e.g. a short-horizon `after` naming the payer — and claw the payment back after the payee accepted it as satisfying the request. The vectors imply byte-for-byte reproduction of `l` with blinded substitutions; making "no more and no fewer leaves than `l`, with exactly the `b` keys blinded" explicit removes the ambiguous middle that a clawback leaf would slip through.

04.md
```diff
@@ -127,6 +127,16 @@ The mint then responds with:
 
 where `signatures` is an array of blind signatures on the outputs.
 
+## Taproot transactions (v3 keysets)
+
+Minting onto a keyset with version byte `02` or later follows [NUT-10's transaction model](10.md#the-transaction-transcript): the paid quote is a transaction **input**, and inputs sign.
+
+- The quote **MUST** include the `pubkey` locking field. Mints **MUST** reject minting an unlocked quote onto a v3 keyset: an unlocked quote has no key to sign with.
+- The quote's witness travels in the mint request's `signature` field. It signs the [transaction digest](10.md#the-transaction-transcript) over the quote input and all `outputs`, replacing [NUT-20][20]'s message for v3 keysets. The field carries either a BIP-340 signature as 128 hex characters (key path) or a serialized [script-path witness](10.md#script-path) JSON string.
```
a1denvalu3: **Which message is signed when one mint request's `outputs` span keyset versions?** The transcript rule is stated for "minting onto a keyset with version byte `02` or later" and NUT-20's message for `00`/`01`, but a request mixing v2 and v3 outputs is neither. One sentence (e.g. "if any output is on a v3 keyset, the transaction digest is used") closes it. NUT-29's batched form has the same hole: the v3 rule requires every quote locked while the NUT-20 flow permits `null` signatures — a mixed batch is undefined.

10.md
```diff
@@ -2,19 +2,252 @@
...
+
+- A secret without conditions is a bare key: the tweak exists only when spending conditions do, with one exception below.
+- An **aggregated** internal key (MuSig2, FROST: nobody holds its scalar) **MUST** carry at least the empty tweak `t = tagged_hash("Cashu_TapTweak", K)`, no merkle root bytes, even when scriptless: cosigners can then verify no script path is hidden.
+- A **script-only** secret (no cooperative key path) uses BIP341's NUMS point as its internal key:
+
+  ```
+  K = 0250929b74c1a04954b78b4b6035e97a5e078a5a0f28ec96d547bfee9ace803ac0
+  ```
+
+  Wallets **MUST** use this specific point (`lift_x(SHA256(G))`, provably without a known discrete log) so any holder can recognize a script-only proof on sight. It is always used **verbatim**, never blinded ([NUT-28][28]): nobody holds its scalar for the receiver's half of the ECDH, so a blinded NUMS is indistinguishable from a sender-owned key with a hidden key path. Script-only proofs take their uniqueness from the tree (see [Key uniqueness](#key-uniqueness)).
```
a1denvalu3: **The `lift_x(SHA256(G))` parenthetical is misleading as written.** BIP341's H is `lift_x` of the SHA-256 of the *uncompressed* SEC1 encoding of G. I verified: `SHA256(0x04‖Gx‖Gy) = 50929b74…03ac0` ✓, but `SHA256(compressed G) = 0f715baf…0554` — a different value that isn't even a valid x-coordinate, so an implementer following the natural Cashu reading ("G" = `0279be66…`) can't reproduce the point and stalls. Suggest `lift_x(SHA256(G_uncompressed))`. The constant itself is correct and mandated verbatim, so this is purely a construction-recipe fix.

10.md
```diff
@@ -2,19 +2,252 @@
...
+A proof without spending conditions uses a bare key. The owner holds the keypair `(k, K = k*G)`:
+
+```
+secret = K
+```
+
+A proof with spending conditions commits them into the key, following [BIP341](https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki)'s construction with Cashu tags. The secret is the tweaked key:
+
+```
+secret = P = K + t*G,  where  t = tagged_hash("Cashu_TapTweak", K || merkle_root)
```
a1denvalu3: `t` is used directly as a scalar, but the spec never states the rule for `t ≥ n`: BIP341 *rejects* such outputs (and ported taproot code tends to carry that check), while `K + t*G` and `(k + t) mod n` here implicitly reduce. Astronomically rare (~2⁻¹²⁸), but this is exactly the kind of silent divergence a normative spec should pin: one line — "t is used mod n; never rejected" (or the opposite) — removes it.

10.md
```diff
@@ -2,19 +2,252 @@
...
+}
+```
+
+The presence of `leaf` selects script-path verification. The verifier **MUST**:
+
+1. Recompute `leaf_hash -> root` through `path`, then the tweak from `K` and the root, and check that `K + t*G` equals the secret; otherwise reject.
+2. Parse the leaf, failing closed on any unknown version, type, or even numbered field.
+3. Evaluate the leaf: for `after`, check its local clock is at or past `time`; for `hashlock`, check `SHA256(preimage) == hash`, with the preimage at most 32 bytes; then count the distinct listed keys with a valid BIP-340 signature over the digest and require at least `n`.
+
+Because Schnorr signatures are non-deterministic, thresholds count distinct satisfied keys, not signatures (as in [NUT-11][11]). Only the exercised leaf is revealed; sibling leaves remain opaque hashes in `path`. Given the caps above, a valid serialized witness never exceeds 8192 characters, and mints **MAY** enforce that as a size bound.
```
a1denvalu3: The depth cap (≤ 8) is stated for trees, but a verifier-side cap on the witness `path` array is only implied. Suggest making it explicit ("a witness whose `path` exceeds 8 elements MUST be rejected") — the 8192-character bound covers the DoS side, but the structural rule should stand on its own so verifiers don't have to derive it.

10.md
```diff
@@ -2,19 +2,252 @@
...
+A melt quote binds its quote id and its `amount`; the fee reserve is not part of the transcript, and [NUT-05][05] balance rules are unchanged. Outputs are bound exactly as they appear in the request, including blank change outputs. A transaction **MUST** have at least one input and at least one output.
+
+### The signing rule
+
+**Inputs sign; outputs never do.** Every v3 input **MUST** carry a witness containing a BIP-340 Schnorr signature over the transaction digest:
+
+- each proof input signs with its secret's key,
+- each mint quote input signs with its quote lock key ([NUT-04][04]).
+
+Verification is per input. Pre-v3 inputs in the same transaction keep their own rules (the [well-known Secret](#well-known-secret), [NUT-11][11], [NUT-14][14]), so transactions mixing keyset versions are valid. `SIG_ALL`, `sigflag`, and NUT-11's shared-secret rules do not exist for v3 inputs: inputs with different locks, or none, mix freely. (A pre-v3 `SIG_ALL` input cannot appear in a mixed transaction: [NUT-11][11] already requires every input to share its secret, so the transaction is rejected.) Inputs commit amount and keyset id explicitly, so each signature stands alone as evidence.
```
a1denvalu3: Small precision issue in the parenthetical: NUT-11 requires all inputs to share *kind, flag, `Secret.data` and `Secret.tags`* — not the secret itself (each proof's nonce differs). The conclusion is unaffected (a v3 point secret can't satisfy the same-kind requirement, so a mixed transaction containing a pre-v3 `SIG_ALL` input is rejected), but the stated reason should match NUT-11's actual rule.

00.md
```diff
@@ -418,6 +427,8 @@ Within the `t` (token) array, `i` denotes the keyset ID associated with the proo
 
 Unless otherwise stated, fields of type `bytes` represent byte strings in the CBOR encoding. In the original JSON representation of `Proof` objects, these values are encoded as hexadecimal strings. Implementations MUST convert between hex strings and raw byte arrays when translating between JSON and CBOR representations.
 
+`si` carries what the next holder of a v3 keyset proof needs beyond the proof itself; its fields and rules are defined in [NUT-10's spend info](10.md#spend-info).
```
a1denvalu3: Since "Receivers MUST ignore unknown fields" remains the rule just below, a v3-capable but `si`-unaware wallet that decodes and re-encodes a token will silently drop `si` and render any locked proof unspendable (the tweak needs the root). Given `si` is fund-critical, worth one sentence here: wallets MUST preserve `si` when re-serializing proofs they are not spending.

28.md
```diff
@@ -171,6 +171,21 @@ d. use standard derivation if `parity(P) == parity(pG)`, otherwise use negated d
...
+- **Slot `0` is always blinded**: a receiver-keyed send _is_ that blinding. The one exception is a [NUMS point](10.md#bare-empty-tweaked-and-script-only-secrets) base, which travels verbatim: nobody holds its scalar for the receiver's half of the ECDH, so a blinded NUMS is unverifiable. Slot indices are unchanged either way.
+- **Receivers match by value, never by position.** The root does not commit the transmitted leaf order ([NUT-10](10.md#receive-time-verification)). The receiver derives its candidate key for each slot in `1..N` (`N` = number of enumerated keys) and takes whichever slot reproduces a key in the tree. (A canonical leaf order is impossible: blinding rewrites the leaf bytes, so an order derived from them exists only after the slots it would decide.)
+- **At most 255 enumerated slots per secret**: slot `0` plus at most 254 leaf keys. Both sides **MUST** enforce the cap: a 256th key would derive slot `0`'s tweak, so an over-long tree is refused rather than truncated onto a colliding index.
```
orangeshyguy21: Slot numbers are one byte, so 0–255. Slot 0 is the base key, leaving 255 for leaf keys before anything wraps. But the cap here says 254.

Is 254 deliberate, or should it be 255?

tests/10-tests.md
```diff
@@ -0,0 +1,279 @@
...
+```
+
+```json
+{
+  "transcript": "0200160100010802000f71756f74652d6d696e742d3030303103005b0100010802002102b7e077d020fabed456a6be138a8e20e9ef40b44d873fa12c005b656eb0cf99f6030030b42a0bcc39598db1dca617aeea6bc367f2566636826dc961a54faae15b3b8d10afc1cb0206e70ab3b0e12c2b9478cd55",
+  "digest": "096a9b2002cc0b8ebc9b79e0902159385a929f4e63f35eb9e1dee0119205efb6"
+}
+```
+
+**Melt.** Paying melt quote `quote-melt-0001` (amount 8, no change outputs) with the swap's 8-sat proof as the only input; the melt quote is the only output, binding its quote id and amount:
```
orangeshyguy21, line 338: Worth adding a second melt vector be added with a couple of blank change outputs, showing the 0x03 records ahead of the 0x04? 

It's the only shape that exercises the container ordering rule, and NUT-05's wording reads the other way round.

robwoodgate: Good call - test vector and NUT-05 note added in 9a1d6601ee04c554ba89d16423e481bda09b6eac

10.md
robwoodgate: Now noted below in the notes - added in 9a1d6601ee04c554ba89d16423e481bda09b6eac

10.md
robwoodgate: Good point, added in 9a1d6601ee04c554ba89d16423e481bda09b6eac

10.md
robwoodgate: Thank you - addressed in 9a1d6601ee04c554ba89d16423e481bda09b6eac

10.md
robwoodgate: Thank you - clarified in 9a1d6601ee04c554ba89d16423e481bda09b6eac

00.md
robwoodgate: Good point - addressed in 9a1d6601ee04c554ba89d16423e481bda09b6eac

04.md
robwoodgate: Clarified in 46789a2d4f168a8968d7b20c97fb4ae62af1606e

10.md
robwoodgate: Thank you - clarified in a53e25888257e5fdbe1cdd702ca7bf5b8d1f53cf

28.md
robwoodgate: Good catch - an off-by-one oversight, not deliberate. Fixed in 0139aa05434e4fee3b341c5fc706934c4d722481

10.md
robwoodgate: Clarified in 2ea7bbb75afde5c36cd24dbacccba26efbfce40f

10.md
robwoodgate: Clarified in 2ea7bbb75afde5c36cd24dbacccba26efbfce40f

10.md
robwoodgate: Clarified in 2ea7bbb75afde5c36cd24dbacccba26efbfce40f

18.md
robwoodgate: Good catch - clarified in aee4c9fd95498795a009740da0a3d0a87cbb7e88

00.md
```diff
@@ -253,6 +259,8 @@ A `Proof` is also called an _input_ and is generated by `Alice` from a `BlindSig
 
 `amount` is the amount of the `Proof`, `secret` is the secret message and is a utf-8 encoded string (the use of a 64 character hex string generated from 32 random bytes is recommended to prevent fingerprinting), `C` is the unblinded signature on `secret` (hex string), `id` is the [keyset id][02] of the mint public keys that signed the token (hex string).
 
+`secret` **MUST** be unique per `Proof`, on every keyset version: blinding a repeated secret on the same keyset and amount reproduces the same `C`, so the second issuance re-creates an existing `Proof` and its value is lost.
+
```
callebtc: I think mints require also that it's unique across keysets. the double-spend check is also on Y.

robwoodgate: Good catch, fixed in 99c6a0c

SatsAndSports:
So duplicate leaves are allowed, right? The NUT isn't super clear, and the PoCs accept them without deduplication

I'm asking my agent to propose changes to make it explicit that duplicates leaves are allowed, and to add tests to the PoCs to show that duplicates can be spend safely

But feel free to make any changes yourself of course.

Or perhaps the sorting should discard duplicates?

robwoodgate:
> So duplicate leaves are allowed, right? The NUT isn't super clear, and the PoCs accept them without deduplication

Good spot, yes they are allowed and have modified spec to specify. A de-dup would introduce an ambiguity ("equal by what?": bytes or semantics) and whilst we anticipate no-one would WANT to duplicate a leaf (it wastes a slot) there is no reason to try fix it (and BIP341 allows duplicates too)

SatsAndSports:
The proposed transcript binds to the _output_ keyset ids. This may be a problem for the spilman channels

The expiry date for a keyset only requires the mint to keep the keyset usable _as inputs_ up until the expiry date. A mint is free to make a keyset inactive at any time, without warning and regardless of any expiry date.

And because the mint has this freedom to de-activate any keyset at any time, we should ensure that (partially-)pre-signed transactions stay valid.

When the SIG_ALL message was extended last year #302, there was a deliberate decision to keep the output keyset id out of the message and this was done to allow presigned transactions to survive a rotation of the active keyset(s).

Maybe we could add an `active_until` field to a keyset, to require the mint to keep it active until at least that timestamp?

a1denvalu3:
@robwoodgate Here's one aspect that was probably overlooked: NUT28 caps enumerated keys at 255, as any subsequent key would wrap around. Now we allow 256 leaves, each of 1024 bytes. That gives us ~30 x 256 possible keys while the specification only accounts for up to 256.

a1denvalu3 (commented):
Two issues from a security pass over the nutroot design:

### 1. Parity twins: `02‖x` and `03‖x` are distinct secrets controlled by one key

The two parity encodings of the same x-coordinate are different byte strings, so they hash to unrelated `Y` points — two mint spent-set slots. But key-path verification is x-only: the scalar holder spends `02‖x` with `d` and `03‖x` with `n−d`. One key, two spend slots.

This quietly breaks the one-key-one-secret invariant the "Key uniqueness" section reasons from: the same `(K, tree)` with flipped parity is a distinct secret with identical spend authority. The leaf rules already reject parity twins where they protect thresholds; nothing analogous exists at the secret level.

Fallout: anything assuming "value locked under key `x` is a single slot" — escrow bookkeeping, the auditable-lock verification, NUT-07 state checks (per-`Y`, so twins are invisible to each other) — is off by a factor of two. Cosigners of an aggregated `K` can also be surprised by a twin proof funded without their knowledge — not theft, but the key no longer pins down how much value or how many slots it controls.

Mandating even-Y secrets can't fix this: a tweaked `P = K + t·G` has pseudorandom parity. So the realistic remediation is documentation — state the twin relation in the uniqueness section and make sure higher-level logic never assumes one key ⇒ one slot.

### 2. NUT-07 witness disclosure reaches every past holder

NUT-07 returns `witness` (and `digest`) for every spent proof, and `Y` is computable from the secret — so any past holder of the token can query checkstate and read the eventual spend's witness. A script-path spend therefore discloses the exercised leaf (refund keys, timelocks, hashlock preimages) to the proof's entire prior custody chain, not just to the mint.

That undercuts NUT-10's privacy claim ("the mint learns of conditions … only when a script path is exercised"): the disclosure set is the mint plus everyone who ever held the token. A sender can watch a payment they made and learn which leaf the receiver exercised, and when.

If intentional (spend auditability), state it as an explicit privacy property of script-path spends; if not, consider gating witness disclosure or documenting that leaf contents are visible to prior holders.

SatsAndSports:
_deleted my comment of a few seconds ago. I was wrong_

a1denvalu3:
Benchmarked the witness verification cost of this spec as implemented in cashubtc/nutshell#999 (`_verify_nutroot_transaction_witnesses`, Python 3.12, i7-1365U, per-call medians; script on branch `bench/nutroot-witness-verification`):

**By witness type** (1 input, 1 output): key path ~87 µs (bare or tweaked) · script path 1-of-1 ~171 µs (`after`/`hashlock`/p2bk-blinded keys identical) · 2-of-3 ~395 µs.

**Scaling**: linear in inputs (~75 µs/key-path input, ~159 µs/script input; 64 script inputs ≈ 10.3 ms); flat in tree size (~2.2 µs per merkle level, 256-leaf cap adds ~18 µs over a single leaf).

**One spec-relevant observation**: the threshold evaluation loop (count satisfied keys against provided signatures) costs m·n − n(n−1)/2 schnorr verifies at ~55 µs each — 8-of-15 ≈ 5.2 ms, ~30× a 1-of-1 spend. Bounded by the 1024-byte leaf cap (~31 keys ⇒ up to ~496 verify attempts ≈ 27 ms per proof in the worst case). If large thresholds are meant to be practical, the spec could suggest matching signatures to keys by position or a canonical sig order; otherwise the leaf cap already bounds it.

secp256k1 ops dominate everywhere (~43 µs/schnorr verify, ~34 µs/tweak-add); witness JSON parsing and merkle paths are noise.

robwoodgate:
> @robwoodgate Here's one aspect that was probably overlooked: NUT28 caps enumerated keys at 255, as any subsequent key would wrap around. Now we allow 256 leaves, each of 1024 bytes. That gives us ~30 x 256 possible keys while the specification only accounts for up to 256.

Great catch! The limits have been reduced, 8 leaves each of 512 bytes. This brings the total keys down to ~15 x 8 = 120, allowing plenty of room for expansion later if needed.

https://github.com/cashubtc/nuts/pull/421/changes/ef2919462b5cd0d1ae90a774b8eb3a09e1266178

robwoodgate:
> Two issues from a security pass over the nutroot design:
> 
> ### 1. Parity twins: `02‖x` and `03‖x` are distinct secrets controlled by one key

Normative note added to NUT-10: https://github.com/cashubtc/nuts/pull/421/changes/27771dde084d25589a0e78630e39b03fd8109bd1

> ### 2. NUT-07 witness disclosure reaches every past holder

This has been addressed by returning only a binding spend commitment for v3 proofs by default, keeping the witness and input digest private. A leaf may explicitly opt into public disclosure (disclosure = 0x01) for hashlocks/auditable locks, with identical behavior over NUT-07 and NUT-17

https://github.com/cashubtc/nuts/pull/421/changes/b2c6696b4a1963733db648f48146c15d5a86383b

https://github.com/cashubtc/nutshell/pull/1139

CTS PR updated

robwoodgate:
Merged into the holding branch by mistake; the branch is restored to 68840cc and this PR continues as #443 (same-repo head).

PR #422, opened 2026-08-12
Loosen NUT-29 batch quote checks

Egge21M:
# Implementations

- [ ] Nutshell https://github.com/cashubtc/nutshell/pull/1114
- [ ] CDK TBC
- [x] Cashu-TS https://github.com/cashubtc/cashu-ts/pull/953

## What changed

- require mints to omit unknown or malformed quote IDs from NUT-29 batch-check responses
- preserve the relative request order of quotes that are returned
- define an empty array response when no requested quote IDs can be handled
- update the NUT-29 test vector for partial and empty results
- replace stale mint-quote `state` references with NUT-04's `amount_paid`, `amount_issued`, and `updated_at` accounting model
- validate batch-mint eligibility using the currently mintable amount, `amount_paid - amount_issued`

## Why

Batch quote checking is a read-only query. A wallet should still receive usable quote states when some requested identifiers are malformed or unknown, without weakening the atomic validation of the batch mint endpoint.

## Impact

Mints can process batch-check requests on a best-effort basis. Wallets must treat the response as an order-preserving subset of the requested quotes.

Batch-check responses and batch-mint validation now use NUT-04's canonical mint-quote accounting fields instead of the deprecated state model.

Batch mint execution remains atomic and unchanged.

## Validation

- `npx prettier@3.9.1 --check 29.md tests/29-tests.md`
- `git diff --check origin/main...HEAD`

### DIFF
--- a/29.md
+++ b/29.md
@@ -74,13 +74,16 @@ Content-Type: application/json
     "unit": "sat",
     "amount": 50,
     "expiry": 1234567890
-  }
-]
-```
-
 #### Unknown or Malformed Quote IDs
 
-This is a query endpoint and does not use the all-or-nothing error handling of the batch mint endpoint. If a quote ID is unknown or cannot be parsed, the mint MUST omit it from the response and continue processing the other quote IDs. If none of the quote IDs can be handled, the mint MUST return an empty JSON array.
+This is a query endpoint and does not use the all-or-nothing error handling of the batch mint endpoint. If a quote ID is unknown or cannot be parsed, the mint **MUST** return an `unknown` entry for it as below and continue processing the remaining IDs.
+
+```json
+{
+   "quote": <str>,
+   "unknown": true
+}
+```
 
 ---
 
--- a/29.md
+++ b/29.md
@@ -36,11 +36,11 @@ Mints **MUST** return the quote object for every quote ID they hold a record of,
 
 Mints **MUST NOT** delete a quote for which ecash is still owed (amount_paid - amount_issued > 0).
 
-#### Example
+### Example
 
 Below is an example for checking two bolt11 mint quotes.
 
-##### Request
+#### Request
 
 ```http
 POST https://mint.host:3338/v1/mint/quote/bolt11/check
@@ -51,9 +51,9 @@ Content-Type: application/json
 }
 ```
 
-##### Response
+#### Response
 
-````json
+```json
 [
   {
     "quote": "019e6d5a-2347-7000-8037-b42dae20f1fe",
@@ -74,7 +74,10 @@ Content-Type: application/json
     "unit": "sat",
     "amount": 50,
     "expiry": 1234567890
-#### Unknown or Malformed Quote IDs
+  }
+```
+
+### Unknown or Malformed Quote IDs
 
 This is a query endpoint and does not use the all-or-nothing error handling of the batch mint endpoint. If a quote ID is unknown or cannot be parsed, the mint **MUST** return an `unknown` entry for it as below and continue processing the remaining IDs.
 
@@ -83,7 +86,7 @@ This is a query endpoint and does not use the all-or-nothing error handling of t
    "quote": <str>,
    "unknown": true
 }
-````
+```
 
 ---
 
--- a/29.md
+++ b/29.md
@@ -89,6 +89,8 @@ This is a query endpoint and does not use the all-or-nothing error handling of t
 }
 ```
 
+`max_batch_size` (see [Batch Size Limits](#batch-size-limits)) applies to this endpoint. A request that exceeds it **MUST** be rejected in full with error `11017` rather than answered partially, so a successful response always has exactly one entry per requested quote ID.
+
 ---
 
 ## 2. Executing the Batched Mint
--- a/29.md
+++ b/29.md
@@ -53,7 +53,7 @@ Content-Type: application/json
 
 ##### Response
 
-```json
+````json
 [
   {
     "quote": "019e6d5a-2347-7000-8037-b42dae20f1fe",
@@ -83,7 +83,7 @@ This is a query endpoint and does not use the all-or-nothing error handling of t
    "quote": <str>,
    "unknown": true
 }
-```
+````
 
 ---
 
--- a/29.md
+++ b/29.md
@@ -12,7 +12,7 @@ This spec describes how a wallet can mint multiple quotes in one batched operati
 
 ## 1. Batch Checking Mint Quotes
 
-Before minting, the wallet SHOULD verify each mint quote's current accounting state. It does this by sending:
+Before minting, the wallet SHOULD verify each mint quote's current accounting data. It does this by sending:
 
 ```http
 POST https://mint.host:3338/v1/mint/quote/{method}/check
@@ -55,7 +55,6 @@ Content-Type: application/json
     "amount_paid": 100,
     "amount_issued": 0,
     "updated_at": 1234567800,
-    "state": "PAID",
     "unit": "sat",
     "amount": 100,
     "expiry": 1234567890
@@ -66,7 +65,6 @@ Content-Type: application/json
     "amount_paid": 0,
     "amount_issued": 0,
     "updated_at": 1234567800,
-    "state": "UNPAID",
     "unit": "sat",
     "amount": 50,
     "expiry": 1234567890
@@ -189,7 +187,7 @@ The mint MUST validate the following before processing a batch mint request:
 3. **Valid quote IDs**: All quote IDs MUST exist in the mint's database
 4. **Payment method consistency**: All quotes MUST have the same payment method, matching `{method}` in the URL path
 5. **Currency unit consistency**: All quotes MUST use the same currency unit
-6. **Quote state**: All quotes MUST be in PAID state (or have a mintable amount for payment methods that allow multiple mint operations like bolt12)
+6. **Mintable amount**: All quotes MUST have a positive currently mintable amount (`amount_paid - amount_issued`)
 7. **Amount balance**: The sum of amounts contained in the `outputs` MUST equal the sum of `quote_amounts` (bolt11) or MUST NOT exceed it (bolt12)
 8. **Signature validation (NUT-20)**: The `signatures` array length MUST match the `quotes` array length; locked quotes MUST include a valid signature; unlocked quotes MUST NOT include one
 
--- a/29.md
+++ b/29.md
@@ -28,13 +28,13 @@ The wallet includes the following body in its request:
 
 where `quotes` is an array of _unique_ mint quote IDs.
 
-The mint returns a JSON array with exactly one entry per requested quote ID, in request order. Entry `I` corresponds to `quotes[I]`, and its quote field **MUST** equal the requested ID.
+The mint returns a JSON array with exactly one entry per requested quote ID, in request order. Entry `i` corresponds to `quotes[i]`, and its `quote` field **MUST** equal the requested ID.
 
 Known quotes are returned as mint quote objects as defined by the payment method's NUT specification. A quote ID that the mint does not know or cannot parse is returned as an [unknown entry](#unknown-or-malformed-quote-ids).
 
-Mints **MUST** return the quote object for every quote ID they hold a record of, including quotes whose expiry has passed. An unknown entry indicates only that the mint holds no record of the ID.
+Mints **MUST** return the quote object for every quote ID they hold a record of, including quotes whose `expiry` has passed. An unknown entry indicates only that the mint holds no record of the ID.
 
-Mints **MUST NOT** delete a quote for which ecash is still owed (amount_paid - amount_issued > 0).
+Mints **MUST NOT** delete a quote for which ecash is still owed (`amount_paid - amount_issued > 0`).
 
 ### Example
 
@@ -75,6 +75,7 @@ Content-Type: application/json
     "amount": 50,
     "expiry": 1234567890
   }
+]
 ```
 
 ### Unknown or Malformed Quote IDs
@@ -83,8 +84,8 @@ This is a query endpoint and does not use the all-or-nothing error handling of t
 
 ```json
 {
-   "quote": <str>,
-   "unknown": true
+  "quote": <str>,
+  "unknown": true
 }
 ```
 
--- a/29.md
+++ b/29.md
@@ -28,7 +28,7 @@ The wallet includes the following body in its request:
 
 where `quotes` is an array of _unique_ mint quote IDs.
 
-The mint returns a JSON array of mint quote objects as defined by the payment method's NUT specification. The quotes in this array MUST be in the same order as in the request.
+The mint returns a JSON array of mint quote objects as defined by the payment method's NUT specification. The mint MUST omit quote IDs that it does not know or cannot parse. The quotes that are returned MUST be in the same relative order as their quote IDs in the request.
 
 #### Example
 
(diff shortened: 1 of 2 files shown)


asmogo:
I got two questions: 

1. Why are we still using the state in the mint quote batch response? I think this was deprecated in nut-04. 

2. Do you think we should omit unknown quote IDs? Instead of omitting, we could signal to the client that there  is an unknown id included. This should make wallet implementations more straightforward.

29.md
```diff
@@ -28,7 +28,7 @@ The wallet includes the following body in its request:
 
 where `quotes` is an array of _unique_ mint quote IDs.
 
-The mint returns a JSON array of mint quote objects as defined by the payment method's NUT specification. The quotes in this array MUST be in the same order as in the request.
+The mint returns a JSON array of mint quote objects as defined by the payment method's NUT specification. The mint MUST omit quote IDs that it does not know or cannot parse. The quotes that are returned MUST be in the same relative order as their quote IDs in the request.
```
robwoodgate:
```suggestion
The mint returns a JSON array with exactly one entry per requested quote ID, in request order. Entry `I` corresponds to `quotes[I]`, and its quote field **MUST** equal the requested ID.

Known quotes are returned as mint quote objects as defined by the payment method's NUT specification. A quote ID that the mint does not know or cannot parse is returned as an [unknown entry](#unknown-or-malformed-quote-ids).

Mints **MUST** return the quote object for every quote ID they hold a record of, including quotes whose expiry has passed. An unknown entry indicates only that the mint holds no record of the ID.

Mints **MUST NOT** delete a quote for which ecash is still owed (amount_paid - amount_issued > 0).
```

29.md
```diff
@@ -66,20 +65,16 @@ Content-Type: application/json
...
 ```
 
-#### Error Handling
+#### Unknown or Malformed Quote IDs
 
-This is a query endpoint that uses all-or-nothing error handling, matching the behavior of the batch mint endpoint:
-
-- If any `quote_id` is not known by the mint, the mint MUST reject the entire request and return an appropriate error
-- If any `quote_id` cannot be parsed (invalid format), the mint MUST reject the entire request and return an appropriate error
+This is a query endpoint and does not use the all-or-nothing error handling of the batch mint endpoint. If a quote ID is unknown or cannot be parsed, the mint MUST omit it from the response and continue processing the other quote IDs. If none of the quote IDs can be handled, the mint MUST return an empty JSON array.
```
robwoodgate:
````suggestion
#### Unknown or Malformed Quote IDs

This is a query endpoint and does not use the all-or-nothing error handling of the batch mint endpoint. If a quote ID is unknown or cannot be parsed, the mint **MUST** return an `unknown` entry for it as below and continue processing the remaining IDs.

```json
{
   "quote": <str>,
   "unknown": true
}
```
````

File: tests/29-tests.md [hunk 3]
```diff
@@ -20,23 +20,35 @@ The following is the corresponding response with a blind signature.
...
+    "amount_issued": 0,
+    "updated_at": 1234567800
+  },
+  {
+    "quote": "known-2",
+    "amount_paid": 0,
+    "amount_issued": 0,
+    "updated_at": 1234567800
+  }
+]
```
robwoodgate, line 96:
```suggestion
[
  {
    "quote": "known-1",
    "amount_paid": 5,
    "amount_issued": 0,
    "updated_at": 1234567800
  },
  { "quote": "not-a-valid-quote-id", "unknown": true },
  { "quote": "unknown-2", "unknown": true },
  {
    "quote": "known-2",
    "amount_paid": 0,
    "amount_issued": 0,
    "updated_at": 1234567800
  }
]
```

robwoodgate (changes requested):
ACK the concept. This will only be implemented on mints who will also be aware of the UUIDv7 quote id requirement, so my concerns about creating a batch oracle are avoided (UUIDv7 has 74 bits of entropy per millisecond timestamp).

We should also consider making UUIDv7 a **MUST**, at least until all quotes are locked.

I still think we should return unknowns, as this tells a wallet unequivocably the quote should be dropped (just omitting unknowns could indicate a mint bug, truncated reponse or other ambiguities).

My suggestion is to include a positional "unknown", as this also makes response processing simpler: quoteids in = quote responses out in same order

29.md
```diff
@@ -66,20 +71,23 @@ Content-Type: application/json
...
-- If any `quote_id` is not known by the mint, the mint MUST reject the entire request and return an appropriate error
-- If any `quote_id` cannot be parsed (invalid format), the mint MUST reject the entire request and return an appropriate error
+```json
+{
+  "quote": <str>,
+  "unknown": true
+}
+```
 
 ---
```
robwoodgate, line 98: Probably worth stating aloud:

```suggestion
`max_batch_size` (see [Batch Size Limits](#batch-size-limits)) applies to this endpoint. A request that exceeds it **MUST** be rejected in full with error `11017` rather than answered partially, so a successful response always has exactly one entry per requested quote ID.

---
```

tests/29-tests.md
```diff
@@ -20,23 +20,37 @@ The following is the corresponding response with a blind signature.
 }
 ```
 
-## Check endpoint with unknown quotes
+## Check endpoint omits unknown and malformed quotes
```
robwoodgate:
```suggestion
## Check endpoint marks unknown and malformed quotes
```

tests/29-tests.md
```diff
@@ -20,23 +20,37 @@ The following is the corresponding response with a blind signature.
...
-The following is an invalid check request containing an unknown quote ID.
+The following check request contains two known quote IDs, one malformed quote ID, and one unknown quote ID.
 
 ```json
-{ "quotes": ["known-1", "bogus", "unknown-2"] }
+{ "quotes": ["known-1", "not-a-valid-quote-id", "unknown-2", "known-2"] }
 ```
 
-Per NUT-29, quote check uses all-or-nothing error handling. If any quote is unknown, the entire request must be rejected.
+The mint omits the malformed and unknown quote IDs. It returns the known quotes in the same relative order as their IDs in the request.
```
robwoodgate:
```suggestion
The mint returns one entry per requested quote ID, in request order. The malformed and unknown quote IDs are returned as `unknown` entries.
```

tests/29-tests.md
```diff
@@ -20,23 +20,37 @@ The following is the corresponding response with a blind signature.
...
+  {
+    "quote": "known-2",
+    "amount_paid": 0,
+    "amount_issued": 0,
+    "updated_at": 1234567800
+  }
+]
 ```
 
+If the mint cannot handle any of the requested quote IDs, it returns an empty JSON array.
```
robwoodgate:
```suggestion
If the mint cannot handle any of the requested quote IDs, every entry in the response is an `unknown` entry. The response is never shorter than the request.
```

robwoodgate (changes requested):
Apologies I munged the suggestions last time, so made two commits to fix them up after you merged them. Below are the remaining suggestions to make it consistent.

29.md
```diff
@@ -66,20 +71,25 @@ Content-Type: application/json
...
     "expiry": 1234567890
   }
 ]
 ```
 
-#### Error Handling
+### Unknown or Malformed Quote IDs
 
-This is a query endpoint that uses all-or-nothing error handling, matching the behavior of the batch mint endpoint:
+This is a query endpoint and does not use the all-or-nothing error handling of the batch mint endpoint. If a quote ID is unknown or cannot be parsed, the mint **MUST** return an `unknown` entry for it as below and continue processing the remaining IDs.
```
robwoodgate, line 85: I wonder if we should name the return type, as we do for other response shapes.

```suggestion
This is a query endpoint and does not use the all-or-nothing error handling of the batch mint endpoint. If a quote ID is unknown or cannot be parsed, the mint **MUST** return an `UnknownQuote` for it as below and continue processing the remaining IDs.
```

a1denvalu3 (commented):
.

tests/29-tests.md
```diff
@@ -20,23 +20,37 @@ The following is the corresponding response with a blind signature.
...
```
a1denvalu3: This contradicts 29.md: the spec now says a successful response always has exactly one entry per requested ID, but this line still says an empty array is returned. An empty response to a non-empty request breaks the positional contract. Shouldn't this read that every entry is an `unknown` entry (response never shorter than the request)?

EDIT: just needs to be updated ig

29.md
```diff
@@ -192,7 +202,7 @@ The mint MUST validate the following before processing a batch mint request:
 3. **Valid quote IDs**: All quote IDs MUST exist in the mint's database
 4. **Payment method consistency**: All quotes MUST have the same payment method, matching `{method}` in the URL path
 5. **Currency unit consistency**: All quotes MUST use the same currency unit
-6. **Quote state**: All quotes MUST be in PAID state (or have a mintable amount for payment methods that allow multiple mint operations like bolt12)
+6. **Mintable amount**: All quotes MUST have a positive currently mintable amount (`amount_paid - amount_issued`)
 7. **Amount balance**: The sum of amounts contained in the `outputs` MUST equal the sum of `quote_amounts` (bolt11) or MUST NOT exceed it (bolt12)
```
a1denvalu3: With rule 6 loosened to "positive mintable amount", a batch can now include an already partially-minted bolt11 quote — but this equality check doesn't say what the target is in that case: the quote's original amount, or its currently mintable amount (`amount_paid - amount_issued`)? If it's the original amount, a mint can be pushed past NUT-04's cap. Worth defining, including what `quote_amounts` means for a partially-issued quote.

29.md
```diff
@@ -28,13 +28,19 @@ The wallet includes the following body in its request:
 
 where `quotes` is an array of _unique_ mint quote IDs.
 
-The mint returns a JSON array of mint quote objects as defined by the payment method's NUT specification. The quotes in this array MUST be in the same order as in the request.
+The mint returns a JSON array with exactly one entry per requested quote ID, in request order. Entry `i` corresponds to `quotes[i]`, and its `quote` field **MUST** equal the requested ID.
 
-#### Example
+Known quotes are returned as mint quote objects as defined by the payment method's NUT specification. A quote ID that the mint does not know or cannot parse is returned as an [unknown entry](#unknown-or-malformed-quote-ids).
+
+Mints **MUST** return the quote object for every quote ID they hold a record of, including quotes whose `expiry` has passed. An unknown entry indicates only that the mint holds no record of the ID.
```
a1denvalu3: What about an ID the mint *does* hold, but under a different method (a bolt12 ID sent to `/quote/bolt11/check`), or a melt quote? Returning the object leaks cross-method existence; returning `unknown` contradicts "holds no record of the ID". Suggest specifying the endpoint is scoped to `{method}` and anything outside that scope MUST be returned as unknown.

29.md
```diff
@@ -28,13 +28,19 @@ The wallet includes the following body in its request:
 
 where `quotes` is an array of _unique_ mint quote IDs.
 
-The mint returns a JSON array of mint quote objects as defined by the payment method's NUT specification. The quotes in this array MUST be in the same order as in the request.
+The mint returns a JSON array with exactly one entry per requested quote ID, in request order. Entry `i` corresponds to `quotes[i]`, and its `quote` field **MUST** equal the requested ID.
 
-#### Example
+Known quotes are returned as mint quote objects as defined by the payment method's NUT specification. A quote ID that the mint does not know or cannot parse is returned as an [unknown entry](#unknown-or-malformed-quote-ids).
```
a1denvalu3: Worth being explicit about what "mint quote objects as defined by the payment method's NUT" includes here. Returning the full object (incl. `request`) makes this endpoint an efficient existence oracle: per-entry unknowns leak ~N bits per query vs ~1 bit with all-or-nothing, and NUT-04 only makes UUIDv7 a SHOULD. Either require strong quote IDs for mints advertising NUT-29, or return only the accounting subset (`quote`, `amount_paid`, `amount_issued`, `updated_at`) — the test vector already shows that minimal shape.

29.md
```diff
@@ -66,20 +71,25 @@ Content-Type: application/json
...
+### Unknown or Malformed Quote IDs
 
-This is a query endpoint that uses all-or-nothing error handling, matching the behavior of the batch mint endpoint:
+This is a query endpoint and does not use the all-or-nothing error handling of the batch mint endpoint. If a quote ID is unknown or cannot be parsed, the mint **MUST** return an `unknown` entry for it as below and continue processing the remaining IDs.
+
+```json
+{
+  "quote": <str>,
+  "unknown": true
+}
```
a1denvalu3, line 91: Suggest stating that unknown entries contain exactly `quote` and `unknown: true`, so wallets have a clean discriminator and an entry can't be both a full quote object and `unknown`.

29.md
```diff
@@ -192,7 +202,7 @@ The mint MUST validate the following before processing a batch mint request:
 3. **Valid quote IDs**: All quote IDs MUST exist in the mint's database
 4. **Payment method consistency**: All quotes MUST have the same payment method, matching `{method}` in the URL path
 5. **Currency unit consistency**: All quotes MUST use the same currency unit
-6. **Quote state**: All quotes MUST be in PAID state (or have a mintable amount for payment methods that allow multiple mint operations like bolt12)
+6. **Mintable amount**: All quotes MUST have a positive currently mintable amount (`amount_paid - amount_issued`)
```
a1denvalu3: Dropping the PAID-state rule also drops any mention of expiry from mint validation: nothing in this list says whether an expired (but paid) quote can still be batch-minted. Some mints will reject, some won't — should be stated explicitly, especially since the check section now requires returning expired quotes.

29.md
```diff
@@ -28,13 +28,19 @@ The wallet includes the following body in its request:
 
 where `quotes` is an array of _unique_ mint quote IDs.
```
a1denvalu3: For the check endpoint specifically: is an empty `quotes` array allowed (→ empty response)? And are duplicate IDs rejected like in batch mint (11016), or answered positionally? Both are undefined for check.

a1denvalu3 (commented):
Splitting my earlier review into threads on the relevant lines. The two I'd block on are the empty-array contradiction in the test vector and the undefined equality check for partially-minted quotes.

tests/29-tests.md
robwoodgate: In CTS, I have a failIf(quotes.length != responses.length), so if the view is to make an exception for fully failed lookup, I will need to loosen that.

29.md
Egge21M: With real-life rate limits and mints already shipping UUIDv7 I dont think that is an issue. We previously discussed this. We could change to UUIDv7 to MUST though just to be explicit

29.md
robwoodgate: I believe my suggestion to define an "UnknownQuote" object would accomplish the same thing?

https://github.com/cashubtc/nuts/pull/422/changes/e5cb468d190a846b3144358a4863e7c0c2d4bbe5..84a34a8005468735b685532f17709712d421082a#r3768043952

tests/29-tests.md
```diff
@@ -20,23 +20,145 @@ The following is the corresponding response with a blind signature.
...
+    "amount_paid": 0,
+    "amount_issued": 0,
+    "updated_at": 1234567800
+  }
+]
+```
+
+Each unknown entry contains exactly `quote` and `unknown: true`. Known quote entries do not contain the `unknown` field.
+
+If the mint cannot handle any of the requested quote IDs, every entry in the response is an `unknown` entry. The response is never shorter than the request.
```
crodas, line 101: What is preventing a malicious actor from crafting the maximum number of quote IDs and sending requests?  What are the limits?

Egge21M: What would the malicious intend be? Mints can enforce their own limits on batch sizes

PR #426, opened 2026-08-14
NUT-08: require amount 0 on blank change outputs

robwoodgate:
The amount of a blank output is ignored by the mint, and NUT-08 currently suggests setting it to `1`. Current implementations (cdk, cashu-ts) send `0`, as this is a stronger "blank" signal than `1`, so this makes `0` a requirement for wallets and updates the example to match.

Mints must still ignore the field, so wallets that send `1` today keep working.

## Implementations

- [x] Cashu-TS: Already compliant
- [x] CDK: Already compliant
- [x] Nutshell: https://github.com/cashubtc/nutshell/pull/1118

### DIFF
--- a/08.md
+++ b/08.md
@@ -112,7 +112,7 @@ curl -X POST https://mint.host:3338/v1/melt/bolt11 -d \
 }'
 ```
 
-Everything here is the same as in [NUT-05][05] except for `outputs`. The `amount` field in the `BlindedMessage`s here are ignored by `Bob` so they can be set to any arbitrary value by `Alice` (they should be set to a value, like `1` so potential JSON validations do not error).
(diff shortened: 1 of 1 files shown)


robwoodgate:
@ye0man - this wording now reflects what all implementations do consistently

prusnak:
Shouldn't we require setting to 0 in the NUT?

robwoodgate:
> Shouldn't we require setting to 0 in the NUT?

I was keeping it advisory as it was before, but you are right - it's overall a better semantic setting for a blank

Addressed in e4e7cf9c7c6e5e24c006cd6611aab83d053f3296, along with example fix

thesimplekid (approved):

a1denvalu3 (approved):

PR #429, opened 2026-08-20
NUT-05: Clarify melt retries and the state transitions

### LINKED ISSUE #428: Does the spec mandate that mints should support retries on failed `melt` quotes? or discourage retries? or they are optional?
## Does the spec mandate that mints should support retries on failed `melt` quotes? or discourage retries? or they are optional?

_**Update: now that I've read more and tested more and thought more, I think I'm fine with the spec as-is. I _might_ propose a small adjustment to the NUT to clarify**_

All three mints that I'm aware of (Nutshell, Nutmix and CDK) support retries. See the comment below for more

I'm thinking of bolt11 melt here, but I/we should think about all melt methods.

(this is inspired by trying to understand the various 'state machines' in Cashu and Lightning, where I like to know if a given state is _terminal_ or not)

### details (a bolt11 example)

Context: A wallet has a `quote_id`, and then successfully begins an execution via `/v1/melt/bolt11`. Eventually the Lightning payment fails (maybe a bad route) and the quote's status `/v1/melt/quote/bolt11/{quote_id}` returns to `UNPAID` (which means _(non-terminally?)_ 'failed', in this context)

If the wallet now makes another attempt to execute `/v1/melt/bolt11` the same quote, should the mint accept it and try again?

### Suggested outcome

In each of the melt types (bolt11, bolt12, on-chain), make clear whether mints are expected to be able to support retries. Or just put this in NUT-05 if we have a single answer that applies to all melt types

And maybe adjust the NUTs to clarity that UNPAID can mean (among other things) that an attempt was made and that it failed

SatsAndSports:
## Summary

- define `UNPAID -> PENDING -> PAID|UNPAID` melt quote transitions
- require failed attempts to release inputs and remain idle at `UNPAID` until a wallet explicitly retries
- prevent repeated execution requests from causing a second payment while `PENDING` or after terminal `PAID`, without restricting payment-backend route attempts
- apply the common retry rules to BOLT11 and BOLT12, and preserve the selected onchain `fee_index` across retries of the same quote

## Context

The original NUT-05 state change explicitly allowed wallets to repeat an `UNPAID` melt request until payment succeeded. That sentence was lost when NUT-05 was later abstracted across payment methods. This restores and tightens the existing behavior without changing request or response formats.

Closes #428.

## Validation

- `npx prettier@3.9.1 --check .`

### DIFF
--- a/05.md
+++ b/05.md
@@ -122,7 +122,7 @@ When the mint accepts an execution request for a quote in the `"UNPAID"` state, 
 
 A wallet **MAY** retry a quote in the `"UNPAID"` state by sending another execution request for the same quote. The mint **MUST NOT** reject an otherwise valid execution request solely because a previous payment attempt failed. A retry is a new payment attempt, so the mint **MUST** validate its inputs, outputs, and method-specific fields as it would for an initial attempt.
 
-The mint **MUST NOT** start another payment attempt while the quote is `"PENDING"` or after it is `"PAID"`. A quote in the `"UNPAID"` state **MUST** remain `"UNPAID"` unless the wallet sends another execution request that the mint accepts. Checking the quote state can reconcile an existing payment attempt but **MUST NOT** start a new one.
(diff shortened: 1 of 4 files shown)


SatsAndSports:
@callebtc , I made this PR while thinking about state transitions, but you've been doing much more research on all of this, especially on the contract with the lightning nodes and I hadn't thought about that contract at all.

So I'll likely close this soon. Feel free to close early, or re-open, or whatever, with the contents of this PR

PR #430, opened 2026-08-20
Spec reorg: current NUTs only, everything pre-v3 in legacy/

robwoodgate:
Draft, for a read on the approach rather than a merge.

Stacks on #443, based on its branch so the diff here is the reorg alone. Retargets to `main` once #443 merges.

## What this does

The NUTs have been accumulating generations: three derivations in NUT-13, two locking schemes in NUT-10, and whole NUTs (11, 12, 14) that v3 keysets never touch. This moves everything pre-v3 into `legacy/`, in NUT order, so each document in the root describes exactly one protocol.

| | |
| --- | --- |
| Moved whole | NUT-11, NUT-12, NUT-14, with their vectors |
| Split | NUT-00 (secp256k1 BDHKE, V3 tokens), NUT-02 (keyset IDs V1 and V2), NUT-10 (JSON spending conditions), NUT-13 (V1 and V2 derivation), NUT-20 (pre-v3 quote message) |
| Untouched | everything with no pre-v3 half |

`legacy/README.md` maps each document to the specification that supersedes it. Nine commits, sliced per NUT.

The pass is deliberately mechanical: sections move, links follow. Where prose had to be written it is because the split left a statement wrong, and each case is called out in its commit message.

## Calls worth arguing about

**NUT-10 and NUT-20 are now mandatory.** A v3 proof cannot be spent without the transaction transcript and signing rule in NUT-10, and a mint must reject minting an unlocked quote onto a v3 keyset. `optional` reads wrong once the pre-v3 alternatives are gone. Both move to the mandatory table in the README.

**NUT-20 stays a NUT.** It still owns the `pubkey` and `signature` wire shape and the v3 lock key derivation, and defers the message to NUT-10's transcript. Dissolving it into NUT-04, NUT-10 and NUT-13 is the more aggressive simplification, and is a one-line change from here if that is preferred.

**TokenV3 is treated as legacy.** Not keyset-versioned, but deprecated, and "V3 token" against "v3 keyset" is a live confusion in a v3-only spec.

## Two things the split exposed, already fixed in #443

The TypeScript NUT-13 example built its V3 KDF message with the unframed V2 helper, so it derived different values from the Python beside it. Three orphaned links, in NUT-00, NUT-11 and the NUT-13 vectors.

## Not included

The mkdocs `nav` lives on `gh-pages-builder` and needs matching entries, or the published site loses the moved pages. Worth knowing it is already stale: it stops at NUT-22 and is missing 23 through 30.

## Checks

`prettier@3.9.1 --check` clean. `check-links.py`, added in the last commit, verifies every relative link and anchor across the tree and greps for pre-v3 prose left in a current NUT. It is meant to run after each rebase onto #443: a legacy paragraph added upstream to a file this branch does not touch applies with no conflict at all, which is the failure mode worth guarding.


robwoodgate:
Closing: keeping this on the fork for now as robwoodgate/nuts#3, so the upstream repo does not carry a duplicate taproot-secrets branch. Will reopen here once #421 merges and this can target main directly.

PR #433, opened 2026-09-03
NUT-XX: add compact state filters

crodas:
[View the NUT rendered](https://github.com/crodas/nuts/blob/compact-state-fitlers/XX.md)

Following the state of ecash or a quote today means handing the mint the exact identifier. NUT-07 carries the Y values themselves, and a NUT-17 subscription binds those identifiers to a long-lived connection, so the mint learns which proofs sit behind one IP and TLS session.

Remove the query instead. The mint groups state changes into fixed epochs and publishes one Golomb-Rice coded filter per epoch. Every wallet downloads the same bytes and matches its own objects offline, so there is nothing to correlate and no per-wallet parameter to fingerprint. Binding the new state into each element means a match reports the state directly and the wallet never has to reveal Y to learn it.

This is BIP-158 in shape, without the SipHash step and with M fixed to 2^P, which leaves a single parameter and one hash primitive to implement. The filter history also gives NUT-13 an offline way to recover state after a seed restore.

- [ ] cdk: https://github.com/cashubtc/cdk/pull/2477


robwoodgate:
Is this a redesign of: https://github.com/cashubtc/nuts/pull/261 ?

crodas:
@robwoodgate I will check #261 now because I'm not sure

XX.md
```diff
@@ -0,0 +1,255 @@
...
+
+For every object whose observable state changes during an epoch, the mint inserts exactly one element into that epoch's filter:
+
+```
+E = SHA256(DOMAIN_SEPARATOR || len32(kind) || kind || len32(id) || id || len32(state) || state)
+```
+
+Where:
+
+- `DOMAIN_SEPARATOR` is the constant byte string `b"Cashu_StateFilter_v1"` as raw ASCII bytes, not length-prefixed
```
robwoodgate, line 29: We should compute `tagged_hash("Cashu_StateFilter_v1", bytes)`, where `tagged_hash` is the BIP-340 construction `SHA256(SHA256(tag) || SHA256(tag) || message)` over the tag's UTF-8 bytes. `SHA256("Cashu_StateFilter_v1")` = `d0e04d5f4c6de0dcccf52063ba15e85402f753915cecfd566a52cae4e007647e`.

Tagged hash removes the prefix ambiguity of an unframed separator (`Cashu_StateFilter_v1` + `0...` reads the same as a future `Cashu_StateFilter_v10`), and the 64-byte midstate is cacheable for a busy mint (one SHA-256 block per element saved). We are also moving this way in other NUTs, such as Nutroot, Mint info signing, etc

crodas:
> Is this a redesign of: https://github.com/cashubtc/nuts/pull/261 ?

Not a redesign; it's a different approach with a different goal. Rather than the mint sending a wallet updates about objects the wallet has named, the mint publishes one Golomb-Rice-coded (I will update it to be a Bloom filter) filter per epoch over all its state changes, with no payload and no identifiers. A wallet coming back online fetches the filters covering the gap, computes candidate elements for whatever it cares about (proofs, mint and melt quotes), and matches them locally. Every wallet downloads the same bytes, so there is nothing to correlate. A match is probabilistic, so the wallet confirms it with NUT-07 before acting, which means it only makes an identifying request when something has probably happened.

crodas:
## Updates: went from Golomb-Rice epoch filters to per-kind Bloom block sequences.

### Encoding
- Golomb-Rice out, Bloom bitmaps in. A GR position depends on the filter's element count, so adding one element moved every other position and you had to decode the whole filter to test anything. A Bloom position depends only on the bitmap size: an insert sets bits, a test is bit reads.
- 16 bits per element, taken as the 16 words of two SHA-256 digests. Fixed 32-bit windows, so positions nest (the position at a smaller width is just a right shift). That's what lets a mint shrink a bitmap without any wallet rehashing.
- Went from two parameters (P, epoch length) down to one (b).

### Blocks instead of epochs
- A block seals when it fills, with a timeout as a backstop. Sealing on fill is what pins the false-positive rate; a clock-sealed filter's rate would just follow the mint's volume.
- Capacity is an exact integer ratio (2^b * ln2 / 16), b from 5 to 26.
- The probability claim is now honest: expected fill at capacity is about half, and the estimate is f^16 on the popcount you actually measure. Rate moved from 1 in 268M to 1 in 65,536, which is what buys the smaller filters.
- A block that seals early gets folded down to the smallest width that still holds it.

###  Per-kind sequences
- Every kind gets its own sequence, own b, own rotation, so a quiet kind isn't carried in a noisy one's bytes. Splitting by payment method stays forbidden since it'd leak the rail.
- New blind_signature kind covering issuance. That's what lets a wallet find its own past outputs without asking, and gives NUT-13 a restore that doesn't hand over every candidate B_.

### Wire format got cut down a lot
- Endpoints were /v1/filters/info + /v1/filters/{page} + /v1/filters/pending. Now just /v1/filters/info/{kind} and /v1/filters/blocks/{kind}/{height}.
- Pagination gone. Hash chain gone (prev, digest, version and the headers endpoint).
- /info/{kind} gives you b, timeout, open_interval and a list of {start, end}, one per block. A block's height is its array index.
- A block response is the raw bitmap bytes and nothing else. Kind and height are in the URL, timestamps are in /info, and b is the body length.
- count is no longer published. It's mint-side now, used for sealing and folding.

###  Open block is back
- The block still filling is served at the head of its sequence, so you don't wait a full rotation to see a state change.
- Served on a refresh interval with an ETag, never marked immutable. A clear bit in it only means "nothing yet, as of this snapshot".

### Also in there
- Normative table of exactly when a mint inserts, including what it must not insert (the UNSPENT that issuance creates, for one).
- Mints may add random noise to obscure activity. Retention is now mandatory, which is the price of height being the array position.
- Test vectors roughly doubled: seven full cases (preimage, E, both digests, all 16 words, positions, bitmap, base64) plus the script that generates them, so it's all reproducible.
- Error codes renamed: 40001 filter block out of range, 40002 unknown filter kind.

PR #435, opened 2026-09-05
NUT-08: wallets MUST ignore received zero-value blind signatures

Amperstrand:
## Background

NUT-08 (as of #426) is now fully normative on the mint's sending side: `Alice` **MUST** set each blank output's `amount` to 0, `Bob` **MUST** ignore it and imprint the amount himself, and he `MUST` return all value>0 blind signatures in order and *should* omit value-0 ones.

The wallet's **receiving** side is still unspecified: nothing tells an implementer what to do if a value-0 blind signature is nevertheless received — which is spec-legal, since omission is a SHOULD, not a MUST.

That gap recently produced a real interop failure:

1. A Cashu mint implementation built directly from the spec text (not ported from a reference codebase) signed melt change blanks verbatim instead of imprinting the refund — returning the value-0 signatures the spec only says *should* be omitted. (The mint bug is fixed on its side; the point here is what it *revealed*.)
2. cashu-ts v4 wallets then crashed during client-side change reconstruction on an **already-paid melt** — `OutputData.toProof` looks up the keyset key for the signature amount, and there is no key for `0` — instead of ignoring the value-0 signature: cashubtc/cashu-ts#1079. A self-contained repro + logs against cdk-mintd 0.17.0-rc.3 and 0.18.0-rc.0 are linked there: conformant mints omit, so the natural flow works fine — the crash is purely the receiving-side handling.

## What confused the implementer

Reading this section as the sole source of truth, "*should omit all blind signatures with value 0*" reads as a guarantee that value-0 signatures are never received, and the worked example ("`Bob` will only return these 4 blind signatures…") reinforces it. So no receiving-side handling is specified, and none feels required. That is a reasonable reading when the implementation is derived from the spec text alone — which is increasingly common (including AI-agent-implemented mints and wallets) rather than ports of CDK/Nutshell. #426 tightened the sending side; this is the mirror clarification for the receiving side.

## Proposed change

One additive sentence at the end of the ordering paragraph:

> Consequently, if `Alice` receives a blind signature with a value of 0 (e.g., from a mint that does not omit them), she **MUST** ignore it — a zero-value blind signature carries no ecash and is not an error condition; `Alice` unblinds and keeps only the blind signatures with a value greater than 0, mapped to their blank outputs in order.

Strictly additive: mints that omit (all known implementations) and wallets that already tolerate value-0 signatures are unaffected, and the order-based mapping stays well-defined for both omission styles.


robwoodgate:
How would a mint sign a 0 value amount unless it had a 0 value key in its keyset? 

And why would it have a zero value key in the first place?

This honestly sounds like agent logic, obvious to humans.

If anything, the normative change should be to change the SHOULD omit to a MUST omit?

08.md
```diff
@@ -40,7 +40,7 @@ If the `overpaid_fees = input_amount - fees - total_paid` is positive, `Bob` dec
...
+Importantly, while `Bob` does not necessarily return the same number of blind signatures as it received blank outputs from `Alice` (since some of them may be of value 0), `Bob` **MUST** return the all blank signatures with a value greater than 0 in the same order as the blank outputs were received and should omit all blind signatures with value 0. For example, if `Bob` receives 10 blank outputs but the overpaid fees only occupy 4 blind signatures, `Bob` will only return these 4 blind signatures with the appropriate imprinted amounts and omit the remaining 6 blind signatures with value 0. Due to the well-defined order of the returned blind signatures, `Alice` can map the blind signatures returned from `Bob` to the blank outputs it provided so that she can further apply the correct unblinding operations on them. Consequently, if `Alice` receives a blind signature with a value of 0 (e.g., from a mint that does not omit them), she **MUST** ignore it — a zero-value blind signature carries no ecash and is not an error condition; `Alice` unblinds and keeps only the blind signatures with a value greater than 0, mapped to their blank outputs in order.
```
robwoodgate, line 43: Thanks for surfacing this. The zero-value case turns out to be one part of a wider question about what a mint does with melt change when a keyset rotates mid-melt (see cashubtc/cdk#2447), and the two need settling together so the mint-flow wording stays consistent.

I have opened #436 to do that: it makes omission of zero-value signatures a MUST, requires amounts to be imprinted into the leading blank outputs, and settles the keyset question (change stays on the blank outputs' keyset, no substitution). It credits this PR for finding the ambiguity. Review there would be welcome.

robwoodgate:
Related: https://github.com/cashubtc/cashu-ts/issues/1079

The mint in question somehow allowed 0 value signatures to be return (because it only "should" omit, even though it should not be able to sign that value without a 0 value key), which CTS treated as an error.

Amperstrand:
Closing in favour of #436. One bit of background that may be useful: our mint signs by deriving a key for whatever amount is on the output, rather than looking it up in the keyset's amount→key map like the reference implementations do. Signing that way can produce a value-0 signature even though 0 isn't in the keyset — and with omission a SHOULD, that was wire-legal. That's how we ran into it: a refactor dropped the imprint step, v4's zero-amount blanks turned it into the wallet crash, and it was fixed the same day. #436's MUST-omit settles it.

PR #436, opened 2026-09-06
NUT-02, NUT-08: melt change keyset after rotation; zero-value signatures MUST be omitted

robwoodgate:
## Supersedes #435.

NUT-08 blank outputs are checked against an active keyset when the melt is accepted, but signed only after the payment settles. If the keyset rotates in between, NUT-02 forbids signing on it and NUT-08 says nothing, so today CDK and Nutshell refuse and the change is lost (cashubtc/cdk#2447).

Substituting the active keyset was proposed in cashubtc/cdk#2448 and rejected: it splits the keyset the outputs were derived from, the keyset that signed, and the keyset the proof carries, which breaks seed recovery and cannot work across curves.

Separately, @Amperstrand found in #435 that a mint returning zero-value change signatures is spec-legal (omission was a SHOULD) while no wallet handles it.

## Changes

- NUT-02: melt change is the one exception to active-only outputs. The mint MUST sign it with the blank outputs' keyset if that keyset was inactivated meanwhile, and MUST NOT substitute. A mint MUST NOT accept blanks on a keyset whose final_expiry precedes the melt quote expiry.
- NUT-08 mint flow: sign with the blanks' keyset; imprint into the leading blanks; MUST omit zero-value signatures (was SHOULD).
- NUT-08 wallet flow: unblind each signature with the keyset its id names, and do not reject change because that keyset is inactive.

## Implementations

- Mint: cashubtc/cdk#2460 (signs on the inactive keyset; needs the wallet-side unblinding fix noted there).
- Wallet: cashubtc/cashu-ts#1081 (main) and #1083 (v4).


crodas:
@robwoodgate I think CDK will comply with this PR https://github.com/cashubtc/cdk/pull/2460, but I will double-check. I +1 this change.

crodas (approved):

02.md
```diff
@@ -22,6 +22,8 @@ The keyset `id` is part of each `Proof` object which allows wallets to identify
 
 Mints can have multiple keysets at the same time but **MUST** have at least one `active` keyset (see [NUT-01][01]). The `active` property determines whether the mint allows generating new ecash from this keyset. `Proofs` from inactive keysets with `active=false` are still accepted as inputs but new outputs (`BlindedMessages` and `BlindSignatures`) **MUST** be from `active` keysets only.
 
+The one exception is change for overpaid Lightning fees ([NUT-08][08]). Blank outputs are checked against `active` keysets when the melt request is accepted, but are signed only after the payment settles. If their keyset has been inactivated in between, the mint **MUST** still sign the change with that keyset and **MUST NOT** substitute another. Swap and mint outputs have no such exception.
```
thesimplekid: "Special cases aren't special enough to break the rules."

robwoodgate: "Although practicality beats purity"

robwoodgate: The special case rule can be honoured if we say _"that are active when the mint accepts the request."_:

> Mints can have multiple keysets at the same time but **MUST** have at least one `active` keyset (see [NUT-01][01]). The `active` property determines whether the mint allows generating new ecash from this keyset. `Proofs` from inactive keysets with `active=false` are still accepted as inputs but new outputs (`BlindedMessages` and `BlindSignatures`) **MUST** be from keysets that are `active` when the mint accepts the request. 

We can then optionally clarify that further with:

> Accepted outputs are bound to their keyset: the mint **MUST** sign them with that keyset even if it is inactivated before signing completes, and **MUST NOT** substitute another. For swap and mint, acceptance and signing are a single step. For melt, blank change outputs ([NUT-08][08]) are accepted with the request but signed only after the payment settles.

robwoodgate: I amended the wording so there are no special cases as above

08.md
```diff
@@ -38,9 +38,13 @@ Here we describe how the mint generates `BlindSignature`s for the overpaid fees.
 
 If the `overpaid_fees = input_amount - fees - total_paid` is positive, `Bob` decomposes it to amounts supported by the keyset, typically `2^n`, and imprints them into the `blank_outputs` provided by `Alice`.
 
-`Bob` then signs these blank outputs (now with the imprinted amounts) and thus generates `BlindSignature`s. `Bob` then returns a payment status to the wallet, and, in addition, all blind signatures it generated for the overpaid fees.
+`Bob` then signs these blank outputs (now with the imprinted amounts) and thus generates `BlindSignature`s. `Bob` **MUST** sign them with the keyset of the blank outputs, even if that keyset has been inactivated since the request was accepted (see [NUT-02][02]). `Bob` then returns a payment status to the wallet, and, in addition, all blind signatures it generated for the overpaid fees.
```
callebtc, line 41: I'm not sure if this is a good idea...

thesimplekid: I agree. Inactive means it will not sign I don't think we should create an exception.

crodas: What would be the alternative? Use the new keyset to sign the change?

robwoodgate: A melt flow is:

- melt transaction accepted by mint
- irreversible and potentially lengthy payment initiated
- change blanks signed once it completes

If inactive must mean _"no new signature made by that keyset"_ then we still have a problem.

If inactive can mean _"no new transaction accepted using that keyset"_ then we do not.

Transactions always preceed output signing, so the two are effectively the same gate.

The difference is the former snapshots "issued signatures" at point of going inactive, the latter is "issued and accrued signatures".

The _accrued_ signatures are finite and knowable at the cutoff date, like accruals in financial year end accounting. The maximum liability of the accruals is `fee_reserve` plus any overpayment of inputs.

They can only reduce, never grow. If the melt fails, the original proofs can be returned as with swap.

If that's not possible for any reason (eg PoL can't handle accruals), then we are left with:

- Mint substitutes keyset - bad because it detaches `Proof` from the keyset `B_` was created under and we may soon have mixed curve keysets
- Mint denies change - bad because it burns overpayments (potentially large)

There really are no other great options I can think of?

robwoodgate: Ooh, there is one more option!... the mint doesn't return signatures, it hands back a paid mint quote.

Would require some wrinkles ironing out, like mint quote locking (nutroot), change of melt response shape etc

But might be a better overall option than NUT-08 blanks (which are lost if wallet doesn't persist the state).

Mint quote is, after all, the canonical "proofs owed, keyset TBD" vehicle.

crodas: > Ooh, there is one more option!... the mint doesn't return signatures, it hands back a paid mint quote.
> 
> Would require some wrinkles ironing out, like mint quote locking (nutroot), change of melt response shape etc
> 
> But might be a better overall option than NUT-08 blanks (which are lost if wallet doesn't persist the state).
> 
> Mint quote is, after all, the canonical "proofs owed, keyset TBD" vehicle.

I love this approach. We already have all the code for that in the wallet and the mints. We just to change some flows.

PR #438, opened 2026-09-08
NUT-02: keyset active window (active_from, active_until)

robwoodgate:
Locked outputs (HTLCs, atomic swaps, Spilman channels) need certainty that the keyset they are signed on will still be active when the lock expires. Today a mint can inactivate a keyset at any moment with no notice.

## Implementations

- [x] CDK cashubtc/cdk#2573
- [x] Nutshell cashubtc/nutshell#1166
- [x] Cashu-TS cashubtc/cashu-ts#1188

## Changes

- New keyset fields `active_from` and `active_until` (unix epoch). Neither is part of the keyset ID, so mints can add or extend them without rotating.
- `active_from` is optional and announces a keyset ahead of time. It is listed with `active=false` until then and wallets must not use it for outputs before the mint flips it.
- `active_until` is how a mint announces inactivation, and a mint must not inactivate a keyset without one. It may be any period when the keyset is first published. If added or changed later it must be at least 30 days out, and it may never move earlier. Where set, `active_from < active_until < final_expiry`.
- Net guarantee for wallets: an active keyset with `active_until` null stays active for at least 30 days from now.
- Mints should at all times have at least one active keyset per unit with `active_until` null or 30+ days out, and may overlap active keysets.
- Wallets should prefer the active keyset with the latest `active_until` (null counting as latest) for outputs with a future locktime.

The blank-change rule from #436 is deliberately not extended to `active_until`: #436 makes signing change on an inactive keyset legal, so only `final_expiry` matters there.


02.md
```diff
@@ -123,6 +123,18 @@ This effectively implies that the Mint can irrevocably remove all of the nullifi
 
 The final expiry is optional and **MAY** be omitted or `null` if the keyset has no final-expiry.
 
+### Keyset active window
+
+A mint **MAY** announce when a keyset becomes active with `active_from`, and **MUST** announce when it intends to inactivate a keyset with `active_until`. Both are unix epoch timestamps. A mint **MUST NOT** inactivate a keyset that has no published `active_until`. Neither field is part of the keyset ID, so a mint can add or extend them without rotating.
```
SatsAndSports, line 128: Technically, a breaking change isn't it? i.e. Today's mints would breach this immediately if this was merged, because they fail to satisfy _"**MUST NOT** inactivate a keyset that has no published `active_until`."_

Can we say something like: _"if a mint supports v3 keysets, then it must have an `active_until`. Otherwise, the field is optional. Once added, it should not be removed."_

That's an ugly hack, but I'm just trying to allow today's mints to be compatible, while also strongly encouraging adoption

(Although, maybe I'm just being too pedantic 😀. Maybe it's just enough to advise wallet devs that older mints might not have this field)

robwoodgate: I think it's ok here because the break doesn't require compatible mints/wallets to block older ones.

And if we wait for keyset v3, it may take longer to bring it into force?

02.md
```diff
@@ -123,6 +123,18 @@ This effectively implies that the Mint can irrevocably remove all of the nullifi
...
+
+- `active_from` announces a keyset ahead of time. The keyset is listed with `active=false` until that time, and wallets **MUST NOT** use it for outputs until the mint reports it as `active`.
+
+- `active_until` is the earliest time the mint may inactivate the keyset. The mint **MUST NOT** inactivate the keyset before `active_until` and **MUST NOT** move `active_until` earlier once published. It **MAY** be any time when the keyset is first published. If it is added or changed later, it **MUST** be at least 30 days in the future. Where set, the timestamps **MUST** satisfy `active_from < active_until < final_expiry`.
+
+A mint **SHOULD** at all times have at least one active keyset per unit whose `active_until` is `null` or at least 30 days in the future, and **MAY** keep several keysets active at once so that a successor overlaps its predecessor.
```
callebtc, line 134: maybe a qualifier here: for mints that are continuing operations and not planning to shut down

callebtc: I was referring to line 134 "mint **SHOULD** at all times have at least one..."

robwoodgate: I think the SHOULD covers that case, unless you want to make it something stronger like:

```
Unless intending to cease operations, a mint MUST at all times...
```

robwoodgate: Wording updated

callebtc (approved):

a1denvalu3 (approved):

PR #440, opened 2026-09-09
NUT-11: define lock state when now equals locktime

### LINKED ISSUE #431: NUT-11 locktime equal to mint local time undefined
The NUT-11 specification for the locktime does not define the case for when locktime equals the mint's local time and which state would fall under: Permanant, Active, or Expired.

robwoodgate:
# Closes #431

The Active Lock and Expired Lock definitions used "less than" and "greater than", leaving the exact second where the mint's clock equals `locktime` unassigned.

This PR makes the boundary explicit: the lock is active while `now <= locktime` and expired once `now > locktime`.

## Implementations

All three reference implementations already behave as specified, so no implementation PRs are needed:

- nutshell `cashu/mint/conditions.py`: `locktime_passed = secret.locktime is not None and secret.locktime < current_time`
- CDK `crates/cashu/src/nuts/nut10/mod.rs`: `.map(|locktime| locktime < current_time)`
- cashu-ts `src/crypto/NUT11.ts`: `return nowSeconds < locktime ? 'ACTIVE' : 'EXPIRED';`

### DIFF
--- a/11.md
+++ b/11.md
@@ -172,13 +172,13 @@ If the `locktime` tag is not present, or is not a valid unix time, the lock is c
 
 ### Active Lock
 
-If the `locktime` tag is a valid unix time and the mint's local clock is less than `locktime`, the lock is "active".
+If the `locktime` tag is a valid unix time and the mint's local clock is less than or equal to `locktime` (`now <= locktime`), the lock is "active".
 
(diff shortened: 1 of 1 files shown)


ye0man (approved):
ACK

thesimplekid (approved):

PR #442, opened 2026-09-10
NUT-10: Nutroot secrets (v3 keysets)

robwoodgate:
# Implementations

These are Proof of concept implementations.

- [ ] Cashu-TS https://github.com/cashubtc/cashu-ts/pull/950
- [ ] Nutshell https://github.com/cashubtc/nutshell/pull/999 (via cashubtc/nutshell/pull/1111)
- [ ] CDK https://github.com/cashubtc/cdk/issues/2433

## Summary

Formalizes nutroot secrets, the secret family for v3 (BLS) keysets, as version-scoped additions to the existing NUTs rather than a new NUT.

The core expands **NUT-10**: secrets are compressed secp256k1 points, bare or tweaked BIP341-style over a tree of declarative condition leaves (`threshold`, `after`, `hashlock`); a shared TLV grammar; the transaction transcript that every v3 input signs; key path and script path witnesses; spend info for transfer; and receive-time verification. Full design rationale lives in the [discussion draft](https://hackmd.io/jdOrJ1COQfOHOkguUF8YOg) and is deliberately kept out of the spec text.

## Per-NUT deltas:

- **NUT-00**: v3 secret bytes for `hash_to_curve_G1`; secret uniqueness stated on the `Proof` model; `si` shown in the V4 token format.
- **NUT-03/05**: every v3 input carries a witness over the transaction transcript; transactions mixing keyset versions verify per input.
- **NUT-04**: v3 mint quotes are always locked and enter the transaction as inputs, signing via the existing `signature` field. **NUT-29**: a batch is one transaction, one digest.
- **NUT-07**: a check-state entry returning a version `02` witness also returns the transaction digest it signed, as a sibling field.
- **NUT-11/14**: scoped to pre-v3 keysets.
- **NUT-13**: the v3 secret branch derives a private key (rejection sampling against `SECP256K1_N`); per-purpose derivation paths are superseded for v3.
- **NUT-18/26**: the `nutroot` payment request option (`k`, `l`, `b`), carried in creqB under new TLV tag `0x0b`.
- **NUT-20**: the pre-v3 quote signature scheme is version-scoped; v3 quote lock keys derive beside it (NUT-13 type `0x04` with an empty keyset id, so a quote key survives keyset rotation and seed-only recovery).
- **NUT-22**: version `02` BATs are point secrets whose required witness signs a request transcript (method, target, body hash: the `0x05` container) under the ordinary transaction digest rules. Spec only; implementation deferred.
- **NUT-28**: the blinding core applied one layer down: positional sender slots, receivers matching by value, verbatim NUMS bases.

## How this differs from BIP341

The scheme is called **nutroot**, not taproot, because what is borrowed is the commitment structure and little else: a public key that secretly commits to a merkle tree of alternative spending conditions, where a key-path spend is byte-identical to an unconditional one. Almost none of Bitcoin's validation rules come with it, so do not port assumptions:

- **The secret is a 33-byte compressed point, not an x-only output key.** Signatures are still BIP340, verified against the secret's x-coordinate.
- **The tweak reduces mod `n` and is never rejected.** BIP341 rejects a tweak digest at or above the curve order; here it reduces.
- **The tweak preimage takes the full 33-byte internal key.** BIP341 hashes a 32-byte x-only key there, so ported code derives a different tweak and an unspendable proof.
- **Leaves are declarative TLV records, not script.** Leaf version `0x00` selects a fixed vocabulary (`threshold`, `after`, `hashlock`). There is no interpreter, no opcodes and no stack. An unknown field of either parity rejects: odd type numbers are reserved for a possible future ignorable class, so neither Lightning's it's-OK-to-be-odd rule nor an annex-style free data channel applies here.
- **The tree shape is fixed by the leaf count, not chosen per leaf.** BIP341 lets the constructor set each leaf's depth, burying rare scripts to shorten the likely spending path. Here the fold decides the shape, so a payer can reproduce a payee's requested tree without being told it.
- **The merkle fold is normative and sorts.** Leaf hashes sort ascending, then pair per level with an unpaired hash promoted unchanged, so the root commits the leaf set rather than any transmitted order and no constructor-chosen shape travels in a control block. Depth is capped at 8.
- **The tagged-hash domains are `Cashu_NutrootLeaf`, `Cashu_NutrootBranch` and `Cashu_NutrootTweak`.** The hash construction is BIP341's, the tags are not, so no derived value is interchangeable with Bitcoin's.
- **The NUMS point is never used verbatim.** A script-only internal key is `H + u*G` with a fresh `u` per proof, disclosed so a holder can verify no key path exists.
- **There is no sighash, annex or input index.** The single digest a witness signs is the NUT-10 transaction transcript, which has no Bitcoin analogue: its inputs are proofs and paid mint quotes, its outputs blinded messages and melt quotes.

## Test vectors

New `tests/10-tests.md` (leaf forms, tree folds, worked derivations, transaction transcripts with signatures, V4 tokens with spend info) plus additions to the NUT-13/18/26/28 vectors. Every value is produced by one of two independent implementations (cashu-ts and nutshell) that pass a shared integration suite against each other, and the tree, tweak, and transcript values were re-derived independently during review.

## Stacking

Builds on #371 (BLS12-381 v3 keyset protocol). The PR is based on `bls-protocol` (#371's content with current main merged in), so the diff here is the nutroot change alone. Once #371 merges to main, this retargets to main and the stack base branch can be deleted.


robwoodgate:
Reopened as #443 from the cashubtc branch so the head and 430's base are one ref.

PR #387, opened 2026-06-10
add method field to quote responses, closes #378

callebtc:
# Implementations

- Cashu-TS: https://github.com/cashubtc/cashu-ts/pull/698
- CDK: https://github.com/cashubtc/cdk/pull/2178
- Nutshell: https://github.com/cashubtc/nutshell/pull/1055

add a `method` field to all quote responses.

closes #378

PR #258, opened 2025-05-08
Abstract NUT-04 and NUT-05 Payment Methods

davidcaseria:
Moves BOLT11 to NUT-23

PR #454, opened 2026-09-22
Editorial pass: broken anchors, invalid JSON, and names no NUT defines

Kelbie:
Nine commits of editorial fixes, grouped so each one is a single kind of
defect: broken links and anchors, JSON examples that do not parse, type and
field names that appear in no NUT, two endpoint paths no mint serves, and a
typo in NUT-13's blinding-factor derivation path. Every change is enumerated
with its reasoning in the commit message.

```
Normalize formatting in spec headers, callouts and code blocks
Use the American spellings the rest of the tree already uses
Copy-edit prose: grammar, agreement and dropped words
Fix heading levels and intro placement
Repair broken cross-references and anchors
Refer to types, fields and states by the names they are defined under
Make JSON examples parse and match their own schemas
tests: fix vectors and prose that contradict the specs they test
Fix endpoint paths and the blinding-factor derivation path
```

The NUT-13 one is the only change that can break a wallet. The legacy
derivation printed `{keyset_id_k_int}` for the blinding factor where the
secret path printed `{keyset_k_int}`. A wallet reading those as two different
account indices derives different `r` values and cannot restore its own ecash.
The path template, both worked examples in that NUT, the test vectors and all
four implementations agree the two paths share one account level.

Nothing here changes what a mint or wallet must do, so there is nothing to
implement. Where the text contradicted the implementations, the text is what
moved.

NUT-20's `PostMintQuoteBolt11Response` is the only schema touched, and the one
thing worth a second look. It declares `depends on: NUT-04`, but NUT-04's
common mint quote response already specifies `method` and `expiry: <int|null>`
alongside NUT-20's own `pubkey`, and NUT-20 has neither. It also claims to be
"the same as the `bolt11` mint quote response in NUT-23 except for `pubkey`"
while omitting NUT-23's `amount`. #258 and #387 each updated NUT-23 without
it; only `unit` was ever caught up, by hand. NUT-25 and NUT-30 are already
correct.

`prettier@3.9.1 --check` passes. Every link and anchor resolves: 7 broken
references on main, 0 here. The NUT-00, NUT-13 and NUT-28 vectors were
recomputed and reproduced byte for byte, so no vector payloads change.


Kelbie:
Each commit has the reasoning behind the changes. I was building a tool that consumes the specification and I came across a bunch of nits to polish the spec. I tried to limit it to non-contentious changes and I'm happy to make edits.

